PR #2, opened 2015-12-10
Format Change

jboy1234:
Add a more specific and branded text


DaneEveritt:
That would make page titles `Pterodactyl - Pterodactyl ...`

PR #5, opened 2015-12-11
Add bold formatting

jboy1234:


DaneEveritt:
Invalid HTML markup, in addition to not using `<strong>` in place of `<b>`. Don't really need bold there I don't think. :)

PR #7, opened 2015-12-11
Added First Implementations of German

jboy1234:


DaneEveritt:
@jboy1234  — due to licensing changes I had to remove your commits since I couldn't just relicense your work under the MIT license.

We'd love it if you would resubmit the translations again, however we have changed our process a bit and require a Contributor License Agreement. More information about it can be [found here](https://github.com/Pterodactyl/Panel/blob/development/CONTRIBUTING.md#license-agreement) and information about why we use a CLA [here](https://www.clahub.com/pages/why_cla).

PR #13, opened 2015-12-12
Account Creation in AdminCP

DDynamic:
This allows accounts to be created in the Admin Control Panel. It was tested manually.

@DaneEveritt remember to add in UUID checking when you get a chance.

### DIFF
--- a/app/Http/Controllers/Admin/AccountsController.php
+++ b/app/Http/Controllers/Admin/AccountsController.php
@@ -2,9 +2,12 @@
 
 namespace Pterodactyl\Http\Controllers\Admin;
 
+use Alert;
 use Debugbar;
-use Pterodactyl\Models\User;
+use Hash;
+use Uuid;
 
+use Pterodactyl\Models\User;
 use Pterodactyl\Http\Controllers\Controller;
 use Illuminate\Http\Request;
 
@@ -32,12 +35,35 @@ class AccountsController extends Controller
 
     public function getNew(Request $request)
     {
-        //
+        return view('admin.accounts.new');
     }
 
     public function getView(Request $request, $id)
     {
         //
+    }
+
+    public function postNew(Request $request)
+    {
+        $this->validate($request, [
+            'username' => 'required|min:4|unique:users,username',
+            'email' => 'required|email|unique:users,email',
+            'password' => 'required|confirmed|regex:((?=.*\d)(?=.*[a-z])(?=.*[A-Z]).{8,})',
+            'password_confirmation' => 'required'
+        ]);
+
+        //@TODO: re-generate UUID if conflict
+        $user = new User;
+        $user->uuid = Uuid::generate(4);
+
+        $user->username = $request->input('username');
+        $user->email = $request->input('email');
+        $user->password = Hash::make($request->input('password'));
+
+        $user->save();
+
+        Alert::success('Account has been successfully created.')->flash();
+        return redirect()->route('admin.accounts.view', ['id' => $user->id]);
     }
 
 }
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/Admin/AccountsController.php
```diff
@@ -2,9 +2,12 @@
 
 namespace Pterodactyl\Http\Controllers\Admin;
 
+use Alert;
 use Debugbar;
-use Pterodactyl\Models\User;
+use Hash;
+use Uuid;
 
+use Pterodactyl\Models\User as User;
```
DaneEveritt: You don't need to define this `as User` since it is implied that it is `User` by the namespace. If it was say `Pterodactyl\Models\LongUserNamespaceKindaVague` you could tack on `as User` to make it easier to write and read.

app/Http/Controllers/Admin/AccountsController.php
```diff
@@ -40,4 +43,28 @@ public function getView(Request $request, $id)
...
+        //@TODO: re-generate UUID if conflict
+        $user = new User;
+
+        $user->uuid = Uuid::generate(4);
+        $user->username = $request->input('username');
+        $user->email = $request->input('email');
+        $user->password = Hash::make($request->input('password'));
+
+        $user->save();
+        $user->setTotpSecret($user->id);
```
DaneEveritt: We should not be setting a TOTP Secret here. This is handled when the user enables TOTP on their own, and is reset each time TOTP is changed.

app/Http/Controllers/Base/IndexController.php
```diff
@@ -42,6 +42,21 @@ public function getIndex(Request $request)
...
 
     /**
+     * Generate a random string.
+     *
+     * @param  \Illuminate\Http\Request $request
+     * @return string
+     */
+    public function getPassword(Request $request, $length = 16)
+    {
+        if($length < 8) {
```
DaneEveritt: Try using `$length = ($length < 8) ? 8 : $length;` here.

app/Http/Routes/BaseRoutes.php
```diff
@@ -14,6 +14,9 @@ public function map(Router $router) {
             return redirect()->route('index');
         });
 
+        // Password Generation
+        $router->get('/password-gen/{length}', [ 'as' => 'admin.accounts.password', 'uses' => 'Base\IndexController@getPassword' ]);
```
DaneEveritt: This shouldn't be named `admin.accounts.password` since it is a base route. I might suggest `password-gen`.

resources/views/admin/accounts/new.blade.php
```diff
@@ -0,0 +1,89 @@
...
+
+@section('title')
+    New Account
+@endsection
+
+@section('content')
+<div class="col-md-9">
+    <ul class="breadcrumb">
+		<li><a href="/admin">Admin Control</a></li>
+		<li class="active">New Account</li>
```
DaneEveritt: The breadcrumbs should be similar to adding a new server.

```
<ul class="breadcrumb">
    <li><a href="/admin">Admin Controls</a></li>
    <li><a href="/admin/accounts">Accounts</a></li>
    <li class="active">Add New Account</li>
</ul>
```

app/Http/Controllers/Admin/AccountsController.php
```diff
@@ -40,4 +43,28 @@ public function getView(Request $request, $id)
...
+
+        Alert::success('Account has been successfully created.')->flash();
+        return redirect()->route('admin.accounts');
```
DaneEveritt: If we are successful we should be redirecting the the view page for the new account.

DaneEveritt:
In addition to those line specific notes, you're using double quotes (`"`) in the JS, I'd prefer we stick to the standard of single quotes (`'`) for the JS.

You're also using tabs in place of 4 spaces.

Thanks for the PR, I'll pull it once we get that stuff fixed!

PR #14, opened 2015-12-12
Removed hard coded words from resources\views\server\index.blade.php

jboy1234:
Added translations English and German, and translation coded


DaneEveritt:
Can you please change this to use spaces rather than tabs.

PR #15, opened 2015-12-13
Fix TOTP AJAX

DDynamic:
This PR fixes a bug when Two-Factor authentication is not enabled, the AJAX "rapid fires" and does not submit the form and log the user in as expected.

### DIFF
--- a/resources/views/auth/login.blade.php
+++ b/resources/views/auth/login.blade.php
@@ -84,7 +84,7 @@
 <div class="col-md-3"></div>
 <script type="text/javascript">
 $(document).ready(function() {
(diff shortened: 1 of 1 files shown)


resources/views/auth/login.blade.php
```diff
@@ -84,7 +84,7 @@
 <div class="col-md-3"></div>
 <script type="text/javascript">
 $(document).ready(function() {
-    $('#login-form').submit(function (event) {
+    $('#login-form').one("submit", function (event) {
```
DaneEveritt: Please use single quotes for all Javascript stuff.

This should be `.one('submit', function (event) {`

PR #17, opened 2016-01-02
View and Delete Users

DDynamic:
This pull requests implements an interface into the Admin Panel that allows
users to be viewed, modified, and deleted.

### DIFF
--- a/app/Repositories/UserRepository.php
+++ b/app/Repositories/UserRepository.php
@@ -41,4 +41,36 @@ class UserRepository
 
     }
 
+    /**
+     * Updates a user on the panel. Returns true if the update was successful.
+     *
+     * @param  string $username
+     * @param  string $email
+     * @param  string $password An unhashed version of the user's password.
+     * @return boolean
+     */
+    public function update($id, $user)
+    {
+        if(array_key_exists('password', $user)) {
+           $user['password'] = Hash::make($user['password']);
+       }
+
+        User::where('id', $id)->update($user);
+        return true;
+    }
+
+    /**
+     * Deletes a user on the panel. Returns true if the deletion was successful.
+     *
+     * @param  string $username
+     * @param  string $email
+     * @param  string $password An unhashed version of the user's password.
+     * @return boolean
+     */
+    public function delete($id)
+    {
+        User::destroy($id);
+        return true;
+    }
+
 }
--- a/resources/views/admin/accounts/view.blade.php
+++ b/resources/views/admin/accounts/view.blade.php
@@ -1,0 +1,155 @@
+@extends('layouts.admin')
+
+@section('title')
+    Viewing User
+@endsection
+
+@section('content')
+<div class="col-md-9">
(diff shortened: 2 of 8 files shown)


app/Repositories/UserRepository.php
```diff
@@ -41,4 +41,43 @@ public function create($username, $email, $password)
...
+
+    }
+
+    /**
+     * Deletes a user on the panel. Returns true if the deletion was successful.
+     *
+     * @param  string $username
+     * @param  string $email
+     * @param  string $password An unhashed version of the user's password.
+     * @return integer
```
DaneEveritt: These docs need to be updated to match the return values.

app/Repositories/UserRepository.php
```diff
@@ -41,4 +41,43 @@ public function create($username, $email, $password)
...
+     */
+    public function delete($id)
+    {
+        User::findOrFail($id)->delete();
+        return true;
```
DaneEveritt: This return seems un-necessary, since `findOrFail()` will throw a `Illuminate\Database\Eloquent\ModelNotFoundException` if it fails to find the user. We don't even really need this function in the repository I would say, since it can all be accomplished directly from the `Models\User` action.

resources/views/admin/accounts/view.blade.php
```diff
@@ -0,0 +1,155 @@
...
+    							<th>Server Name</th>
+    							<th>Node</th>
+    							<th>Connection</th>
+    							<th style="width:10%;"></th>
+    						</tr>
+    					</thead>
+    					<tbody>
+                                @foreach($servers as $server)
+    								<tr>
+    									<td><a href="/index/{{ $server->hash }}"><i class="fa fa-tachometer"></i></a></td>
```
DaneEveritt: This link needs to be updated to point to `/server/{{ $server->uuidShort }}/`

app/Repositories/UserRepository.php
```diff
@@ -41,4 +41,43 @@ public function create($username, $email, $password)
...
 
+    /**
+     * Updates a user on the panel. Returns true if the update was successful.
+     *
+     * @param  string $username
+     * @param  string $email
+     * @param  string $password An unhashed version of the user's password.
+     * @return integer
+     */
+    public function update($id, $email, $isRoot, $password = null)
```
DaneEveritt: This might be possible to do as something like:

```php
public function update($id, $object)
{
    if (array_key_exists('password', $object)) {
        $object['password'] = Hash::make($object['password']);
    }
    User::findOrFail($id)->update($object);
}
```

That would make it so you don't have to pass specific parameters every time, and when it comes time for the API, it will work well for a `PUT/PATCH` request which would only send parameters that need changing.

DaneEveritt:
Thanks for the PR! I left some notes on some potential changes that might be worth implementing. There were also a bunch of mixed tabs and spaces which could use some fixing.

Other than that, it looks great, thanks for the help!

PR #54, opened 2016-02-04
Internal Themes

DDynamic:
This PR implements the `igaster/laravel-theme` library for internal theme usage and creation.


DaneEveritt:
Your `themes.php` file is using a mix of tabs and spaces within it. Could you please make those all spaces?

Thanks! Pull will be approved pending that change.

DDynamic:
Fixed with commit amend.

PR #91, opened 2016-09-25
Add german translation

schrej:
This adds german translation for all currently available strings in resources/lang.
It seems there aren't used strings throughout the whole app because of which the translation is incomplete here and there.


DaneEveritt:
👍

PR #97, opened 2016-09-28
Correct Team Fortress 2 name

MeltedLux:
Although a TF2 + Insurgency game would be 10/10


DaneEveritt:
Also, I'll still need to fix this as a DB migration since current versions won't use the seeder again, but I'll accept this PR to fix the seeder issue at least.

PR #110, opened 2016-10-03
Make sure that you can't delete your own account.

### LINKED ISSUE #106: You can delete yourself.
While helpful after checking someones setup to just go ahead and remove yourself, it's probably not ideal to have that work normally...

dominicmfj:
This PR fixes #106.

Basically, this fix just throws a `DisplayException`, if the currently authenticated user's ID, matches the ID of the user being viewed.

### DIFF
--- a/app/Repositories/UserRepository.php
+++ b/app/Repositories/UserRepository.php
@@ -3,7 +3,6 @@
  * Pterodactyl - Panel
  * Copyright (c) 2015 - 2016 Dane Everitt <dane@daneeveritt.com>
  * Some Modifications (c) 2015 Dylan Seidt <dylan.seidt@gmail.com>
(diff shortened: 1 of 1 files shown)


app/Repositories/UserRepository.php
```diff
@@ -152,6 +154,10 @@ public function delete($id)
             throw new DisplayException('Cannot delete a user with active servers attached to thier account.');
         }
 
+        if(Auth::user()->id == $id) {
```
DaneEveritt: Consider doing a strict check here using `Auth::user()->id === $id`.

Also a typo, `thier`.

dominicmfj: Fixed.

DaneEveritt:
Thanks!

Issue #143, opened 2016-10-15
Can't enable Two-factor Authentication for Administrator account


feardevilz:
## Product
Please check the corresponding boxes below for which products this is about.

- [x] Panel
- [ ] Daemon
- [ ] Dockerfile(s) [Please list if so: __ ]

## Type
- [x] Bug or Issue
- [ ] Feature Request
- [ ] Enhancement
- [ ] Other

## What Happens
Clicking Enable Two-factor Authentication gives an error "An error occured while attempting to perform this action. Please try again."

## How to Reproduce

1. Login
2. Click Account Security
3. Click Enable Two-Factor Authentication

## Error Logs

```

[2016-10-15 01:50:18] production.ERROR: Symfony\Component\HttpKernel\Exception\NotFoundHttpException in /srv/pterodactyl/bootstrap/cache/compiled.php:8831
Stack trace:
#0 /srv/pterodactyl/bootstrap/cache/compiled.php(8155): Illuminate\Routing\RouteCollection->match(Object(Illuminate\Http\Request))
#1 /srv/pterodactyl/bootstrap/cache/compiled.php(8083): Illuminate\Routing\Router->findRoute(Object(Illuminate\Http\Request))
#2 /srv/pterodactyl/bootstrap/cache/compiled.php(8079): Illuminate\Routing\Router->dispatchToRoute(Object(Illuminate\Http\Request))
#3 /srv/pterodactyl/bootstrap/cache/compiled.php(2446): Illuminate\Routing\Router->dispatch(Object(Illuminate\Http\Request))
#4 /srv/pterodactyl/vendor/laravel/framework/src/Illuminate/Routing/Pipeline.php(53): Illuminate\Foundation\Http\Kernel->Illuminate\Foundation\Http\{closure}(Object(Illuminate\Http\Request))
#5 /srv/pterodactyl/app/Http/Middleware/LanguageMiddleware.php(58): Illuminate\Routing\Pipeline->Illuminate\Routing\{closure}(Object(Illuminate\Http\Request))
#6 /srv/pterodactyl/bootstrap/cache/compiled.php(9819): Pterodactyl\Http\Middleware\LanguageMiddleware->handle(Object(Illuminate\Http\Request), Object(Closure))
[...]

PR #149, opened 2016-10-20
Implement better API system

DaneEveritt:
API Keys are now tied to specific users and can allow individual server control even to non-admins.


DaneEveritt:
One of these commits also fixed #143

PR #161, opened 2016-11-01
only push stuff from the terminal outputQueue if there is something i…

schrej:
…nside

this allows to scroll on the console again


DaneEveritt:
Calling this a bug fix for `rc.2` to `rc.3`

PR #187, opened 2016-11-29
Allow API to set a custom ID for server creation.

death-droid:
Useful when dealing with billing systems such as WHMCS


DaneEveritt:
So this is going to have to be slightly different than the code used for adding users with unique IDs because we pass all fields in. What I would do is do `'custom' => 'sometimes|required|numeric|unique:servers,id'` which would then only check it if the field existed.

Your current method requires passing `unique_id` to all requests, which isn't necessarily desirable. Then you could just do an `isset()` check on the field for if the ID should be set.

death-droid:
Ohhh i see, that does make sense. Will make those changes to my code.

Thank you for the feedback!

PR #194, opened 2016-12-02
Add to the API the ability to search for an allocation based on the a…

death-droid:
…ssigned server id.

### DIFF
--- a/app/Http/Controllers/API/NodeController.php
+++ b/app/Http/Controllers/API/NodeController.php
@@ -238,16 +238,14 @@ class NodeController extends BaseController
      */
      public function allocationsView(Request $request, $id)
      {
-        $query = Models\Allocation::where('assigned_to', $id);
+        $query = Models\Allocation::where('assigned_to', $id)->get();
         try {
 			
-			$allocations = $query->get();
-
-            if (!$allocations->first()) {
+            if (empty($query)) {
                 throw new NotFoundHttpException('No allocations for that server were found.');
             }
     
-            return $allocations;
+            return $query;
         } catch (NotFoundHttpException $ex) {
             throw $ex;
         } catch (\Exception $ex) {
--- a/app/Http/Controllers/API/NodeController.php
+++ b/app/Http/Controllers/API/NodeController.php
@@ -226,6 +226,33 @@ class NodeController extends BaseController
          }
          return $allocations;
      }
+	 
+    /**
+     * List Node Allocation based on assigned to ID
+     *
+     * Returns a listing of the allocation for the specified server id.
+     *
+     * @Get("/nodes/allocations/{id}")
+     * @Versions({"v1"})
+     * @Response(200)
+     */
+     public function allocations_view(Request $request, $id)
+     {
+        $query = Models\Allocation::where('assigned_to', $id);
+        try {
+            if (!$query->first()) {
+                throw new NotFoundHttpException('No allocation with that ID was found.');
+            }
+    
+            $allocation = $query->first();
+
+            return $allocation->toArray();
+        } catch (NotFoundHttpException $ex) {
+            throw $ex;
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/API/NodeController.php
```diff
@@ -226,6 +226,33 @@ public function allocations(Request $request)
...
+     * @Get("/nodes/allocations/{id}")
+     * @Versions({"v1"})
+     * @Response(200)
+     */
+     public function allocations_view(Request $request, $id)
+     {
+        $query = Models\Allocation::where('assigned_to', $id);
+        try {
+            if (!$query->first()) {
+                throw new NotFoundHttpException('No allocation with that ID was found.');
```
DaneEveritt: Shouldn't this be something along the lines of 'No allocations for that server were found.'

app/Http/Controllers/API/NodeController.php
```diff
@@ -226,6 +226,33 @@ public function allocations(Request $request)
...
+            }
+    
+            $allocation = $query->first();
```
DaneEveritt: This is just going to return a single allocation for a server, there might be multiple assigned.

app/Http/Routes/APIRoutes.php
```diff
@@ -140,6 +140,11 @@ public function map(Router $router) {
                 'uses' => 'Pterodactyl\Http\Controllers\API\NodeController@allocations'
             ]);
 
+            $api->get('nodes/allocations/{id}', [
+                'as' => 'api.admin.nodes.allocations',
+                'uses' => 'Pterodactyl\Http\Controllers\API\NodeController@allocations_view'
```
DaneEveritt: Function names should follow PSR styling, `allocationsView`

app/Http/Controllers/API/NodeController.php
```diff
@@ -226,6 +226,33 @@ public function allocations(Request $request)
...
+    /**
+     * List Node Allocation based on assigned to ID
+     *
+     * Returns a listing of the allocation for the specified server id.
+     *
+     * @Get("/nodes/allocations/{id}")
+     * @Versions({"v1"})
+     * @Response(200)
+     */
+     public function allocations_view(Request $request, $id)
```
DaneEveritt: Function names should follow PSR styling, `allocationsView`

death-droid:
Let me know if theres anything else that requires changes with it.

DaneEveritt:
Im not sure that code actually works, have you tested it locally?

death-droid:
Yeah i have tested it locally. Using it for my whmcs module

app/Http/Controllers/API/NodeController.php
```diff
@@ -226,6 +226,34 @@ public function allocations(Request $request)
...
+     *
+     * Returns a listing of the allocation for the specified server id.
+     *
+     * @Get("/nodes/allocations/{id}")
+     * @Versions({"v1"})
+     * @Response(200)
+     */
+     public function allocationsView(Request $request, $id)
+     {
+        $query = Models\Allocation::where('assigned_to', $id);
```
DaneEveritt: Can you do `->get()` here, and then check the count in the try block.

app/Http/Controllers/API/NodeController.php
```diff
@@ -226,6 +226,34 @@ public function allocations(Request $request)
...
+        try {
+			
+			$allocations = $query->get();
+
+            if (!$allocations->first()) {
```
DaneEveritt: Should be able to do `if(empty($query))` to accomplish the same thing, or even `if($query->count() < 1)`

death-droid:
Ahhh yeah that's a much better way to handle it then I was handling it, i still have much to learn.
Pushed the changes

PR #204, opened 2016-12-09
Only auto-scroll terminal on new output when at the bottom

### LINKED ISSUE #153: Server console should only auto-scroll at the bottom
## Product
- [x] Panel
- [ ] Daemon
- [ ] Dockerfile(s) [Please list if so: __ ]

## Type
- [ ] Bug or Issue
- [ ] Feature Request
- [x] Enhancement
- [ ] Other

The server console when managing a server should only automatically scroll to the bottom when new messages arrive if the console is scrolled all the way down (or very close to beeing so).
This would eliminate the need for the pause button and make it much easier to read the log.

We could try to set `terminal.scrollOnEcho` to `false` when the terminal isn't scrolled down if possible.

schrej:
This updates jquery.terminal to version 0.11.23 to enable new (requested) features.

The server terminal now only automatically scrolls when it is scrolled all the way down.

If there is new output that is not visible a small notification appears at the bottom right. A click on it will scroll all the way down.

fixes #153


schrej:
There are some issues with Chrome currently, I would put this on hold (but keep the PR open) until this is resolved, then i will update this.

ET-Bent:
With latest addition (d844c41) all Chrome issues on Ubuntu have been resolved. Works flawless.

schrej:
Perfect. In my opinion it's ready to merge then.

PR #206, opened 2016-12-11
Refactor to use more laravel logic and improve compatibility with older PHP versions

spaceemotion:
As discussed in the discord, here is a PR refactoring a few bits and pieces to streamline a few methods (mostly the API) with the help of the integrated laravel helper methods and classes.

While going through the code on my PHP 5.6 development machine, I noticed a few issues regarding the usage of reserved keywords as method names. Those should be fixed, so servers without fancy PHP 7 can now run the panel just fine.

### DIFF
--- a/resources/views/admin/databases/index.blade.php
+++ b/resources/views/admin/databases/index.blade.php
@@ -36,7 +36,7 @@
         <li><a href="{{ route('admin.databases.new') }}"><i class="fa fa-plus"></i></a></li>
     </ul>
     <div class="tab-content">
-        <div class="tab-pane active" id="tab_databases">
+        <div class="tab-pane {{ Request::input('tab') == 'tab_dbservers' ? '' : 'active' }}" id="tab_databases">
             <div class="panel panel-default">
                 <div class="panel-heading"></div>
                 <div class="panel-body">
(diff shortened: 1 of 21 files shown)


resources/views/admin/databases/index.blade.php
```diff
@@ -36,7 +36,7 @@
         <li><a href="{{ route('admin.databases.new') }}"><i class="fa fa-plus"></i></a></li>
     </ul>
     <div class="tab-content">
-        <div class="tab-pane active" id="tab_databases">
+        <div class="tab-pane {{ Request::input('tab') == 'tab_dbservers' ? '' : 'active' }}" id="tab_databases">
```
DaneEveritt: Alright, that sounds like a bug with the panel that should be looked at. I'm trying to avoid those PHP blocks in the templates themselves.

DaneEveritt: Actually, I see the issue right there. it should be `?tab=tab_databases` not `?tab=tab_dbservers`, or the id needs to be changed.

spaceemotion: Not really, if you change it to `tab_databases` no tab will be displayed on page load, that's why I inverted the logic (only show this if we're not on `tab_dbservers` and show the other if we're on that specific one) :)

DaneEveritt:
I'll fix the tab things later. Thanks for the PR, looks great! 👍

Issue #221, opened 2016-12-19
Bootstrap row overflow due to too long FQDN


simpleauthority:
See this picture: 
![img](http://i.imgur.com/r34Rzva.png)

I have long FQDNs. Most users may not, however due to this long FQDN the bootstrap row overflows. It may be resolved by resizing the text for abnormally long FQDNs.

This is not a usability issue, just a cosmetic issue and bothers me a bit. :)

DaneEveritt:
Closed with new theme.

PR #222, opened 2016-12-19
Add styling rules to limit size of the FQDN column.

simpleauthority:
Add styling rules to limit size of the FQDN column. Break words for very long FQDNs.

References #221.

### DIFF
--- a/resources/views/base/index.blade.php
+++ b/resources/views/base/index.blade.php
@@ -59,7 +59,7 @@
                         @endif
                         <td><a href="/server/{{ $server->uuidShort }}">{{ $server->name }}</a></td>
                         <td>{{ $server->nodeName }}</td>
-                        <td><code>@if(!is_null($server->ip_alias)){{ $server->ip_alias }}@else{{ $server->ip }}@endif:{{ $server->port }}</code></td>
+                        <td style="max-width:290px;word-wrap:break-word;"><code>@if(!is_null($server->ip_alias)){{ $server->ip_alias }}@else{{ $server->ip }}@endif:{{ $server->port }}</code></td>
                         <td class="text-center" data-action="players">--</td>
                         <td class="text-center hidden-sm hidden-xs"><span data-action="memory">--</span> / {{ $server->memory === 0 ? '&infin;' : $server->memory }} MB</td>
                         <td class="text-center hidden-sm hidden-xs"><span data-action="cpu" data-cpumax="{{ $server->cpu }}">--</span> %</td>


simpleauthority:
You may not enjoy the fact that it breaks words. In this case, I may suggest simply lowering the font-size or using internal php magic to force-shorten the too-long FQDN and add an ellipsis (...) to the end which will avoid needing to add styling rules.

simpleauthority:
How it looks with this:
img

resources/views/base/index.blade.php
```diff
@@ -59,7 +59,7 @@
                         @endif
                         <td><a href="/server/{{ $server->uuidShort }}">{{ $server->name }}</a></td>
                         <td>{{ $server->nodeName }}</td>
-                        <td><code>@if(!is_null($server->ip_alias)){{ $server->ip_alias }}@else{{ $server->ip }}@endif:{{ $server->port }}</code></td>
+                        <td style="max-width:290px;word-wrap:break-word;"><code>@if(!is_null($server->ip_alias)){{ $server->ip_alias }}@else{{ $server->ip }}@endif:{{ $server->port }}</code></td>
```
schrej: Not very elegant in my opinion. style="" attributes are to be avoided in general and this will also only work on large screens. This should be done with media queries and the same different widths bootstrap is using so it will work across various screen sizes.

simpleauthority: @schrej I agree it's not elegant though I am not a front-end developer so I will leave that to them. I was proposing a fix that worked for me.

simpleauthority: @schrej However, your comment about style="" attributes being avoided is kinda weird, seeing as they use them everywhere in that file.

schrej: Yeah, there are a few, I also use them sometimes, especially when it is only one attribute (like colour or width that I want to change). But mostly it's better to use CSS classes so things can be changed more easily in the future. (You won't have to change stuff at multiple occurrences)

DaneEveritt:
I think we could just add a CSS class that handles word-wrapping since I can see this being needed on quite a few other columns around the panel.

Also, with 0.6.0 implementing a new theme it might be best to wait for that to be released before implementing this fix. I'm going to leave this open for discussion, let me know what you think.

DaneEveritt:
I'm going to close this as we've changed our theme for 0.6.0 which addresses these issues by making scrollable tables when things get too long.

Thanks for your PR, and I hope you'll continue to support this project! 🛩

PR #226, opened 2016-12-29
Allow listing a user by both ID and email. Useful for checking if a u…

death-droid:
…ser exists by its email.


DaneEveritt:
Maybe we should rename `$id` to something else to prevent any confusion later on? Looks fine to me otherwise, other than the StyleCI change. 👍

Issue #242, opened 2017-01-04
Adding Locations and Nodes from CLI


parkervcp:
### What to add to
- [x] Panel
- [x] Daemon
### Request Type
- [x] Feature Request
- [x]  Enhancement

Add CLI commands for adding locations and nodes to the panel.

Also support for auto-grabbing daemon configs for daemons to connect with a single use key for initial setup.

parkervcp:
I am looking at "I have the node installed" and have a 
`npm start -panel_url "http(s)://subdomain.domain.tld" -key <insertrandomstringhere>` 
and the daemon starts and grabs it's config

schrej:
But that requires to manually install the daemon before doing that. I guess a fully automated installer with automatically added config would be even better. But we could do both I'd say. For when the auto installer fails to at least automatically configure.

parkervcp:
I have docker containers with a working daemon install...

schrej:
ah right, completely forgot about those, ok, then we/i'll do both

Issue #218, opened 2016-12-17
Fully automate wings installation


schrej:
Now that we have an auto-install script for wings, installing it should be fully automated by automatically downloading the config from the panel.

A one-time link should be creatable that works for a few minutes with a token that allows downloading the install script and the config (something like `panel.something.tld/install-node/<somerandomtokenwhichmakesitsecure>`).
Also, a ready-to-paste command should be shown in the panel that can be copied to the node server to make installing as easy as possible.

- [x] Automatic wings configuration
- [ ] wings auto-install script with auto-configuration

DaneEveritt:
I'm going to assign this to you to work out a plan for how it should work. I can add endpoints if needed, just let me know. 👍

schrej:
Well its rather simple. We just serve the install script on that Endpoint. We just need to add a Part to write the specific config for the node in question. I will implement that as well though. ;)

schrej:
As we have a way to download the configuration from the panel, we only need an install script for wings hosted somewhere (still voting for `fly.pterodactyl.io/wings`) and make it consume a configuration retrieval token and call the auto configure script to make the setup fully automatic.

PR #245, opened 2017-01-07
Configuration retrival for wings via token

schrej:
Allows to create tokens to retrieve the configuration of specific nodes. The tokens are valid for 5 minutes and can only be used once.

On creation of a node an info message is shown about this feature.

#242 and #218

### DIFF
--- a/app/Http/Controllers/Admin/NodesController.php
+++ b/app/Http/Controllers/Admin/NodesController.php
@@ -279,7 +279,8 @@ class NodesController extends Controller
         ]);
     }
 
-    public function getConfigurationToken(Request $request, $id) {
+    public function getConfigurationToken(Request $request, $id)
+    {
         // Check if Node exists. Will lead to 404 if not.
         Models\Node::findOrFail($id);
 
@@ -290,10 +291,10 @@ class NodesController extends Controller
         $token->expires_at = Carbon::now()->addMinutes(5); // Expire in 5 Minutes
         $token->save();
 
-        $token_response = array(
+        $token_response = [
             'token' => $token->token,
-            'expires_at' => $token->expires_at->toDateTimeString()
-        );
+            'expires_at' => $token->expires_at->toDateTimeString(),
+        ];
 
         return response(json_encode($token_response), 200)
             ->header('Content-Type', 'application/json');
--- a/app/Http/Controllers/Admin/NodesController.php
+++ b/app/Http/Controllers/Admin/NodesController.php
@@ -27,8 +27,8 @@ namespace Pterodactyl\Http\Controllers\Admin;
 use DB;
 use Log;
 use Alert;
+use Carbon;
 use Validator;
-use Carbon\Carbon;
 use Pterodactyl\Models;
 use Illuminate\Http\Request;
 use Pterodactyl\Exceptions\DisplayException;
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/Admin/NodesController.php
```diff
@@ -28,6 +28,7 @@
 use Log;
 use Alert;
 use Validator;
+use Carbon\Carbon;
```
DaneEveritt: You can access Carbon with `use Carbon;` due to [this](https://github.com/Pterodactyl/Panel/blob/develop/config/app.php#L185).

app/Http/Controllers/Admin/NodesController.php
```diff
@@ -276,4 +278,25 @@ public function deleteNode(Request $request, $id)
...
+        $token->token = str_random(32);
+        $token->expires_at = Carbon::now()->addMinutes(5); // Expire in 5 Minutes
+        $token->save();
+
+        $token_response = [
+            'token' => $token->token,
+            'expires_at' => $token->expires_at->toDateTimeString(),
+        ];
+
+        return response(json_encode($token_response), 200)
```
DaneEveritt:
```php
return response()->json($token_response, 200);
```

Will return it as JSON.

app/Http/Controllers/Remote/RemoteController.php
```diff
@@ -107,4 +108,30 @@ public function event(Request $request)
 
         return response('', 201);
     }
+
+    public function getConfiguration(Request $request, $tokenString)
+    {
+        // Try to query the token and the node from the database
+        try {
+            $token = Models\NodeConfigurationToken::where('token', $tokenString)->firstOrFail();
```
DaneEveritt: I think we could include the date/time checking in this query as well to avoid doing it later.

app/Http/Routes/AdminRoutes.php
```diff
@@ -286,6 +286,11 @@ public function map(Router $router)
                 'as' => 'admin.nodes.delete',
                 'uses' => 'Admin\NodesController@deleteNode',
             ]);
+
+            $router->get('/{id}/configurationtoken', [
+                'as' => 'admin.nodes.configurationtoken',
```
DaneEveritt: `admin.nodes.configuration-token` would probably match the other routes better.

database/migrations/2017_01_07_154228_create_node_configuration_tokens_table.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public function up()
+    {
+        Schema::create('node_configuration_tokens', function (Blueprint $table) {
+            $table->increments('id');
+            $table->char('token', 32);
+            $table->timestamp('expires_at');
+            $table->integer('node')->unsigned();
+            $table->foreign('node')
+                ->references('id')->on('nodes');
```
DaneEveritt: This should be one line.

app/Http/Controllers/Admin/NodesController.php
schrej: And there is my lack of laravel knowledge 😅

database/migrations/2017_01_07_154228_create_node_configuration_tokens_table.php
schrej: Oh yes, it's a legacy from the onDelete

app/Http/Controllers/Remote/RemoteController.php
schrej: If i do it further down I can show a "token expired" error. I also use that in the configure script in the node and also the expired token gets deleted.

PR #268, opened 2017-01-23
Repair the admin side add user functionality.

death-droid:


DaneEveritt:
Good call, hadn't thought about that one. Will probably get cleaned up more whenever I get to re-theming the Admin side, but I'll pull this so we have it in there for now.

Issue #271, opened 2017-01-24
Autodeploy throws unhandled exception when no location is selected


schrej:
EDIT: I'm a little bit stupid.

We should provide a better error message when using autodeploy without selecting a location. It's an unhandled exception at the moment.

DaneEveritt:
Closed by #272

PR #272, opened 2017-01-25
Location should be validated on creation of a server,

death-droid:
Fixes issue #271


death-droid:
styleci just hates me :'(

app/Repositories/ServerRepository.php
```diff
@@ -92,6 +92,7 @@ public function create(array $data)
             'disk' => 'required|numeric|min:0',
             'service' => 'required|numeric|min:1|exists:services,id',
             'option' => 'required|numeric|min:1|exists:service_options,id',
+            'location' => 'required|numeric|min:1|exists:locations,id',
```
schrej: Could we validate the others as well, depending on wether `auto_deploy` is set or not?

DaneEveritt: That is done slightly lower in the code.

PR #274, opened 2017-01-28
show users’ subuser servers on user detail view

### LINKED ISSUE #209: Feature Request: User's associated subuser
You have a section for a user's associated servers how about a users associated subuser?

Eg associated to servers x and y
associated subuser for z

schrej:
Shows servers a user is assigned to as subuser in the user detail view.
Removed some useless information (Server user and Server node) and optimized query.

image

Also adds a links from the subusers table of the server to the admin user detail view (if the logged in user is an admin).

Fixes #209

### DIFF
--- a/app/Http/Controllers/Admin/UserController.php
+++ b/app/Http/Controllers/Admin/UserController.php
@@ -30,6 +30,7 @@ use Alert;
 use Illuminate\Http\Request;
 use Pterodactyl\Models\User;
 use Pterodactyl\Models\Server;
+use Pterodactyl\Models\Subuser;
 use Pterodactyl\Exceptions\DisplayException;
 use Pterodactyl\Http\Controllers\Controller;
 use Pterodactyl\Repositories\UserRepository;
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Admin/UserController.php
```diff
@@ -86,10 +87,9 @@ public function getView(Request $request, $id)
     {
         return view('admin.users.view', [
             'user' => User::findOrFail($id),
-            'servers' => Server::select('servers.*', 'nodes.name as nodeName', 'locations.long as location')
-                ->join('nodes', 'servers.node', '=', 'nodes.id')
-                ->join('locations', 'nodes.location', '=', 'locations.id')
-                ->where('owner', $id)
+            'servers' => Server::select('servers.*')
+                ->whereIn('servers.id', Subuser::select('server_id')->where('user_id', $id)->get())
```
schrej: Is there a better way to do nested queries with laravel?

DaneEveritt: So I think this is going to break as soon as we implement #286, but as soon as that is implemented there should be more straightforward logic for this.

PR #285, opened 2017-02-02
add phraseapp in context editor

schrej:
I've added a `PHRASE_IN_CONTEXT` .env option and replaced the TranslationServiceProvider with a custom one that uses a custom Translator when the option is set to true. This allows to then replace every translation with `{{__phrase_<key>__}}` as required by PhraseApp. Also in the layouts the PhraseApp JavaScript code gets added when the option is enabled.

I also cut away the first part of the key because PhraseApp ignores those. We need to make sure to have unique keys across all files to make this work as expected.

### DIFF
--- a/app/Extensions/PhraseAppTranslator.php
+++ b/app/Extensions/PhraseAppTranslator.php
@@ -22,42 +22,24 @@
  * SOFTWARE.
  */
 
-namespace Pterodactyl\Providers;
+namespace Pterodactyl\Extensions;
 
-use Pterodactyl\Extensions\Translator as PterodactylTranslator;
-use Illuminate\Translation\Translator;
-use Illuminate\Translation\TranslationServiceProvider;
+use Illuminate\Translation\Translator as LaravelTranslator;
 
-class TranslationProvider extends TranslationServiceProvider {
-
+class PhraseAppTranslator extends LaravelTranslator
+{
     /**
-     * Register the service provider.
+     * Get the translation for the given key.
      *
-     * @return void
+     * @param  string  $key
+     * @param  array   $replace
+     * @param  string|null  $locale
+     * @param  bool  $fallback
+     * @return string|array|null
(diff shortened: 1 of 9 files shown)


config/app.php
```diff
@@ -137,7 +139,7 @@
         Illuminate\Redis\RedisServiceProvider::class,
         Illuminate\Auth\Passwords\PasswordResetServiceProvider::class,
         Illuminate\Session\SessionServiceProvider::class,
-        Illuminate\Translation\TranslationServiceProvider::class,
+        //Illuminate\Translation\TranslationServiceProvider::class,
```
DaneEveritt: I'd say just remove this completely.

app/Providers/PhraseAppTranslationProvider.php
```diff
@@ -0,0 +1,63 @@
...
+ * SOFTWARE.
+ */
+
+namespace Pterodactyl\Providers;
+
+use Pterodactyl\Extensions\PhraseAppTranslator;
+use Illuminate\Translation\TranslationServiceProvider;
+use Illuminate\Translation\Translator as IlluminateTranslator;
+
+class PhraseAppTranslationProvider extends TranslationServiceProvider {
```
DaneEveritt: Brace should be on a new line.

app/Extensions/PhraseAppTranslator.php
```diff
@@ -0,0 +1,45 @@
...
+     *
+     * @param  string  $key
+     * @param  array   $replace
+     * @param  string|null  $locale
+     * @param  bool  $fallback
+     * @return string|array|null
+     */
+    public function get($key, array $replace = [], $locale = null, $fallback = true)
+    {
+        $key = substr($key, strpos($key, '.') + 1);
```
DaneEveritt: Should be a blank line before the return.

PR #286, opened 2017-02-02
Implementation of improved Model logic and relationships throughout code base.

DaneEveritt:
Generally improve all Models with better use of Eloquent ORM features and getters for some stuff like hashed IDs etc.

https://laravel.com/docs/5.4/eloquent
https://laravel.com/docs/5.4/eloquent-relationships
https://laravel.com/docs/5.4/eloquent-mutators


app/Http/Controllers/API/ServerController.php
```diff
@@ -122,7 +122,7 @@ public function view(Request $request, $id)
                 'pack'
             )->first();
             if ($request->input('daemon') === 'true') {
-                $node = Models\Node::findOrFail($server->node);
+                $node = Models\Node::findOrFail($server->node_id);
```
schrej: shouldn't this just be `$server->node` as that will lazy load it?

app/Http/Controllers/API/User/InfoController.php
```diff
@@ -37,7 +37,7 @@ public function me(Request $request)
                 'id' => $server->uuidShort,
                 'uuid' => $server->uuid,
                 'name' => $server->name,
-                'node' => $server->nodeName,
+                'node' => $server->node_idName,
```
schrej: Either `node_id_name` or better `node_name` to have a little consistency

app/Http/Controllers/Remote/RemoteController.php
```diff
@@ -66,7 +66,7 @@ public function postInstall(Request $request)
             ], 422);
         }
 
-        $node = Models\Node::findOrFail($server->node);
+        $node = Models\Node::findOrFail($server->node_id);
```
schrej: Again, why not `$server->node` ?

app/Http/Controllers/Remote/RemoteController.php
```diff
@@ -93,7 +93,7 @@ public function event(Request $request)
...
```
schrej: see above

app/Models/Server.php
```diff
@@ -140,11 +120,11 @@ public static function getUserServers($paginate = null)
...
+        )->join('nodes', 'servers.node_id', '=', 'nodes.id')
         ->join('locations', 'nodes.location', '=', 'locations.id')
-        ->join('services', 'servers.service', '=', 'services.id')
-        ->join('service_options', 'servers.option', '=', 'service_options.id')
-        ->join('allocations', 'servers.allocation', '=', 'allocations.id');
+        ->join('services', 'servers.service_id', '=', 'services.id')
+        ->join('service_options', 'servers.option_id', '=', 'service_options.id')
+        ->join('allocations', 'servers.allocation_id', '=', 'allocations.id');
 
         if (self::$user->root_admin !== 1) {
```
schrej: could be replaced by `self::$user->isRootAdmin()`

app/Models/Server.php
```diff
@@ -140,11 +120,11 @@ public static function getUserServers($paginate = null)
...
             'service_options.name as a_serviceOptionName'
-        )->join('nodes', 'servers.node', '=', 'nodes.id')
+        )->join('nodes', 'servers.node_id', '=', 'nodes.id')
         ->join('locations', 'nodes.location', '=', 'locations.id')
-        ->join('services', 'servers.service', '=', 'services.id')
-        ->join('service_options', 'servers.option', '=', 'service_options.id')
-        ->join('allocations', 'servers.allocation', '=', 'allocations.id');
+        ->join('services', 'servers.service_id', '=', 'services.id')
+        ->join('service_options', 'servers.option_id', '=', 'service_options.id')
+        ->join('allocations', 'servers.allocation_id', '=', 'allocations.id');
```
schrej: Do we need all this information to list user servers? This should be done with the new model stuff instead of manually joining.

app/Repositories/ServerRepository.php
```diff
@@ -374,7 +374,7 @@ public function updateDetails($id, array $data)
 
         try {
             $server = Models\Server::findOrFail($id);
-            $owner = Models\User::findOrFail($server->owner);
+            $owner = Models\User::findOrFail($server->owner_id);
```
schrej: replace with `$server->owner` (and add necessary Model changes)

app/Repositories/ServerRepository.php
```diff
@@ -405,8 +405,8 @@ public function updateDetails($id, array $data)
             }
 
             // If we need to update do it here.
-            $node = Models\Node::getByID($server->node);
-            $client = Models\Node::guzzleRequest($server->node);
+            $node = Models\Node::getByID($server->node_id);
```
schrej: `$server->node`

app/Repositories/ServerRepository.php
```diff
@@ -461,8 +461,8 @@ public function updateContainer($id, array $data)
             $server->image = $data['image'];
             $server->save();
 
-            $node = Models\Node::getByID($server->node);
-            $client = Models\Node::guzzleRequest($server->node);
+            $node = Models\Node::getByID($server->node_id);
```
schrej: `$server->node`

app/Repositories/ServerRepository.php
```diff
@@ -520,7 +520,7 @@ public function changeBuild($id, array $data)
 
         try {
             $server = Models\Server::findOrFail($id);
-            $allocation = Models\Allocation::findOrFail($server->allocation);
+            $allocation = Models\Allocation::findOrFail($server->allocation_id);
```
schrej: use `$server->allocation` (add required model changes)

app/Repositories/ServerRepository.php
```diff
@@ -635,8 +635,8 @@ public function changeBuild($id, array $data)
             $server->save();
 
             if (! empty($newBuild)) {
-                $node = Models\Node::getByID($server->node);
-                $client = Models\Node::guzzleRequest($server->node);
+                $node = Models\Node::getByID($server->node_id);
```
schrej: `$server->node`

app/Repositories/ServerRepository.php
```diff
@@ -747,8 +747,8 @@ public function updateStartup($id, array $data, $admin = false)
                 $model->save();
             }
 
-            $node = Models\Node::getByID($server->node);
-            $client = Models\Node::guzzleRequest($server->node);
+            $node = Models\Node::getByID($server->node_id);
```
schrej: `$server->node`

app/Repositories/ServerRepository.php
```diff
@@ -797,7 +797,7 @@ public function deleteServer($id, $force)
     public function deleteNow($id, $force = false)
     {
         $server = Models\Server::withTrashed()->findOrFail($id);
-        $node = Models\Node::findOrFail($server->node);
+        $node = Models\Node::findOrFail($server->node_id);
```
schrej: `$server->node`

app/Repositories/ServerRepository.php
```diff
@@ -888,7 +888,7 @@ public function toggleInstall($id)
     public function suspend($id, $deleted = false)
     {
         $server = ($deleted) ? Models\Server::withTrashed()->findOrFail($id) : Models\Server::findOrFail($id);
-        $node = Models\Node::findOrFail($server->node);
+        $node = Models\Node::findOrFail($server->node_id);
```
schrej: `$server->node`

app/Repositories/ServerRepository.php
```diff
@@ -928,7 +928,7 @@ public function suspend($id, $deleted = false)
     public function unsuspend($id)
     {
         $server = Models\Server::findOrFail($id);
-        $node = Models\Node::findOrFail($server->node);
+        $node = Models\Node::findOrFail($server->node_id);
```
schrej: `$server->node`

app/Repositories/ServerRepository.php
```diff
@@ -963,7 +963,7 @@ public function unsuspend($id)
     public function updateSFTPPassword($id, $password)
     {
         $server = Models\Server::findOrFail($id);
-        $node = Models\Node::findOrFail($server->node);
+        $node = Models\Node::findOrFail($server->node_id);
```
schrej: `$server->node`

app/Repositories/SubuserRepository.php
```diff
@@ -184,8 +184,8 @@ public function create($sid, array $data)
             // We contact even if they don't have any daemon permissions to overwrite
             // if they did have them previously.
 
-            $node = Models\Node::getByID($server->node);
-            $client = Models\Node::guzzleRequest($server->node);
+            $node = Models\Node::getByID($server->node_id);
```
schrej: `$server->node`

app/Repositories/SubuserRepository.php
```diff
@@ -240,8 +240,8 @@ public function delete($id)
         try {
             Models\Permission::where('user_id', $subuser->user_id)->where('server_id', $subuser->server_id)->delete();
 
-            $node = Models\Node::getByID($server->node);
-            $client = Models\Node::guzzleRequest($server->node);
+            $node = Models\Node::getByID($server->node_id);
```
schrej: `$server->node`

app/Repositories/SubuserRepository.php
```diff
@@ -318,8 +318,8 @@ public function update($id, array $data)
             // Contact Daemon
             // We contact even if they don't have any daemon permissions to overwrite
             // if they did have them previously.
-            $node = Models\Node::getByID($server->node);
-            $client = Models\Node::guzzleRequest($server->node);
+            $node = Models\Node::getByID($server->node_id);
```
schrej: `$server->node`

app/Services/NotificationService.php
```diff
@@ -48,7 +48,7 @@ class NotificationService
     public function __construct(Server $server)
     {
         $this->server = $server;
-        $this->user = User::findOrFail($server->owner);
+        $this->user = User::findOrFail($server->owner_id);
```
schrej: `$server->owner` (with required model stuff)

resources/themes/pterodactyl/base/index.blade.php
```diff
@@ -62,7 +62,7 @@
                         <tr class="dynamic-update" data-server="{{ $server->uuidShort }}">
                             <td><code>{{ $server->uuidShort }}</code></td>
                             <td><a href="{{ route('server.index', $server->uuidShort) }}">{{ $server->name }}</a></td>
-                            <td>{{ $server->nodeName }}</td>
+                            <td>{{ $server->node_idName }}</td>
```
schrej: as above, `node_id_name` or `node_name`, shouldn't it be `$server->node->name` actually?

DaneEveritt:
I just want to say that that was some dedicated review right there @schrej. I'm starting that implementation right now Admin side. Will try to move as much stuff into corrected models as possible.

app/Http/Controllers/API/ServerController.php
DaneEveritt: Yes, as mentioned in Discord, hadn't gotten to this part yet.

schrej: Just do it! 😂 
Before merging, this PR should be fully done with 0.6.0 (user AND admin side)

schrej:
You are doing the same, even beeing picky about StyleCI stuff that can be fixex by automatic PRs 😜

database/migrations/2017_02_02_175548_UpdateColumnNames.php
```diff
@@ -0,0 +1,78 @@
...
+            $table->dropIndex('servers_allocation_id_foreign');
+            $table->dropIndex('servers_service_id_foreign');
+            $table->dropIndex('servers_option_id_foreign');
+
+            $table->renameColumn('node_id', 'node');
+            $table->renameColumn('owner_id', 'owner');
+            $table->renameColumn('allocation_id', 'allocation');
+            $table->renameColumn('service_id', 'service');
+            $table->renameColumn('option_id', 'option');
+            $table->renameColumn('pack_id', 'pack');
```
schrej: you need to drop the index for that as well. Also remember `dropForeign(['column])`(discord)

app/Models/Allocation.php
```diff
@@ -48,8 +48,30 @@ class Allocation extends Model
...
+         return (is_null($this->ip_alias)) ? $this->ip : $this->ip_alias;
+     }
+
+     /**
+      * Accessor to quickly determine if this allocation has an alias.
+      *
+      * @param  null|string $value
+      * @return boolean
+      */
+     public function getHasAliasAttribute($value)
```
DaneEveritt: Can't do that, Laravel conventions.

app/Models/User.php
```diff
@@ -156,4 +146,73 @@ public function isRootAdmin()
...
+        return $subuser->daemonSecret;
+    }
+
+    /**
+     * Returns an array of all servers a user is able to access.
+     * Note: does not account for user admin status.
+     *
+     * @return array
+     */
+    public function serverAccessArray()
```
DaneEveritt: Sometimes I only need the array of server IDs to use in custom queries, othertimes I want the collection.

app/Observers/ServerObserver.php
```diff
@@ -59,12 +59,12 @@ public function created(Server $server)
         event(new Events\Server\Created($server));
 
         // Queue Notification Email
-        $user = Models\User::findOrFail($server->owner);
-        $node = Models\Node::select('name')->where('id', $server->node)->first();
+        $user = Models\User::findOrFail($server->owner_id);
```
DaneEveritt: Its `$server->user` but good catch.

app/Http/Controllers/Admin/LocationsController.php
```diff
@@ -43,35 +43,21 @@ public function __construct()
...
             return response()->json([
-                'error' => 'No location with that ID exists on the system.',
-            ], 404);
-        }
-
-        if ($model->a_nodeCount > 0 || $model->a_serverCount > 0) {
-            return response()->json([
-                'error' => 'You cannot remove a location that is currently assigned to a node or server.',
+                'error' => 'You cannot remove a location that is currently assigned to a node.',
             ], 422);
```
DaneEveritt:
```
The request could not be completed due to a conflict with the current state of the target resource. This code is used in situations where the user might be able to resolve the conflict and resubmit the request.
```

That seems like the wrong code to use here.

```
422 UNPROCESSABLE ENTITY
The server understands the content type of the request entity (hence a 415 Unsupported Media Type status code is inappropriate), and the syntax of the request entity is correct (thus a 400 Bad Request status code is inappropriate) but was unable to process the contained instructions.

For example, this error condition may occur if an XML request body contains well-formed (i.e., syntactically correct), but semantically erroneous, XML instructions.```

File: app/Models/Permission.php [hunk 27]
```diff
@@ -42,22 +50,35 @@ class Permission extends Model
...
+     * Cast values to correct type.
+     *
+     * @var array
+     */
+    protected $casts = [
+        'subuser_id' => 'integer',
+    ];
 
+    /**
+     * Find permission by permission node.
```
DaneEveritt: I call permission's a "permission node", pretty sure thats what you were referring to here.

If I make them static I have to call them statically, which I do not do.

app/Models/Server.php
```diff
@@ -102,111 +104,116 @@ public function __construct()
...
-     * DO NOT USE THIS TO MODIFY SERVER DETAILS OR SAVE THOSE DETAILS.
-     * YOU WILL OVERWRITE THE SECRET KEY AND BREAK THINGS.
+     * Return an instance of the Guzzle client for this specific server using defined access token.
      *
-     * @param  string $uuid The Short-UUID of the server to return an object about.
-     * @return \Illuminate\Database\Eloquent\Collection
+     * @return \GuzzleHttp\Client
      */
-    public static function getByUUID($uuid)
+    public function guzzleClient()
```
DaneEveritt: I think this way works, as it returns the client which is then used. Don't think we necessarily need to make it more complex.

app/Policies/ServerPolicy.php
```diff
@@ -40,6 +40,23 @@ public function __construct()
...
+     * @param $permission
+     * @return bool
+     */
+    private function checkPermission(User $user, Server $server, $permission)
+    {
+        if ($this->isOwner($user, $server)) {
+            return true;
+        }
+
+        return $user->permissions()->server($server)->permission($permission)->exists();
```
DaneEveritt: This works out nicely because we can always change this query in the future. I see what you mean, but this keeps it pretty flexible for now.

database/migrations/2017_02_02_175548_UpdateColumnNames.php
DaneEveritt: The indexes are being dropped on this, are they not?

app/Models/Server.php
schrej: It's just better naming IMO, not that important though

database/migrations/2017_02_02_175548_UpdateColumnNames.php
schrej: The foreign key for 'pack' is what I meant, sorry

app/Models/Permission.php
schrej: Well there is a similar method below where it's called filter. Also there could be the same node in different servers right? So it actually does filter, not find

If you don't need an instance, why create one to call the method?

app/Models/ServiceOptions.php
```diff
@@ -48,10 +48,52 @@ class ServiceOptions extends Model
       * @var array
```
DaneEveritt: Fixed

database/migrations/2017_02_02_175548_UpdateColumnNames.php
DaneEveritt: all fixed, thanks.

app/Repositories/ServerRepository.php
```diff
@@ -708,13 +687,13 @@ public function updateStartup($id, array $data, $admin = false)
...
                     if (isset($data[$variable->env_variable]) && empty($data[$variable->env_variable])) {
-                        if ($variable->required === 1) {
+                        if ($variable->required) {
                             throw new DisplayException('A required service option variable field (' . $variable->env_variable . ') was included in this request but was left blank.');
                         }
                     }
 
                     // Variable hidden and/or not user editable
-                    if (($variable->user_viewable === 0 || $variable->user_editable === 0) && ! $admin) {
+                    if ((! $variable->user_viewable || ! $variable->user_editable) && ! $admin) {
```
schrej: i somehow don't like spaces between exclamation marks but thats rather personal opinion i guess :P

app/Repositories/ServerRepository.php
```diff
@@ -875,7 +846,7 @@ public function toggleInstall($id)
         if ($server->installed === 2) {
             throw new DisplayException('This server was marked as having a failed install, you cannot override this.');
         }
-        $server->installed = ($server->installed === 1) ? 0 : 1;
+        $server->installed = ($server->installed) ? 0 : 1;
```
schrej: the parentheses can be omitted

app/Repositories/ServiceRepository/Variable.php
```diff
@@ -60,28 +60,29 @@ public function create($id, array $data)
...
+        $variable = Models\ServiceVariable::create($data);
 
-        return $variable->save();
+        return $variable;
     }
 
     public function delete($id)
     {
-        $variable = Models\ServiceVariables::findOrFail($id);
+        $variable = Models\ServiceVariable::with('ServerVariable')->findOrFail($id);
```
schrej: Why `ServerVariable` and not `serverVariable`? Arent the other ones lowercase? (not sure)

app/Repositories/SubuserRepository.php
```diff
@@ -170,47 +164,31 @@ public function create($sid, array $data)
...
 
-            $node = Models\Node::getByID($server->node);
-            $client = Models\Node::guzzleRequest($server->node);
-
-            $res = $client->request('PATCH', '/server', [
-                'headers' => [
-                    'X-Access-Server' => $server->uuid,
-                    'X-Access-Token' => $node->daemonSecret,
-                ],
+            $server->node->guzzleClient([
```
schrej: Not sure if I've seen that multiple times already, if true that should be an overloaded method so we don't have to write `X-Access-Server` and `X-Access-Token` every time (it is required every time right?)

app/Repositories/SubuserRepository.php
```diff
@@ -305,27 +282,20 @@ public function update($id, array $data)
...
             // if they did have them previously.
-            $node = Models\Node::getByID($server->node);
-            $client = Models\Node::guzzleRequest($server->node);
-
-            $res = $client->request('PATCH', '/server', [
-                'headers' => [
-                    'X-Access-Server' => $server->uuid,
-                    'X-Access-Token' => $node->daemonSecret,
-                ],
+            $server->node->guzzleClient([
```
schrej: I've seen them multiple times. DRY 😬

database/migrations/2017_02_02_175548_UpdateColumnNames.php
schrej: Why not the variant with the column name in the array btw?

app/Repositories/ServerRepository.php
DaneEveritt: Thats just the code standards we're using. Its kind odd, but does improve readability I think.

app/Repositories/ServiceRepository/Variable.php
DaneEveritt: Grep at work.

app/Repositories/SubuserRepository.php
DaneEveritt: I believe this was because it depends entirely on the user making the call, which isn't necessarily available to the node.

I can look at that again though.

PR #299, opened 2017-02-16
allow to use the username for login as well

schrej:
Detects if login is an email or not, validates as email or not and uses the according user table field to authenticate.


DaneEveritt:
LGTM 👍

PR #381, opened 2017-04-16
change server to server_id since that is the new name

KasperFranz:


KasperFranz:
Thank you for a quick review 👍

schrej:
Well thank you for just fixing it instead of opening an issue ;)

PR #385, opened 2017-04-16
added a check if the task has run yet

KasperFranz:
I also added the last PR I made to the changelog - if that should have been there just remove it 👍

### DIFF
--- a/resources/themes/pterodactyl/server/tasks/index.blade.php
+++ b/resources/themes/pterodactyl/server/tasks/index.blade.php
@@ -65,7 +65,13 @@
                                         <span class="label label-default">@lang('strings.no')</span>
                                     @endif
                                 </td>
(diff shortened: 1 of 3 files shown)


DaneEveritt:
Couldn't this just be a front-end modification to see if `last_run` is null?

resources/themes/pterodactyl/server/tasks/index.blade.php
```diff
@@ -65,7 +65,13 @@
                                         <span class="label label-default">@lang('strings.no')</span>
                                     @endif
                                 </td>
-                                <td class="middle">{{ Carbon::parse($task->last_run)->toDayDateTimeString() }}<br /><span class="text-muted small">({{ Carbon::parse($task->last_run)->diffForHumans() }})</span></td>
+                                <td class="middle">
+                                @if($task->last_run > null)
```
DaneEveritt: `@if($task->last_run)` should work here, or if you want to be extra verbose, `@if(! is_null($task->last_run))`

KasperFranz: I have changed that to `@if($task->last_run) ` instead

PR #388, opened 2017-04-17
show Suspended server in overview fixes #386

### LINKED ISSUE #386: SUSPENDED status not appearing in panel
I just updated today to BETA.2 and this is the first I noticed

Suspended servers will appear as ERROR and not as suspended

http://f.cgm.rs/2017-04-17_08-14-17.mp4

KasperFranz:
Another solution to this would be to change it in the javascript - I wasn't sure which you prefered so i just removed the class for dynamic update so it doesn't call the ajax status (which is hidden behind the checkServer middleware)


KasperFranz:
was just following the example of the other @if for that - changed it so they are the same as you suggested

PR #400, opened 2017-04-21
Set drop-down menu option to be same as help text below

Hwiggy:
The drop down menu option was previously "Send Power Toggle" where the help text below it was referring to it as "Send Power Option"
This could probably be changed either way, however I think "option" makes more sense than toggle.


DaneEveritt:
Please PR this into `develop` and not `v0.6.0-beta.2`.

PR #402, opened 2017-04-22
Missing 'A' in 'Allows'

Hwiggy:


KasperFranz:
I am 90% sure you don't need to escape that single quote

schrej:
We can also just edit those strings in PhraseApp btw.

Hwiggy:
It makes me feel helpful to add one character at a time

PR #404, opened 2017-04-22
TEXT/BLOB defaults not widely supported. Use STRING/VARCHAR instead.

belldandu:
Currently mysql does not support defaults for text and blobs without disabling strict mode in 5.7. https://dev.mysql.com/doc/refman/5.7/en/blob.html

Mariadb (a drop in replacement for mysql made by the same team of devs as mysql) also doesnt support defaults for text and blobs until 10.2.1 (there is no stable release, 10.2.5 is a release candidate) which isnt available on non-bleeding edge distro's. No one is going to install an unstable release of mariadb on production servers. https://mariadb.com/kb/en/mariadb/text/
Latest release for ubuntu xenial is 10.0.29-MariaDB

So for now we change text to string for values that have defaults until defaults are more widely supported for TEXT and BLOB mysql values.


schrej:
Aren't those rather short strings anyways @DaneEveritt ?

DaneEveritt:
Pretty sure I was meaning to use `string` rather than `text` here anyways, none of them should be more than 255 characters as it is.

Thanks for making this PR.

PR #475, opened 2017-06-06
Update MinecraftServiceTableSeeder.php

yoshee08:
Fixed the docker image links for all server types.


DaneEveritt:
That's not fixing anything, and breaks all of the services.

PR #538, opened 2017-07-07
Add CS:GO to Source Service Option

Polarcraft:
This will allow users to select cs:go to be installed without making a custom service.

### DIFF
--- a/database/seeds/SourceServiceTableSeeder.php
+++ b/database/seeds/SourceServiceTableSeeder.php
@@ -213,10 +213,10 @@ cd /mnt/server/steamcmd
 chown -R root:root /mnt
 
 export HOME=/mnt/server
-./steamcmd.sh +login "${STEAM_USER}" "${STEAM_PASS}" +force_install_dir /mnt/server +app_update 740 +quit > /mnt/server/log.txt 2>&1
+./steamcmd.sh +login "${STEAM_USER}" "${STEAM_PASS}" +force_install_dir /mnt/server +app_update 740 +quit
 
-mkdir -p /mnt/server/.steam/sdk32 > /mnt/server/log2.txt 2>&1
-cp -v linux32/steamclient.so ../.steam/sdk32/steamclient.so > /mnt/server/log3.txt 2>&1
+mkdir -p /mnt/server/.steam/sdk32
+cp -v linux32/steamclient.so ../.steam/sdk32/steamclient.so
 EOF;
 
 		$this->option['csgo'] = ServiceOption::updateOrCreate([
--- a/database/seeds/SourceServiceTableSeeder.php
+++ b/database/seeds/SourceServiceTableSeeder.php
@@ -213,7 +213,7 @@ cd /mnt/server/steamcmd
 chown -R root:root /mnt
 
 export HOME=/mnt/server
-./steamcmd.sh +login "${STEAM_USER}" "${STEAM_PASS}" +force_install_dir /mnt/server +app_update 740 +quit
+./steamcmd.sh +login anonymous +force_install_dir /mnt/server +app_update 740 +quit
 
 mkdir -p /mnt/server/.steam/sdk32
 cp -v linux32/steamclient.so ../.steam/sdk32/steamclient.so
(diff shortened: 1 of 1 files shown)


database/seeds/SourceServiceTableSeeder.php
```diff
@@ -192,13 +192,58 @@ private function addCoreOptions()
...
+EOF;
+
+        $this->option['csgo'] = ServiceOption::updateOrCreate([
+            'service_id' => $this->service->id,
+            'tag' => 'csgo',
+        ], [
+            'name' => 'Counter-Strike: Global Offensive',
+            'description' => 'Counter-Strike: Global Offensive is a multiplayer first-person shooter video game developed by Hidden Path Entertainment and Valve Corporation.',
+            'docker_image' => 'quay.io/pterodactyl/core:source',
+            'config_startup' => '{ "done": "VAC secure mode is activated.", "userInteraction": []}',
```
DaneEveritt: Please remove the extra space between `{` and the first `"`

database/seeds/SourceServiceTableSeeder.php
```diff
@@ -319,6 +364,33 @@ private function addArkVariables()
...
 
+    private function addCSGOVariables()
+    {
+        ServiceVariable::updateOrCreate([
+            'option_id' => $this->option['csgo']->id,
+            'env_variable' => 'SRCDS_MAP',
+        ], [
+            'name' => 'Map',
+            'description' => 'The default map for the server.',
+            'default_value' => '',
```
DaneEveritt: Is there a map that is always there in the game that can be defaulted to to avoid confusion on this field?

database/seeds/SourceServiceTableSeeder.php
```diff
@@ -319,6 +364,33 @@ private function addArkVariables()
...
+            'user_viewable' => 1,
+            'user_editable' => 1,
+            'rules' => 'required',
```
DaneEveritt: Most likely this should be `required|string|alpha_dash` to allow `map_name` and similar. Are there names that would violate these rules?

database/seeds/SourceServiceTableSeeder.php
```diff
@@ -319,6 +364,33 @@ private function addArkVariables()
...
+        ServiceVariable::updateOrCreate([
+            'option_id' => $this->option['csgo']->id,
+            'env_variable' => 'STEAM_ACC',
+        ], [
+            'name' => 'Steam Account Token',
+            'description' => 'The Steam Account Token required for the server to be displayed publicly.',
+            'default_value' => '',
+            'user_viewable' => 1,
+            'user_editable' => 1,
+            'rules' => 'required',
```
DaneEveritt: Rule should probably be `required|string|alpha_num|size:32` since those token lengths don't change as far as I am aware. Check the size to be sure what the number of characters is though.

database/seeds/SourceServiceTableSeeder.php
DoNotSpamPls: Dust 2?

database/seeds/SourceServiceTableSeeder.php
Polarcraft: I never noticed that, but I never had an issue with it erroring out. Will change it now.

database/seeds/SourceServiceTableSeeder.php
Polarcraft: I can set it to default to dust2 for example.

database/seeds/SourceServiceTableSeeder.php
Polarcraft: Yes, the max size is 32 and that will work perfectly.

database/seeds/SourceServiceTableSeeder.php
Polarcraft: I don't think there are maps that would violate these rules as they are specified as "de_dust2" or "jb_august".

DaneEveritt: I'd use those rules then, and if there is an issue we can always release an update later to fix it.

Issue #535, opened 2017-07-04
Redirect to the allocations tab after adding a node


parkervcp:
After a node is created and before the user is directed to the node core.json they should be directed to the allocations tab.

They could also get a pop-up notifying them a node is missing allocations.

PR #540, opened 2017-07-08
Fix node creation page not displaying old values and change redirect after creation

OrangeJuiced:
The node creation page now displays the old values upon failing to properly create a node. The second commit changes the redirect to the node allocations page as described in issue #535.


DaneEveritt:
Looks great, thanks for this 👍

PR #543, opened 2017-07-09
Added a Forge Service Option

Polarcraft:
This will allow users to select Forge as a service option.

### DIFF
--- a/database/seeds/MinecraftServiceTableSeeder.php
+++ b/database/seeds/MinecraftServiceTableSeeder.php
@@ -243,6 +243,41 @@ EOF;
             'startup' => null,
             'script_install' => $script,
         ]);
+
+        $script = <<<'EOF'
+#!/bin/ash
+# Forge Installation Script
+#
+# Server Files: /mnt/server
+apk update
+apk add curl openjdk8
+
+FUCK_FORGE_NOT_HAVING_A_GOOD_WAY_TO_GET_VERSIONS=$(curl -sl http://files.minecraftforge.net/maven/net/minecraftforge/forge/ | grep -A1 Latest | grep  -o -e '[1]\.[0-9][0-9] - [0-9][0-9]\.[0-9][0-9]\.[0-9]\.[0-9][0-9][0-9][0-9]')
+LATEST_VERSION=$(echo $FUCK_FORGE_NOT_HAVING_A_GOOD_WAY_TO_GET_VERSIONS | sed 's/ //g')
+
+cd /mnt/server
+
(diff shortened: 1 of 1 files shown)


database/seeds/MinecraftServiceTableSeeder.php
```diff
@@ -243,6 +243,41 @@ private function addCoreOptions()
...
+
+        $script = <<<'EOF'
+#!/bin/ash
+# Forge Installation Script
+#
+# Server Files: /mnt/server
+apk update
+apk add curl openjdk8
+
+FUCK_FORGE_NOT_HAVING_A_GOOD_WAY_TO_GET_VERSIONS=$(curl -sl http://files.minecraftforge.net/maven/net/minecraftforge/forge/ | grep -A1 Latest | grep  -o -e '[1]\.[0-9][0-9] - [0-9][0-9]\.[0-9][0-9]\.[0-9]\.[0-9][0-9][0-9][0-9]')
```
DaneEveritt: While amusing, this needs to be renamed.

database/seeds/MinecraftServiceTableSeeder.php
```diff
@@ -243,6 +243,41 @@ private function addCoreOptions()
...
             'script_install' => $script,
         ]);
+
+        $script = <<<'EOF'
+#!/bin/ash
+# Forge Installation Script
+#
+# Server Files: /mnt/server
+apk update
+apk add curl openjdk8
```
DaneEveritt: Why not use the existing java docker containers that are out there and likely already on the user's system?

database/seeds/MinecraftServiceTableSeeder.php
Polarcraft: That was originally made by parker, so you can see why that is there. I will check it to get_versions though.

database/seeds/MinecraftServiceTableSeeder.php
Polarcraft: In order to install the forge files I have to use that to execute `java -jar` in the setup.

DaneEveritt: Yes, but there are openjdk based containers out there that could be used rather than a generic `alpine` container, that way you don't have to wait on installation.

Plus I believe they are stored on the user's system since other core containers rely on them as a base.

Polarcraft: Alright, I have changed it so the installation container is set to an alpine openjdk8 container.

PR #544, opened 2017-07-10
Add multiple file/directory deletion in the filemanager

OrangeJuiced:
This pull request contains the code for multiple file/directory deletion in the file manager. I added a selection button per item in file manager. The user can delete the selected items using the delete button in the top right.

It looks like this:


public/themes/pterodactyl/js/frontend/files/src/actions.js
```diff
@@ -19,6 +19,9 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
+var files = [];
```
DaneEveritt: These are ES6 based files, please use `let` or `const` rather than `var`.

public/themes/pterodactyl/js/frontend/files/src/actions.js
```diff
@@ -316,6 +319,76 @@ class ActionsClass {
...
+
+    }
+
+    deleteSelected() {
+      var nameBlock;
+
+      swal({
+          type: 'warning',
+          title: '',
+          text: 'Are you sure you want to delete <code>' +
```
DaneEveritt: I think this would render better if each file being deleted was wrapped in `<code>` tags, rather than all wrapped together in a single tag.

public/themes/pterodactyl/js/frontend/files/src/actions.js
```diff
@@ -316,6 +319,76 @@ class ActionsClass {
...
+                + '</code>? There is <strong>no</strong> reversing this action.',
+          html: true,
+          showCancelButton: true,
+          showConfirmButton: true,
+          closeOnConfirm: false,
+          showLoaderOnConfirm: true
+      }, () => {
+          $.ajax({
+              type: 'DELETE',
+              url: `${Pterodactyl.node.scheme}://${Pterodactyl.node.fqdn}:${Pterodactyl.node.daemonListen}/server/file/f/${files}`,
```
DaneEveritt: I'm wondering if we shouldn't have a different endpoint for mass deletions to pass data as a JSON object, rather than a URL argument.

resources/themes/pterodactyl/server/files/list.blade.php
```diff
@@ -85,7 +88,7 @@
                             {{ $carbon->diffForHumans() }}
                         @endif
                     </td>
-                    <td><button class="btn btn-xxs btn-default disable-menu-hide" data-action="toggleMenu" style="padding:2px 6px 0px;"><i class="fa fa-ellipsis-h disable-menu-hide"></i></button></td>
+                    <td><button class="btn btn-xxs btn-default disable-menu-hide" data-action="toggleMenu" style="padding:2px 6px 0px;"><i class="fa fa-ellipsis-h disable-menu-hide"></i></button><button class="btn btn-xxs btn-default disable-menu-hide" data-action="addToList" style="padding:2px 6px 0px;"><i class="fa fa-plus disable-menu-hide"></i></button></td>
```
DaneEveritt: I'm thinking we should make these checkboxes and have a mass-options dropdown menu that for now only contains delete, but could down the road handle moving and such.

ET-Bent:
selection_893

I would rather suggest an approach seen in the image where the checkboxes are either beneath the items or to the left of the items with one checkbox on top that allows to select all. 

This sample image is taken from Nextcloud

public/themes/pterodactyl/js/frontend/files/src/actions.js
```diff
@@ -19,6 +19,9 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
+var files = [];
+var fileselements = [];
```
schrej: this should be camelCase aka `filesElements`

resources/themes/pterodactyl/server/files/list.blade.php
schrej: If you do that, you don't have to manage selections in javascript as well, just find all the selected checkboxes.

public/themes/pterodactyl/js/frontend/files/src/actions.js
schrej: This name is also rather unclear, maybe use `selectedFiles`.

ET-Bent:
@schrej can you post a screenshot of that in action in your dev setup?

schrej:
@ET-Bent just get yourself some vagrant ;)

OrangeJuiced:
The feature should now be more in line with the suggestions.

DaneEveritt:
Looks great! Can we make it so left clicking a row selects the file checkbox?

schrej:
Actually, is the mass actions button disabled when no file is selected? If not, why not?

OrangeJuiced:
That should do it. Both suggestions are implemented.

PR #583, opened 2017-08-14
API Get Console

Dieterfu:
Hello,
it would be very nice If you could add add a endpoint to get the console Output a specific Minecraft Server.

With best regards
Dieter Fugger


DaneEveritt:
Please do not open Pull Requests that do not change anything. Create an issue instead, and please actually read through the issue list and API code before requesting these things as they already exist.

PR #584, opened 2017-08-14
API add cpu and ram usage Endpoint

Dieterfu:
Hello,
would it be possible to get the current cpu and ram usage of a Minecraft Server by a API endpoint?

Thank you
With best regards
Dieter


DaneEveritt:
Please do not open Pull Requests that do not change anything. Create an issue instead, and please actually read through the issue list and API code before requesting these things as they already exist.

PR #604, opened 2017-08-26
Admin permissions should only be visible for Admin users fixes #580

### LINKED ISSUE #580: Admin-only options are shown on the "Create API Key" page
When logged in as a standard (non-admin) user, I noticed while creating an API key that I had a bunch of administrator options (such as create server, add node, etc). These options shouldn't be available to a non-admin user.

KasperFranz:


DaneEveritt:
Hey, resolved this in https://github.com/Pterodactyl/Panel/commit/cb62e6a96da6f02cd2e692108bed3c4601661229

Thanks for the PR though! 👍

PR #615, opened 2017-09-05
Feature/customized theme

schrej:
Uses the new colors across the AdminLTE theme with a custom build of AdminLTE version 2.4. Maybe we should create a fork of the AdminLTE repo with our colors, it's somewhat overkill though...

Also includes a revamped login page with particles-js in the background as it is fancy.

Bonus: Fixed user creation command


DaneEveritt:
Can you please update the ReadMe to reflect this new package we use as well? :)

PR #617, opened 2017-09-06
Feature/dockerfile

schrej:
This pull request adds a docker image for the panel. The `.dev/docker/` folder contains a Readme and an example docker-compose file.

I've tested the image a little bit, it works fine so far.


DaneEveritt:
The `bootstrap/cache/services.php` file should be ignored, not sure whats going on there.

Are there no PHP7 docker images out there based on alpine that we could use as a base? It should also be noted that everything passed in as an environment variable will override the .env file, so if you're passing in blanks in the docker-compose file you'll never be able to change them without re-upping the image.

schrej:
I guess there are php7 docker images, this is pretty easy though and we can install only the requirements we have, nothing more.
Regarding the environment variables: We could also rename the docker environment variable parameters and only use them on first setup. Also users need to persist the .env file either way. And it'll get confusing if they cant modify it. We could also require to run the `pterodactyl:env`, `pterodactyl:mail` and `migrate` commands manually, like the user command.

parkervcp:
From my experience the php7 base images are missing stuff needed.

DaneEveritt:
I'm going to hold off on merging this until develop is more stable as the commands are changing.

lancepioch:
A little late, but you should not commit `/bootstrap/cache/services.php`

DoNotSpamPls:
Any update on this? 🤔

schrej:
We wanted to look at it again, especially as 0.7 is released now. There are a few things left to do, most importantly: We have to get rid of Caddy because of their new licensing. We can either use a fork if there is a proper one already or switch back to nginx which is a little more complicated but should work fine.

parkervcp:
@schrej There is https://github.com/WedgeServer/wedge but it is slightly behind the main release.

We could script certbot with nginx in the container but I think that could be unnecessary bloat. Volumes would also need to be on point for this approach.

Nols1000:
@schrej I think it would also be possible to build caddy in the Dockerfile an remove the source + build dependencies after the build. It wouldn't have an impact on the image size and from my understanding it would solve the licensing problem.

schrej:
Well it seems like they are building it themselves which allows us (and anyone) to use it for free. They also removed the sponsor header they added when starting to sell their own binaries so it should be possible to continue to use Caddy.
Nevertheless a fully open source Webserver might be better if we find a good way to automatically renew LetsEncrypt certificates.

Nols1000:
I just checked out the feature and I couldn't build the image because the XML extension is not installed. 

I think you should also consider using the official php image instead of the alpine as base image as it supports php7.2 and alpine doesn't.

parkervcp:
@Nols1000 The stock php image is missing things we need for the panel. That is why we rolled our own. Also if you use the docker compose script it spins up it's own docker network. That would keep the panel isolated from other containers on the host.

Nols1000:
@parkervcp What is it missing? You can install almost any extension with `docker-php-ext-install`. And even if you cant install your missing dependencies this way it's build upon debian/alpine. I think it would be smart to use the official image to avoid problems in the case of a php update (e.g. php7.2 is not available on alpine yet).

quikstrike:
Two names I'd like to throw in for a possible replacement to Caddy are:

- [Kong](https://getkong.org/)
- [traefik](https://traefik.io/)

Kong is nice since it's basically nginx with some really nice quality of life features, and traefik is a good open source http server that makes working with a http server in docker a lot nicer.

schrej:
traefik as a http reverse proxy/load balancer. While it is very awesome what we need is an actual http web server that can connect to php fpm
Kong is an api gateway, also not a webserver as far as I can see. Imo the only options are Caddy (which should be fine if alpine is building itself) or nginx

Nols1000:
@parkervcp the `docker-compose.yml` is broken. The panel complains about not finding `tcp://cache_1:6379`.
````
php_network_getaddresses: getaddrinfo failed: Name does not resolve [tcp://cache_1:6379]
````
Please check #622 again. Looks like docker became smarter over time ^^.

schrej:
@Nols1000 does it work when you use `cache` as the redis host instead?

Nols1000:
@schrej the docker-compose file should work. But the Dockerfile is missing php7-simplexml. And don't get me started on updating to 0.7.x

parkervcp:
I am going to probably be pushing to this branch here soon for things like the docker-compose file and some other fixes.

Nols1000:
@schrej I found this /usr/sbin/php-fpm7 -c /etc/php7 did you meant to use -y instead of -c to configure the fpm-config file?

parkervcp:
I have found a php7.2 repo for alpine. I will be looking into it.

sachk:
You could use some of the images [here](https://github.com/codecasts/php-alpine), I was experimenting with this as well and the php component of these seemed to work fine. Alternatively, the edge branch of alpine contains PHP 7.2. Have you taken a look at [these](https://github.com/CraftDock/pterodactyl) yet? Whoever's working on them seems to be doing a good job.

DaneEveritt:
Closing in favor of #1117 which implements all of these commits as well 👍

PR #629, opened 2017-09-16
Added SpongeForge and Nukkit. Renamed SpongeVanilla variables.

ScarletWolf275:
I have renamed the variables so they don't conflict with SpongeForge


DaneEveritt:
Just an update for ya, theres going to be changes to the seeder setup, I'll comment here when they are complete. At that point you can update this service to match those changes, and then we should be good.

Thanks again for your patience :)

ScarletWolf275:
Alrighty

DaneEveritt:
So this will actually be handled by service exports. Hold onto your configuration and when 0.7 rolls around you'll be able to export and share this will anyone who would like to use it. 👍

Issue #546, opened 2017-07-10
Use SweetAlert for logout message


Derkades:
![image](https://user-images.githubusercontent.com/15892014/28031698-6b927afe-65a8-11e7-8adf-6afa0d806441.png)
Should be a very simple thing to do

KasperFranz:
Why is it you want that?  - it is very rare to see that on a panel or any software

schrej:
@KasperFranz It appears whenever you click the Logout button.

Derkades:
The panel uses SweetAlert for all other alerts, so it makes sense to use it for the logout alert as well

DaneEveritt:
closed by #634

PR #634, opened 2017-09-21
Fix issue #546 by adding swal to logouts

lancepioch:
## Before:
image

## After:
image


schrej:
I seriously wanted to do that ages ago but always forgot, thank you 🎉

Issue #530, opened 2017-07-01
Feature Request: Force 2FA on Users


Qasimmodo:
Would be nice if you could force all registered users on the panel to enable 2FA. Just to increase security that one bit more.

Derkades:
And an option to force administrators only

PR #635, opened 2017-09-21
Require 2FA for Users

lancepioch:
Reference: #530

Currently you can require 2FA for:

* Nobody
* Admins
* Users
* Everybody

* [x] Make sure exception route list is not missing any routes.
* [x] Make sure design styling for radio buttons is solid
* [x] Discuss use of clever switch statement... it might be better to not have that.
* [x] Discuss use of constants for the settings, should it have its own class??
* [x] No Tests!


app/Http/Middleware/RequireTwoFactorAuthentication.php
```diff
@@ -0,0 +1,98 @@
...
+
+namespace Pterodactyl\Http\Middleware;
+
+use Closure;
+use Krucas\Settings\Settings;
+use Prologue\Alerts\AlertsMessageBag;
+
+class RequireTwoFactorAuthentication
+{
+    const NOBODY = 0;
```
DaneEveritt: I think these would work better as `LEVEL_NONE`, `LEVEL_ADMIN`, `LEVEL_ALL`.

Not entirely sure what the difference between users and everybody would be. Seems like a very strange edge-case to not require 2FA for admins, but require it for normal users.

app/Http/Middleware/RequireTwoFactorAuthentication.php
```diff
@@ -0,0 +1,98 @@
+<?php
+
```
DaneEveritt: All files belonging to the project that aren't Laravel specific should have the license header, should be able to just copy it out of an existing one. 👍

app/Http/Middleware/RequireTwoFactorAuthentication.php
```diff
@@ -0,0 +1,98 @@
...
+     */
+    protected $settings;
+
+    /**
+     * RequireTwoFactorAuthentication constructor.
+     *
+     * @param \Prologue\Alerts\AlertsMessageBag $alert
+     * @param \Krucas\Settings\Settings         $settings
+     */
+    public function __construct(Settings $settings, AlertsMessageBag $alert)
```
DaneEveritt: This is undocumented, so my fault here, but I've been using an Alphabetical scheme for constructor classes and ordering the protected/public variables. So this should be `__construct(AlertsMessageBag $alert, Settings $settings)` and then `$this->alert` before settings.

app/Http/Middleware/RequireTwoFactorAuthentication.php
```diff
@@ -0,0 +1,98 @@
...
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        // Ignore non-users
+        if (! auth()->check()) {
```
DaneEveritt: Can we dep. inject the guard, or just get the user from the request using `$this->user()`?

app/Http/Middleware/RequireTwoFactorAuthentication.php
```diff
@@ -0,0 +1,98 @@
...
+            return $next($request);
+        }
+
+        // All TOTP related routes
+        $validRoutes = [
```
DaneEveritt: This variable name is a bit confusing, maybe `$ignoreRoutes` or do what Laravel often does for things and have `protected $ignore = [...routes];`.

app/Http/Middleware/RequireTwoFactorAuthentication.php
```diff
@@ -0,0 +1,98 @@
...
+
+        // Only allow 2FA pages
+        if (in_array($request->route()->getName(), $validRoutes)) {
+            return $next($request);
+        }
+
+        // Get the setting
+        $tfa = (int) $this->settings->get('2fa', 0);
+
+        switch ($tfa) {
```
DaneEveritt: `switch((int) $this->settings->get('2fa', 0)) {` would be a little cleaner here since we don't use that value anywhere else.

app/Http/Middleware/RequireTwoFactorAuthentication.php
```diff
@@ -0,0 +1,98 @@
...
+                return $next($request);
+
+            case self::ADMINISTRATORS:
+                if (! $request->user()->root_admin) {
+                    return $next($request);
+                }
+                break;
+
+            case self::USERS:
+                if ($request->user()->accessLevel === 'subuser') {
```
DaneEveritt: Subusers are just normal users on this panel, it just refers to their ability to access a specific server as a non-server-owner.

routes/base.php
```diff
@@ -69,7 +69,7 @@
 
     Route::put('/totp', 'SecurityController@generateTotp')->name('account.security.totp');
 
-    Route::post('/totp', 'SecurityController@setTotp');
+    Route::post('/totp', 'SecurityController@setTotp')->name('account.security.totp.set');
```
DaneEveritt: Any reason we need to change the route names? I generally keep them the same if the get/post routes are the same.

app/Http/Middleware/RequireTwoFactorAuthentication.php
lancepioch: Yup, I'll make that change.

app/Http/Middleware/RequireTwoFactorAuthentication.php
lancepioch: Not a problem.

app/Http/Middleware/RequireTwoFactorAuthentication.php
lancepioch: None = nobody obviously
Admin = administrators only
Users = admins + normal users only
All = admins + normal users + sub users + anybody else logged in

How does that sound?

app/Http/Middleware/RequireTwoFactorAuthentication.php
lancepioch: Will do.

app/Http/Middleware/RequireTwoFactorAuthentication.php
lancepioch: `protected $ignore = []` sounds good to me

app/Http/Middleware/RequireTwoFactorAuthentication.php
lancepioch: Yeah I was thinking that the other day.

routes/base.php
lancepioch: Without that explicitly, the route doesn't have a name, if I remove it, then this is what I get:

image

Which means that I can't reference it via any route name so that I can ignore it in the middleware.

app/Http/Middleware/RequireTwoFactorAuthentication.php
lancepioch: Ah alright, I misunderstood then.

routes/base.php
DaneEveritt: Ah, this is fine then for this PR. I think we could make an issue to update all of the route names down the road to be a little more clear. Something using the request type potentiall, not sure if `delete.account.api`/`post.account.api` would make sense, or like `account.delete.api.revoke`?

lancepioch:
I've made the requested changes.

DaneEveritt:
@lancepioch I think you need to merge develop in to get builds to pass.

DaneEveritt:
@lancepioch is this now complete?

lancepioch:
Yes

PR #639, opened 2017-09-21
Automatically register eloquent events

lancepioch:
Make sure those events are registered!


DaneEveritt:
So I took another look at this, and I believe this is being handled by the `AppServiceProvider.php` file which is defining observers for these models.

lancepioch:
Observers yo

PR #647, opened 2017-09-26
Temporary routes fix because these controllers don't exist anymore…

lancepioch:
…apparently ¯\_(ツ)_/¯

This allows you to run artisan route:list and artisan route:cache


DaneEveritt:
This will be fixed shortly when I actually make that route work...

PR #659, opened 2017-10-04
updated vagrant provision to be using the new namespace

KasperFranz:


DaneEveritt:
This doesn't need a changlog update as the fix is expected here 👍

KasperFranz:
removed it

PR #672, opened 2017-10-08
Make config caching less aggressive

arcdigital:


DaneEveritt:
You missed a call in `EnvironmentSettingsCommandTest.php`

PR #680, opened 2017-10-09
Redirect to users list on user deletion

### LINKED ISSUE #675: [0.7] Deleting user redirects to old user page
Deleting a user will redirect you back to the URL you came from, which no longer exists, thus causing a 404.

arioioio:
fixes #675


lancepioch:
@DaneEveritt This is good now.

schrej:
Yeah I would've merged this but I don't seem to be allowed to anymore 😭

PR #681, opened 2017-10-09
Feature/fix terminal notification

schrej:
Fixes the terminal notification behavior so it only shows the notification when there is unseen output compared to at all times when not scrolled down.


public/themes/pterodactyl/js/frontend/console.js
```diff
@@ -149,10 +151,13 @@ function pushToTerminal(string) {
 
     if (TerminalQueue.length > 0) {
         for (var i = 0; i < CONSOLE_PUSH_COUNT && TerminalQueue.length > 0; i++) {
+            const scrolledDown = isTerminalScrolledDown();
```
DaneEveritt: This should't be a `const` since this isn't a file that is transpired into ES5. In addition, why not just call `if (isTerminalScrolledDown()) below for scrolling down. Its a bit inconsistent here.

schrej: My thought probably was that after a line is added the terminal isn't scrolled down anymore. I'll also move that part outside the loop.

public/themes/pterodactyl/js/frontend/console.js
```diff
@@ -148,16 +150,20 @@ function pushToTerminal(string) {
...
 
-            if (! $scrollNotify.is(':visible')) {
-                window.scrollToBottom();
-            }
-
             window.ConsoleElements++;
             TerminalQueue.shift();
         }
+        
+        if (scrolledDown) {
```
DaneEveritt: This is still inconsistent. It should be `if (isTerminalScrolledDown())` and then remove the `var scrolledDown` since it is unnecessary.

schrej: As I said, once new lines are added, the console isn't scrolled down anymore. Therefore the check before adding them, and then scroll down after all are added.

PR #701, opened 2017-10-21
Add database/startup/allocation management to server front-end views

DaneEveritt:


app/Services/Databases/DatabaseManagementService.php
```diff
@@ -95,7 +96,7 @@ public function create($server, array $data)
             $this->database->commit();
         } catch (\Exception $ex) {
             try {
-                if (isset($database)) {
+                if (isset($database) && $database instanceof Database) {
```
lancepioch: Don't need `isset`.

DaneEveritt:
@lancepioch this is ready for a quick skim whenever you're free.

PR #703, opened 2017-10-23
Implement Panel changes to support internal SFTP subsystem on Daemon

DaneEveritt:
Panel changes required by https://github.com/Pterodactyl/Daemon/pull/51


app/Http/Controllers/API/Remote/SftpController.php
```diff
@@ -0,0 +1,86 @@
...
+            );
+
+            $this->clearLoginAttempts($request);
+        } catch (AuthenticationException $exception) {
+            return response()->json([
+                'error' => 'Invalid credentials.',
+            ], 403);
+        } catch (RecordNotFoundException $exception) {
+            return response()->json([
+                'error' => 'Invalid server.',
```
lancepioch: Probably should trans these errors.

app/Http/Controllers/API/Remote/SftpController.php
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Get the throttle key for the given request.
+     *
+     * @param \Illuminate\Http\Request $request
+     * @return string
+     */
+    protected function throttleKey(Request $request)
+    {
+        return Str::lower(array_get(explode('.', $request->input('username')), 0) . '|' . $request->ip());
```
lancepioch: too cool for `strtolower` but you're fine with `array_get` and `explode` 😛

resources/themes/pterodactyl/server/settings/sftp.blade.php
```diff
@@ -66,20 +36,12 @@
                 <div class="form-group">
                     <label for="password" class="control-label">@lang('strings.username')</label>
                     <div>
-                        <input type="text" class="form-control" readonly value="{{ $server->username }}" />
+                        <input type="text" class="form-control" readonly value="{{ auth()->user()->username }}.{{ $server->uuidShort }}" />
```
lancepioch: 👍 Now you're getting into the Laravel spirit!

app/Http/Controllers/API/Remote/SftpController.php
DaneEveritt: Don't tell anyone, I just copy/pasted this out of the trait to modify the username attr.

app/Http/Controllers/API/Remote/SftpController.php
DaneEveritt: Good call. Users will never see this anyways (at least currently).

Issue #706, opened 2017-10-27
Performance Problems with JS Console


kaganus:
* Panel or Daemon: Panel
* Version of Panel: 0.6.4
* Server's OS: Ubuntu 16.04
* Your Computer's OS & Browser: Windows 10 PC with Google Chrome, Mozilla Firefox, Opera and Windows 7 laptop with Google Chrome
* uname -a: Linux 4.4.0-97-generic # 120-Ubuntu SMP Tue Sep 19 17:28:18 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
* php -v: PHP 7.0.22-0ubuntu0.16.04.1 (cli) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with the ionCube PHP Loader (enabled) + Intrusion Protection from ioncube24.com (unconfigured) v10.0.3, Copyright (c) 2002-2017, by ionCube Ltd.
    with Zend OPcache v7.0.22-0ubuntu0.16.04.1, Copyright (c) 1999-2017, by Zend Technologies

## Details:

There are a few problems with the new custom made js console since 0.6.4. Talked about this before on discord with Dane, and he confirmed there was a lack of optimization in console code.

~~Problem # 1: Whole page is frozen/unresponsive until the console fully loads. 
Every time a user clicks on one of his servers, as the server console page is the default page to be shown, browser tab becomes unresponsive for a few seconds (depending on the size of the log) until console is loaded up with all the logs and is rendered. Cannot click on other links on the menu or buttons while browser is in this unresponsive state. (Happens on Chrome, Firefox, Opera)~~

~~Problem # 2: Text selection does not work correctly
When you want to use your mouse cursor to select some text from the console log, the selection disappears as soon as you release the mouse button. It is the same with double clicking on a word in console, the selection is canceled immediately. To copy text, you have to click CTRL + C without releasing the mouse left button as it cancels the selection.~~

~~On Firefox, if you select text on multiple lines, the selection is not cancelled and it works correctly, although if you select text from a single line, selection is cancelled immediately. On Chrome and Opera, it doesn't matter if you select text from a single line or multiple lines, it is cancelled immediately after mouse button release.~~

Problem # 3: Command history sometimes do not work
[...]

PR #709, opened 2017-10-31
Fix highlighting problem in the console.

jasoryeh:
I kept the on click in case someone needed it. 
Problem 2 on #706

### DIFF
--- a/public/themes/pterodactyl/js/frontend/console.js
+++ b/public/themes/pterodactyl/js/frontend/console.js
@@ -47,7 +47,7 @@ $(document).ready(function () {
 
     $terminalInput.focus();
     $('.terminal_input--prompt, #terminal_input, #terminal, #terminalNotify').on('click', function () {
-        $terminalInput.focus();
+        //$terminalInput.focus();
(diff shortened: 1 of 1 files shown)


DaneEveritt:
This should just remove `#terminal` from the list of elements to focus on, rather than disabling all of them.

public/themes/pterodactyl/js/frontend/console.js
```diff
@@ -47,7 +47,7 @@ $(document).ready(function () {
 
     $terminalInput.focus();
     $('.terminal_input--prompt, #terminal_input, #terminal, #terminalNotify').on('click', function () {
```
lancepioch: Remove the whole statement instead of commenting it out.

DaneEveritt: @lancepioch see my comment in the actual PR about just removing `#terminal` from the list.

jasoryeh: Removing #terminal should work.

PR #715, opened 2017-11-02
Fix for js console causing browser to become unresponsive

kaganus:
think it's obvious what the problem was :)
problem 1 on #706

### DIFF
--- a/public/themes/pterodactyl/js/frontend/console.js
+++ b/public/themes/pterodactyl/js/frontend/console.js
@@ -198,8 +198,8 @@ function pushToTerminal(string) {
         $('#terminal').html('');
         data.split(/\n/g).forEach(function (item) {
             pushToTerminal(item);
-            window.scrollToBottom();
         });
+        window.scrollToBottom();
     });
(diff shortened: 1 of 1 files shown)


DaneEveritt:
Doesn't this introduce the possibility of the console sitting at the top for awhile if loading a large enough file? Should we be pushing to the bottom every X lines rather than every line?

I honestly can't remember how many I load at once here, but I'm guessing its not that many so this probably isn't a concern.

schrej:
Well that's better than lagging the whole website. Also adding lines outside of the viewport of the terminal is probably faster than when they are shown all the time by scrolling down

kaganus:
I have a few busy servers with a lot of lines of console log, and when test with this change, i do not ever see the console at the top. Thing is, i don't think lines are rendered one by one or by a buffer size. Initial logs are rendered after every line is pushed and console is scrolled to bottom. So just having a single scrollToBottom at the end will be the best for performance, speed, and also will never cause the console to be stuck at the top

PR #719, opened 2017-11-04
German Translation

DRSchlaubi:
- Added German Translation
- Fixed Syntax mistakes in previous version


resources/lang/de/admin/nests.php
```diff
@@ -0,0 +1,33 @@
+<?php
```
DaneEveritt: Its not documented, but all of the license headers can be removed from these files. We're moving towards that on all files, I just haven't done a gloabl search and replace yet.

So lines 2 - 9 can be removed.

resources/lang/de/admin/nests.php
```diff
@@ -0,0 +1,33 @@
...
+/**
```
DaneEveritt: Its not documented, but all of the license headers can be removed from these files. We're moving towards that on all files, I just haven't done a gloabl search and replace yet.

So lines 2 - 9 can be removed.

DRSchlaubi: I'll remove that

DaneEveritt:
The license header should be removed from all of the files. Also any language files you have where you have no changes (e.g. the validation messages file) can be left out, it'll just default to English.

DRSchlaubi:
I have changed that

schrej:
The next time you merge German translation let me have a look first :sweat_smile:

PR #740, opened 2017-11-07
Set default cache driver to memcached in .env

DoNotSpamPls:
Using redis as a default in that file makes `composer install --no-dev` fail if a redis server doesn't exist. Making a Redis server is not touched on in the documentation, so this would leave users confused.


schrej:
we should rather change the documentation and replace `memcached` with `redis-server`.

DoNotSpamPls:
@schrej Redis requires additional setup, should I edit the docs adding a page for setting up Redis and changing `memcached` to `redis-server`?

DaneEveritt:
This wouldn't have addressed the core problem anyways. `.env` likely needs to be changed to use an array, or have no value here.

PR #754, opened 2017-11-12
Replace magic numbers with constants

lancepioch:
Replace magic numbers with constants


app/Services/Schedules/Tasks/TaskCreationService.php
```diff
@@ -16,6 +16,8 @@
 
 class TaskCreationService
 {
+    const MAX_INTERVAL_TIME_SECONDS = 90;
```
DaneEveritt: Forgot a 0

PR #768, opened 2017-11-18
Push updates to login page, mostly UI enhancements.

DaneEveritt:
In addition to the UI changes below, this change also includes improvements to the way 2FA logins were handled. Previously the user's password and username were encrypted and then passed along to the next page by storing them in the cache and flashing the cache key to the view for submission.

I have changed this method to only store the user ID and wether or not the username and password that were provided are valid. If they are not, the user is still sent to the 2FA form as it is considered a part of their username and password combination. At that point in the process it will determine if their credentials are valid, and if not fail with the same error message as before. If they are valid, it will check that the token provided is valid and then log them in using the user model that is retrieved from the database.

This avoids storing their password in the cache (even if it was encrypted) since that felt very count-intuitive and counter to the security practices of this software.

The UI enhancements below I made because I personally did not like the white box previously, and the box was getting shoved super low on the screen. I know transparent isn't everyones favorite, but I think this looks quite sleek, and can always be adjusted per-installation if people want. The CSS is now a bit cleaner, and the HTML is more "bootstrappy" so it should render better on all screens.

http://i.kumquat.io/0P152k2x1V1H

http://i.kumquat.io/1S2L1q3s092K


schrej:
"enhancements" :stuck_out_tongue:

PR #771, opened 2017-11-19
Change how API keys are validated

DaneEveritt:
Removes the need for HMAC authentication against the API and moves to using Bearer token. This will ultimately help to unify even more of the API (since the Daemon is using this method for almost all calls).


app/Http/Middleware/API/AuthenticateIPAccess.php
```diff
@@ -0,0 +1,39 @@
...
+    public function handle(Request $request, Closure $next)
+    {
+        $model = $request->attributes->get('api_key');
+
+        if (is_null($model->allowed_ips) || empty($model->allowed_ips)) {
+            return $next($request);
+        }
+
+        foreach ($model->allowed_ips as $ip) {
+            if (Range::parse($ip)->contains(new IP($request->ip()))) {
```
lancepioch: This should only be called once:

    new IP($request->ip())

app/Providers/RouteServiceProvider.php
```diff
@@ -29,13 +29,9 @@ public function boot()
...
     public function map()
     {
-        Route::middleware(['api'])->prefix('/api/user')
-              ->namespace($this->namespace . '\API\User')
-              ->group(base_path('routes/api.php'));
-
-        Route::middleware(['api'])->prefix('/api/admin')
-              ->namespace($this->namespace . '\API\Admin')
-              ->group(base_path('routes/api-admin.php'));
+//        Route::middleware(['api'])->prefix('/api/user')
```
lancepioch: Remove instead of commenting out

DaneEveritt: Theres a second PR coming in behind this to implement all of those routes.

resources/lang/en/base.php
```diff
@@ -33,7 +33,7 @@
             'header_sub' => 'Manage your API access keys.',
             'list' => 'API Keys',
             'create_new' => 'Create New API key',
-            'keypair_created' => 'An API Key-Pair has been generated. Your API secret token is <code>:token</code>. Please take note of this key as it will not be displayed again.',
+            'keypair_created' => 'An API Key-Pair has been generated and is listed below.',
```
schrej: Isn't it just an API Token, not a pair?

DaneEveritt: Yeah, good catch!

Issue #758, opened 2017-11-14
CPU Graph Doesn't Scale by CPU Limit [+ Suggested Fix]


jcsnider:
* Panel or Daemon: Panel
* Version of Panel/Daemon: All (Suggested fix is for the latest development branch)
* Server's OS: Centos 7 (But this bug affects all)
* Your Computer's OS & Browser: Win10 + Chrome

## Add Details Below:
This affects all versions of the Panel... I fixed it in my copy of 0.6.4. I'll include a suggested fix in this issue report.

Just giving an example, let's say I have a server with a CPU limit of 50, and I have it running an infinite loop maxing out cpu usage.  On the [server list](http://www.ascensiongamedev.com/resources/filehost/7f49c1943d81d89216805dea04a00390.png), my users would see 100% CPU usage as expected, but on [the console page](http://www.ascensiongamedev.com/resources/filehost/fc8cae53b56d4fd5a3b94b719c9be5f3.png) it would sit around 50%.

The code is in the javascript to scale appropriately but the 'Pterodactyl.server.cpu' was undefined.

**My suggested fix (for the development branch):**
Modify app/Traits/Controllers/JavascriptInjection.php
After
 `'daemonSecret' => $token, on line 57`

Add
`'cpu' => $server->cpu;`

lancepioch:
Made a PR with your suggested fix, thanks.

PR #775, opened 2017-11-22
Define 'Pterodactyl.server.cpu' for #758

lancepioch:
Using @jcsnider's suggested fix from #758


DaneEveritt:
This would inject it unnecessarily on all pages, it should be injected on the pages it is needed on only.

DaneEveritt:
closed in https://github.com/Pterodactyl/Panel/commit/47f2ca0673bc9c1c673ef0d3568a14e3060d19da

PR #777, opened 2017-11-22
Lengthen search box

lancepioch:
Shouldn't use inline styling.

(same for all pages)
Also here's the before:

image

and after:

image


DaneEveritt:
This is not a good fix.

http://i.kumquat.io/0p3R3A203M2Q

lancepioch:
Instead of reverting... CSS should be modified.

PR #797, opened 2017-12-05
enable php 7.2 in travis and fix repository

arcdigital:


.travis.yml
```diff
@@ -1,9 +1,13 @@
...
-  - '7.0'
-  - '7.1'
-#  - '7.2'
+  - 7.0
+  - 7.1
+  - 7.2
+matrix:
+  fast_finish: true
+  allow_failures:
+    - php: 7.2
```
lancepioch: This should not be allowed.

DaneEveritt: Its so ghetto, I love it.

arcdigital:
MERGE IT FAM @DaneEveritt

PR #804, opened 2017-12-10
Feature/usage ping

arcdigital:


app/Console/Commands/Cloud/UsagePingCommand.php
```diff
@@ -0,0 +1,99 @@
...
+    public function handle()
+    {
+        try {
+            $this->client->request('POST', 'https://pterodactyl.cloud/api/usageData/panel', [
+                'timeout' => config('pterodactyl.guzzle.timeout'),
+                'connect_timeout' => config('pterodactyl.guzzle.connect_timeout'),
+                'json' => [
+                    'uuid' => config('pterodactyl.cloud.uuid'),
+                    'hostname' => config('app.url'),
+                    'appAuthor' => config('pterodactyl.service.author'),
```
DaneEveritt: I don't think any identifiable information should be sent in this, including the author and URL. It should only be a UUID that is not tied to anything publicly, and numbers.

DaneEveritt: This also includes the theme.

app/Console/Commands/Environment/AppSettingsCommand.php
```diff
@@ -136,6 +137,10 @@ public function handle()
             array_key_exists($selected, self::ALLOWED_QUEUE_DRIVERS) ? $selected : null
         );
 
+        if (empty($this->config->get('pterodactyl.cloud.uuid'))) {
+            $this->variables['PTERODACTYL_CLOUD_UUID'] = (string) Uuid::uuid4();
```
DaneEveritt: I think this could be renamed better to be `PTERODACTYL_INSTANCE_UUID` or something.

I would prefer `Uuid::uuid4()->toString()` over the `(string)` cast as well, just for consistency with the rest of the application.

app/Console/Commands/Cloud/UsagePingCommand.php
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        try {
+            $this->client->request('POST', 'https://pterodactyl.cloud/api/usageData/panel', [
```
DaneEveritt: Please make this domain configurable (but not necessarily an environment variable, see CDN calls).

app/Console/Commands/Cloud/UsagePingCommand.php
```diff
@@ -0,0 +1,99 @@
...
+                    'version' => config('app.version'),
+                    'features' => [
+                        'newDaemon' => config('pterodactyl.daemon.use_new_daemon'),
+                        'recaptcha' => config('recaptcha.enabled'),
+                        'theme' => config('themes.active'),
+                    ],
+                    'usage' => [
+                        'databases' => Database::count(),
+                        'databaseHosts' => DatabaseHost::count(),
+                        'eggs' => Egg::count(),
```
DaneEveritt: I think we should add some more details to eggs here and include the total number of official eggs and the total number of un-offical. I know that we can probably just do some simple math, but users can also remove official ones and use their own custom ones.

app/Console/Commands/Cloud/UsagePingCommand.php
```diff
@@ -0,0 +1,99 @@
...
+                        'locations' => Location::count(),
+                        'nests' => Nest::count(),
```
DaneEveritt: We probably don't need nests, they're just a category theres no relevant or helpful information with them.

app/Console/Commands/Cloud/UsagePingCommand.php
```diff
@@ -0,0 +1,99 @@
...
+                        'nodes' => Node::count(),
+                        'packs' => Pack::count(),
+                        'schedules' => Schedule::count(),
+                        'servers' => Server::count(),
+                        'users' => User::count(),
+                        'subusers' => Subuser::count(),
```
DaneEveritt: I don't really care about the number of subusers here.

app/Console/Commands/Cloud/UsagePingCommand.php
```diff
@@ -0,0 +1,99 @@
+<?php
```
DaneEveritt: Please include a test for this command.

app/Console/Commands/Environment/AppSettingsCommand.php
arcdigital: The reason it's called PTERODACTYL_CLOUD_UUID is because it's going to send this UUID to the service when it's downloading/uploading eggs, etc... so it's an ID specific to the cloud service (since we killed the generic UUID already and don't have a use for it in other places). Happy to change it though if you think we should have it be a generic UUID.

app/Console/Commands/Cloud/UsagePingCommand.php
arcdigital: I think it's just a good data point to collect, no harm in grabbing the count.

app/Console/Commands/Cloud/UsagePingCommand.php
arcdigital: I think it would be good to collect this to track usage of the feature.

app/Console/Commands/Cloud/UsagePingCommand.php
arcdigital: How do you want to differentiate the two?
official if the author is support@pterodactyl.io
unofficial if anything else?

app/Console/Commands/Cloud/UsagePingCommand.php
arcdigital: I can drop the author/url, what if we sent a custom_theme bool if the theme isn't the default pterodactyl one?

app/Console/Commands/Cloud/UsagePingCommand.php
DaneEveritt: Preferably a test that doesn't break travis every other commit*

app/Console/Commands/Cloud/UsagePingCommand.php
DaneEveritt: That works for me. Its probably not super helpful because most people just edit the pterodactyl one anyways, but a bool is fine.

app/Console/Commands/Cloud/UsagePingCommand.php
DaneEveritt: Correct 👍

app/Console/Commands/Cloud/UsagePingCommand.php
arcdigital: woah, now you're asking for too much...

app/Console/Commands/Environment/AppSettingsCommand.php
DaneEveritt: I think we should make it more generic if possible. I don't know where else it would be used currently, but wouldn't hurt to make it usable down the road if we add more things.

DaneEveritt:
Closing due to inactivity.

PR #814, opened 2017-12-16
Update to Laravel 5.5

DaneEveritt:


arcdigital:
NIIIICE Now I can upgrade to Dusk 2.0 and finish up that branch 🎉

app/Providers/SettingsServiceProvider.php
```diff
@@ -87,6 +87,24 @@ public function boot(ConfigRepository $config, Encrypter $encrypter, SettingsRep
...
+                    break;
+                case 'empty':
+                case '(empty)':
+                    $value = '';
+                    break;
+                case 'null':
+                case '(null)':
+                    $value = null;
+            }
+
```
lancepioch: This is fantastic. Maybe this should be put in a separate (global?) function.

app/Http/Requests/Admin/AdminFormRequest.php
```diff
@@ -44,9 +44,6 @@ public function authorize()
      */
     public function normalize($only = [])
     {
-        return array_merge(
-            $this->only($only),
-            $this->intersect(array_keys($this->rules()))
-        );
+        return $this->all(empty($only) ? array_keys($this->rules()) : $only);
```
lancepioch: This could be written better.

app/Providers/SettingsServiceProvider.php
DaneEveritt: Its actually copy/pasted from Laravel's `env()` function, I put it here because it was causing issues with the true/false stuff.

Shouldn't be needed anywhere else to my knowledge.

app/Http/Requests/Admin/AdminFormRequest.php
DaneEveritt: Yeah, but thats what happen when they deprecate `$request->intersect()` and make all work like it if you pass in keys.

I'll probably end up rewriting that at some point, but need better test coverage first because that controls a lot of functionality.

PR #850, opened 2018-01-04
Add env db prefix in database.php

dyanakiev:


Yomanz:
Although you have updated it in there, what about .env and the `php artisan p:environment:database` (Might be a stupid point but oh well)

lancepioch:
I think @Yomanz has a good point, it should ask you if you want a db prefix in the setup, otherwise default is blank for none.

DaneEveritt:
@lancepioch I do not think it should ask. Prefixing in the DB is unnecessary in almost all cases, and was never supported by the Panel to begin with in terms of setup questions.

If someone really needs this they can add the environment variable, but its not something I feel needs to be supported, and I do not want to add it to the environment setup when it isn't necessary to begin with.

PR #853, opened 2018-01-05
Fixed navigation account click

dyanakiev:
There was no dropdown and the account url was not working because of it.


dyanakiev:
Uh, there is the same issue in admin template, should i fix it and add new pull request?

DaneEveritt:
You don't need to open a new PR, just add to this one and I can squash them.

PR #903, opened 2018-01-31
Add the MAP Varible

rouing:
The MAP Var was added to startup config but never to the user config options.


database/seeds/eggs/source-engine/egg-custom-source-engine-game.json
```diff
@@ -40,6 +40,15 @@
             "user_viewable": 1,
             "user_editable": 0,
             "rules": "required|alpha_dash|between:1,100"
+        },
+        {
+            "name": "Map",
+            "description": "The default map for the server.",
+            "env_variable": "SRCDS_MAP",
+            "default_value": "",
```
DoNotSpamPls: Shouldn't the variable have a default value?

PR #918, opened 2018-02-06
Add support for allowing subusers to access SFTP.

ShadowKitten:
- Fixed unit tests.
- Modified some permissions for more friendly naming.
- Removed subuser SFTP permissions in favor of single 'SFTP Allowed' permission.

### DIFF
--- a/app/Models/Permission.php
+++ b/app/Models/Permission.php
@@ -96,9 +96,9 @@ class Permission extends Model implements CleansAttributes, ValidableContract
         ],
         'file' => [
             'view-sftp' => null,
-            'list-files' => 's:files:list',
-            'edit-files' => 's:files:edit',
-            'save-files' => 's:files:save',
+            'list-files' => 's:files:get',
+            'edit-files' => 's:files:read',
+            'save-files' => 's:files:post',
             'move-files' => 's:files:move',
             'copy-files' => 's:files:copy',
             'compress-files' => 's:files:compress',
--- a/app/Models/Permission.php
+++ b/app/Models/Permission.php
@@ -90,15 +90,15 @@ class Permission extends Model implements CleansAttributes, ValidableContract
             'view-startup' => null,
             'edit-startup' => null,
         ],
-        'sftp' => [
-            'view-sftp' => null,
-            'view-sftp-password' => null,
-            'reset-sftp' => 's:set-password',
+        'database' => [
+            'view-databases' => null,
+            'reset-db-password' => null,
         ],
         'file' => [
-            'list-files' => 's:files:get',
-            'edit-files' => 's:files:read',
-            'save-files' => 's:files:post',
+            'view-sftp' => null,
+            'list-files' => 's:files:list',
+            'edit-files' => 's:files:edit',
+            'save-files' => 's:files:save',
             'move-files' => 's:files:move',
             'copy-files' => 's:files:copy',
             'compress-files' => 's:files:compress',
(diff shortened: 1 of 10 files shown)


ShadowKitten:
I went ahead and reverted the permission changes as there is currently no way to update the daemon with the new permissions mappings. One permission mapping s:files:download is still going to be needed to be updated on the daemon for each subuser so they will be able to download files from SFTP.

I haven't been able to find a good solution to this yet besides possibly adding a CLI for creating and updating subusers.

app/Models/Permission.php
```diff
@@ -90,12 +90,12 @@ class Permission extends Model implements CleansAttributes, ValidableContract
...
-        'sftp' => [
-            'view-sftp' => null,
-            'view-sftp-password' => null,
-            'reset-sftp' => 's:set-password',
+        'database' => [
+            'view-databases' => null,
+            'reset-db-password' => null,
         ],
         'file' => [
+            'view-sftp' => null,
```
DaneEveritt: This should be renamed to `access-sftp` for clarity I think.

app/Services/Sftp/AuthenticateUsingPasswordService.php
```diff
@@ -25,21 +26,29 @@ class AuthenticateUsingPasswordService
...
+     * @param \Pterodactyl\Contracts\Repository\ServerRepositoryInterface  $repository
+     * @param \Pterodactyl\Contracts\Repository\UserRepositoryInterface    $userRepository
+     * @param \Pterodactyl\Contracts\Repository\SubuserRepositoryInterface $subuserRepository
      */
     public function __construct(
         DaemonKeyProviderService $keyProviderService,
         ServerRepositoryInterface $repository,
-        UserRepositoryInterface $userRepository
+        UserRepositoryInterface $userRepository,
+        SubuserRepositoryInterface $subuserRepository
```
DaneEveritt: Please keep these in alphabetical order, same for the variable setting in the constructor itself (and the order at the top of the file).

app/Services/Sftp/AuthenticateUsingPasswordService.php
```diff
@@ -73,10 +82,19 @@ public function handle(string $username, string $password, int $node, string $se
         }
 
         $server = $this->repository->setColumns(['id', 'node_id', 'owner_id', 'uuid', 'installed', 'suspended'])->getByUuid($server);
-        if ($server->node_id !== $node || (! $user->root_admin && $server->owner_id !== $user->id)) {
+        if ($server->node_id !== $node) {
             throw new RecordNotFoundException;
         }
 
+        if ((! $user->root_admin && $server->owner_id !== $user->id)) {
```
DaneEveritt: The double parenthesis are not needed here.

tests/Unit/Services/Sftp/AuthenticateUsingPasswordServiceTest.php
```diff
@@ -39,6 +45,7 @@ public function setUp()
         $this->keyProviderService = m::mock(DaemonKeyProviderService::class);
         $this->repository = m::mock(ServerRepositoryInterface::class);
         $this->userRepository = m::mock(UserRepositoryInterface::class);
+        $this->subuserRepository = m::mock(SubuserRepositoryInterface::class);
```
DaneEveritt: Same as above with the alphabetical order here.

PR #924, opened 2018-02-09
Add Squad Egg

Exonical:


schrej:
You somehow messed up there I think ;)

PR #935, opened 2018-02-11
Add Squad Egg

Exonical:


Exonical:
dont commit yet

Exonical:
Good to commit when ready

DaneEveritt:
Thanks for the PR. Going to have you hold onto this and once our sharing service is ready you can add it there!

Trying to avoid shipping too much with the core, especially eggs that we would then be response for maintaining. 👍

PR #976, opened 2018-02-23
Added a hostname variable.

ir-3:


DaneEveritt:
Please make sure to export these through the Panel so that the exported_at timestamps get updated as well. :)

ir-3:
I will get to that in a bit. Sorry.

rouing:
Hostname should be handled through server.cfg

DaneEveritt:
I agree with @rouing on this, seems better handled though a config file.

PR #993, opened 2018-02-26
Added filesystem as cache option

XMGamingMaster:


DaneEveritt:
Filesystem is not supported as we make use of the tagging system for the cache. If we want to support files you'd need to also look through the code and find anywhere using tags and see if we can either remove those tags, or handle not tagging for file cache.

XMGamingMaster:
We are currently using Filesystem as cache option and everything seems to work perfectly fine.

DaneEveritt:
It might be better in 0.7, but I know there were places using tagging, so you might want to do a check through the code to make sure those places no longer exist.

XMGamingMaster:
Yeah, i will do that.

PR #996, opened 2018-02-26
Added 2 Vars

rouing:
One var for Workshop Collection
One var for Gamemode

These are almost always added later on. I see no reason to not add them here with defaults. If the workshop collection is blank, it doesnt error out, but just ignores it.


rouing:
Closed in order to make further edits to the install script such as to add CS:S Mounting (Which is 99% of the time with Gmod Servers)

PR #998, opened 2018-02-27
Added 2 Vars and commonly mounted games

rouing:
Added 2 new extremely common vars used in the startup command and 2 extremely common mounted games for Garry's Mod.

Vars:

Workshop Collection (This is not a required var, ignored if blank. Almost always used and should be added)
Gamemode (Almost ALWAYS used and should be added)

Games commonly mounted:
Counter-Strike Source (Every server has this mounted [99.9999%])
Team Fortress 2 (Almost every server has this mounted [99.9%])

Despite taking up space, they are almost 100% always added and 99.9% of the time, disk space is never an issue. Without having them in the install script, it'll force non-technical users to find workarounds with the startup script to try to download these 2 commonly mounted games. By adding them to the install script, they can avoid messing up the server.

Install has been verified on my server multiple times and no errors presented.

### DIFF
--- a/database/seeds/eggs/source-engine/egg-garrys-mod.json
+++ b/database/seeds/eggs/source-engine/egg-garrys-mod.json
@@ -3,12 +3,12 @@
     "meta": {
         "version": "PTDL_v1"
     },
-    "exported_at": "2018-01-21T16:59:47-06:00",
+    "exported_at": "2018-02-26T19:08:48-06:00",
     "name": "Garrys Mod",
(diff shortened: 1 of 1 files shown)


TrixterTheTux:
I think forcing someone to download over 20 gb of content is a bad idea, especially when 99% of servers don't actually require tf2. Maybe make some type of optional variable to download it instead?

rouing:
I like that Idea. Have an idea on how to go about it? Possibly modify the Install Script to include the changes to mount.cfg

PR #999, opened 2018-02-27
Allow for 3 decimals MC Ver, future proof Forge Install

rouing:
The GET_VERSIONS Grep Command will fail with 1.12.2 but will work with 1.12. Added optional check for 3rd decimal place in the MC Version of Forge and future proofed the forge version for Double Numeric SubMinor version.


parkervcp:
I was actually working on this in my eggs repo.

rouing:
Would you like me to close to the PR then?

parkervcp:
If anything look into adding it to my "forge-enhanced" egg on my repo.

PR #1003, opened 2018-02-27
[WIP] Change egg script into an array called commands without breaking BC for script

lancepioch:
Initial PR

The default will be the new `commands`.


DaneEveritt:
@lancepioch can you fix that conflict please?

lancepioch:
@DaneEveritt I should be able to get to this in a weekend or two. And yes there'll be a test that confirms both versions work.

PR #1033, opened 2018-03-07
Feature/PTDL-1030

### LINKED ISSUE #1030: Swap reCAPTCHA "Secret Key" position with "Website Key"
* Panel or Daemon: Panel
* Version of Panel/Daemon: 0.7.5

------------------------

In my opinion the position of the reCAPTCHA "Secret Key" should be swapped with the reCAPTCHA "Website Key" in the panel's advanced settings. Why?
If you compare the position of both keys with the position of the keys at Google's reCAPTCHA dashboard, you'll notice that they are positioned the other way round.

_Pterodactyl Panel:_
recaptcha_pterodactyl

_Google reCAPTCHA Dashboard:_
recaptcha_google

I know this is only a minor point of difference, but I think it's important to be consistent with Google here (even because many people are just copy-pasting the keys without looking which one belongs where).

Pixelhash:
Fixes #1030

### DIFF
--- a/resources/themes/pterodactyl/admin/settings/advanced.blade.php
+++ b/resources/themes/pterodactyl/admin/settings/advanced.blade.php
@@ -35,7 +35,7 @@
                                 </div>
                             </div>
                             <div class="form-group col-md-4">
-                                <label class="control-label">Website Key</label>
+                                <label class="control-label">Site Key</label>
                                 <div>
                                     <input type="text" required class="form-control" name="recaptcha:website_key" value="{{ old('recaptcha:website_key', config('recaptcha.website_key')) }}">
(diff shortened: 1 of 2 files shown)


resources/themes/pterodactyl/admin/settings/advanced.blade.php
```diff
@@ -35,16 +35,16 @@
                                 </div>
                             </div>
                             <div class="form-group col-md-4">
-                                <label class="control-label">Secret Key</label>
+                                <label class="control-label">Website Key</label>
```
schrej: We should call it "Site Key" like in the dashboard then, right?

Issue #1034, opened 2018-03-07
Allocated ports should be validated to make sure they are in the allowed range


schrej:
When adding allocations to a node the port range should be validated (0 - 65535), we should *maybe* even forbid using ports below 1024 as they are system ports and should not be used, although that might restrict some users.

Issue #1028, opened 2018-03-07
Pressing "cancel" on "move" file,  it delete the file


xion87:
Panel or Daemon: Both?
Version of Panel/Daemon: 0.7.3/0.5.4
Server's OS: Ubuntu 16 LTS
Your Computer's OS & Browser: Chrome/Windows 10

When i use "Move" command on file manager and press "Cancel" it delete the file instead to reject the operation, pressing "Esc" instead it "fake delete" the file until the refresh of page
![image](https://user-images.githubusercontent.com/3611852/37098432-23e714c0-221e-11e8-8779-6091bb5d2d5f.png)

schrej:
Is this just a visual glitch or does it actually remove the file?

xion87:
If you press cancel it delete the file (for real)
if you press esc it fake delete this file (reloading the page it appear again)

DaneEveritt:
I am unable to reproduce this on both Chrome and Safari.

xion87:
I tried again:
- create a file called test.yml put some text into it
- use move tool on that file, click cancel
- the file will gone, tryed also to look into sftp, is not here anymore.

DaneEveritt:
I still cannot reproduce this.

xion87:
btw i was using the italian translation maybe is related?

DaneEveritt:
Theres no reason the language setting would have anything to do with that, but I went ahead and changed it, still unable to replicate this issue. In addition there is absolutely 0 logic in the move function that would delete a file, or otherwise move it to a location in which it would not show up somewhere in the file manager.

xion87:
at least is fixed now for me for any strange reason,

DaneEveritt:
I'm going to close this and assume it was a really strange fluke in the caching mechanics. I'm working on fixing the caching behavior on the Panel as well to avoid that ever being a potential issue down the road.

Arsoth:
I know this is closed, but I just had the same thing occur to me while attempting... not to move a .zip file I suppose?

Panel or Daemon: Both I guess?
Version of Panel/Daemon: Panel: 0.7.6, Daemon: 0.5.5
Server's OS: Debian 9 "Stretch"
Your Computer's OS & Browser: Mac OS 10.12.6 and Firefox Nightly 61.0a1

I selected move, then attempted to cancel (honestly can't remember if it was the cancel button or the escape key, but I assume the button), and the .zip no longer was shown on the filesystem. 
[...]

PR #1038, opened 2018-03-08
#1034 and #1028

XMGamingMaster:


DaneEveritt:
Hey, thanks for this PR. Two things of note though:

1.) Generally we prefer to keep a PR limited to one issue, it makes it eaiser to review and follow along with, and
2.) Its much easier to read if you do shorter if/else blocks, for example in the file manager fixes, doing `if (!val) { return; } <remaining logic>` rather than `if (val !== false) { <remaining logic> }`.

👍

DaneEveritt:
Not really sure what you did to this PR, but I am closing it as theres no changed code, and it looks like you merged something badly.

Issue #936, opened 2018-02-14
Add descriptions for nodes


NEXUS2345:
Could we get a description field for nodes? It would help in environments with multiple administrators so that the user can be given an idea what the hardware or software configuration is at a quick glance.

PR #1065, opened 2018-03-17
Add description field to nodes

stanjg:
Adds a description fields to nodes to identify them easier. For issue #936


database/migrations/2018_03_15_124536_add_description_to_nodes.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('nodes', function (Blueprint $table) {
+            $table->text('description')->nullable()->after('name');
```
DaneEveritt: You cannot have a `TEXT` field that is nullable in MySQL.

resources/themes/pterodactyl/admin/nodes/new.blade.php
```diff
@@ -32,15 +32,16 @@
                         <input type="text" name="name" id="pName" class="form-control" value="{{ old('name') }}"/>
                         <p class="text-muted small">Character limits: <code>a-zA-Z0-9_.-</code> and <code>[Space]</code> (min 1, max 100 characters).</p>
                     </div>
+                    <div class="form-group">
+                        <label for="pDescription" class="form-label">Description</label>
+                        <textarea name="description" id="pDescription" rows="4" class="form-control" style="resize: vertical;">{{ old('description') }}</textarea>
+                        <p class="text-muted small">(max 2000 characters)</p>
```
DaneEveritt: There should not be a limit on description size.

resources/themes/pterodactyl/admin/nodes/new.blade.php
```diff
@@ -32,15 +32,16 @@
                         <input type="text" name="name" id="pName" class="form-control" value="{{ old('name') }}"/>
                         <p class="text-muted small">Character limits: <code>a-zA-Z0-9_.-</code> and <code>[Space]</code> (min 1, max 100 characters).</p>
                     </div>
+                    <div class="form-group">
+                        <label for="pDescription" class="form-label">Description</label>
+                        <textarea name="description" id="pDescription" rows="4" class="form-control" style="resize: vertical;">{{ old('description') }}</textarea>
```
DaneEveritt: Please remove the `resize:vertical` to match the way the rest of the panel works.

app/Models/Node.php
```diff
@@ -81,6 +82,7 @@ class Node extends Model implements CleansAttributes, ValidableContract
      */
     protected static $applicationRules = [
         'name' => 'required',
+        'description' => 'sometimes|max:2000',
```
DaneEveritt: This rule can simply be `string`

database/migrations/2018_03_15_124536_add_description_to_nodes.php
stanjg: Oh, didn't know about that. Thanks.

DaneEveritt:
Please include additions/modifications to the tests to cover cases where the description is passed in as a null value to ensure that it is properly converted to a string. I don't see any of that happening in the services currently.

PR #1080, opened 2018-03-22
Translation to Brazilian Portuguese - pt-br

mullernato:
Translation to Brazilian Portuguese


DaneEveritt:
Please do not make multiple PRs changing one file each. We also already have most of these translated on an external service, if you get in contact with @schrej on Discord he can help get you setup on there to translate them.

PR #1101, opened 2018-04-08
Public is just the boolean

lancepioch:


DaneEveritt:
If it isn't cast to an int you'll see issues when doing strict checks due to MySQL returning a string. If we're going to call it a boolean value we should just cast it as a bool as well.

lancepioch:
@DaneEveritt Here's that array expanded: 
image

DaneEveritt:
oh.

PR #1104, opened 2018-04-09
no longer lie about spigot being built when no url provided

schrej:


sachk:
Over [here](https://github.com/parkervcp/eggs) parkervcp has been working on a new similar egg that actually builds spigot. So this change seems a bit pointless since this egg is most likely going to be replaced soon.

Daxorinator:
@sachk ParkerVCP has stated that the built in eggs won't be updated or replaced, even if his one builds correctly. All eggs will still be available from his repo.

parkervcp:
My repo does not feed into the main repo. The building function was removed a long while ago due to technical issues and will stay removed.

DaneEveritt:
I’m fine with adding it back if it is working smoothly now. Generally I won’t accept PRs to add new eggs, but updates and improvements to existing ones are fine.

Issue #1106, opened 2018-04-10
Per user panel language


jer3m01:
Feature request:

Would be nice if you could have an option (under profile) where you can change your language of the panel (not being admin and without changing everyone's panel's language).
Of course the language should be in the available languages on the server.

Thanks for your consideration.

LeCodeCo:
+1

JanDevDE:
+1

LeCodeCo:
I think this already exists. https://file.properties/host/MgHGkveO.png

DaneEveritt:
Closing in favor of #2518 which can handle this at the same time.

PR #1128, opened 2018-05-04
Add support for user specific languages

stanjg:
For issue #1106


.env.example
```diff
@@ -6,6 +6,7 @@ APP_TIMEZONE=America/New_York
 APP_CLEAR_TASKLOG=720
 APP_DELETE_MINUTES=10
 APP_ENVIRONMENT_ONLY=true
+LANG_MUST_BE_GLOBAL=false
```
schrej: maybe call this `APP_LOCALE_IS_GLOBAL=false`

app/Http/Controllers/Base/AccountController.php
```diff
@@ -60,7 +65,7 @@ public function update(AccountDataFormRequest $request)
         } elseif ($request->input('do_action') === 'email') {
             $data['email'] = $request->input('new_email');
         } elseif ($request->input('do_action') === 'identity') {
-            $data = $request->only(['name_first', 'name_last', 'username']);
+            $data = $request->only(['name_first', 'name_last', 'username', 'language']);
```
schrej: The language should be validated

stanjg: Yes, you're right. I'm on it.

.env.example
stanjg: Done

.env.example
```diff
@@ -6,7 +6,7 @@ APP_TIMEZONE=America/New_York
 APP_CLEAR_TASKLOG=720
 APP_DELETE_MINUTES=10
 APP_ENVIRONMENT_ONLY=true
-LANG_MUST_BE_GLOBAL=false
+APP_LOCALE_IS_GLOBAL=false
```
schrej: As you're at it: Could you also add `APP_LOCALE=en` to the .env.example? Then this setting isn't that out-of-context

app/Http/Controllers/Base/AccountController.php
stanjg: Nevermind, it's already being validated. Take a look at `gatherRules()` in `Models\User`

.env.example
stanjg: Sure

schrej:
How did that coverage.xml get added?

stanjg:
I have no idea, it got added after I ran the tests.

app/Http/Middleware/LanguageMiddleware.php
```diff
@@ -47,7 +48,11 @@ public function __construct(Application $app, Repository $config)
      */
     public function handle(Request $request, Closure $next)
     {
-        $this->app->setLocale($this->config->get('app.locale', 'en'));
+        if (! Auth::check() || $this->config->get('pterodactyl.lang.global')) {
```
DaneEveritt: This should avoid using the Auth facade (use `$request->user()`).

Also, theres no reason to have the global language setting. Just always allow the user to select a language, and if one is not selected or it doesn't exist, fallback to the default language setting.

stanjg: Ok, will avoid the facade.

For the option to have a global language, I added that as I could imagine some companies would want to enable it. For example for a support matter it might be hard to give instructions in another language. However I can see how it can be unnecessary. Let me know if I should remove it or not.

DaneEveritt: I would remove it, just have the default language be used if the client selects an invalid language.

Worst case the company can just disable this middleware.

stanjg: Yeah, I already did. It should be ready to merge.

DaneEveritt:
As discussed on Discord, holding off on merging this in until I know how we're handling them with the new vuejs stuff.

PR #1129, opened 2018-05-04
Added a maintenance mode for nodes

### LINKED ISSUE #142: Maintenance mode per node
Would love some sort of a maintenance mode per node that an admin can turn on in the panel. This could send a notification to the users with servers on the node and display some error/notification on the server management page for the user. Possibly also disable server controls or pop a notification when they are clicked informing about the current maintenace and unavailability of the node.

stanjg:
When a node is under maintenance it blocks access to the pages of the servers on that node to prevent the users from seeing the not so pretty error you get when a daemon is down.

Closes #142


DaneEveritt:
It might take me a little bit to review this, but it is seen 👍

app/Http/Kernel.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Pterodactyl\Http;
 
+use Pterodactyl\Http\Middleware\MaintenanceMiddleware;
```
DaneEveritt: Run the linter against these files, these imports should be in length based order (and then alphabetical if needed).

app/Http/Kernel.php
```diff
@@ -108,6 +109,7 @@ class Kernel extends HttpKernel
         'can' => Authorize::class,
         'bindings' => SubstituteBindings::class,
         'recaptcha' => VerifyReCaptcha::class,
+        'maintenance' => MaintenanceMiddleware::class,
```
DaneEveritt: I'm wondering if we should re-label this as `NodeMaintenanceMode` (`node.maintenance`) so its not confused with the Laravel specific maintenance mode.

app/Http/Middleware/MaintenanceMiddleware.php
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        $server = $request->attributes->get('server');
```
DaneEveritt: You probably just want to add a repository function to the `NodeRepository` class called `isInMaintenanceMode(int $nodeId): bool` and then you can just call `if ($repository->isInMaintenanceMode($server->node_id)` which then allows this to be unit tested better.

Currently the way you're doing this, $server->node will fire off a request to the database unless the node relationship is already loaded (which it generally is, but not the best assumption).

If you want to, you can do a `$server->relationLoaded('node')` call and if it is loaded check for `->maintenance`, otherwise fallback to the repository call.

app/Models/Node.php
```diff
@@ -48,6 +48,7 @@ class Node extends Model implements CleansAttributes, ValidableContract
         'daemonSFTP' => 'integer',
         'behind_proxy' => 'boolean',
         'public' => 'boolean',
+        'maintenance' => 'boolean',
```
DaneEveritt: Lets rename this to `maintenance_mode`, just for readability down the road.

resources/themes/pterodactyl/admin/nodes/view/settings.blade.php
```diff
@@ -108,6 +108,20 @@
                         </div>
                         <p class="text-muted small">If you are running the daemon behind a proxy such as Cloudflare, select this to have the daemon skip looking for certificates on boot.</p>
                     </div>
+                    <div class="form-group col-xs-12">
```
DaneEveritt: Lets just make this `Maintenance Mode: Enabled|Disabled`

resources/themes/pterodactyl/base/index.blade.php
```diff
@@ -64,9 +64,15 @@
                                         <span class="label bg-blue">@lang('strings.subuser')</span>
                                     @endif
                                 </td>
-                                <td class="text-center" data-action="status">
-                                    <span class="label label-default"><i class="fa fa-refresh fa-fw fa-spin"></i></span>
-                                </td>
+                                @if($server->node->maintenance)
+                                    <td class="text-center">
+                                        <span class="label label-warning">Maintenance</span>
```
DaneEveritt: Maybe refer to this as 'Under Maintenance' and also make it a translatable string. :)

DaneEveritt:
I think this would be cool to have specific messages on each node, but I like this for now, we can always come back and add the specific messages. Might also be cool to add a way to do a semi-available maintenance mode which allows a user to access the server but puts a banner at the top telling them things are being worked on and might be unstable.

That can all be done as a part 2 down the road though. Looks good so far!

DaneEveritt:
Oh, also could you add some unit tests for the middleware you added?

app/Http/Kernel.php
schrej: I told him to leave it as style ci will fix it automatically when we run it again, no real need to do an extra commit for such little stuff imo

DaneEveritt: Ah didn't notice the styleci failure.

PR #1130, opened 2018-05-04
Added a statistics page to monitor the panel usage

### LINKED ISSUE #156: Statistics page
A statistic page with information (shown per node and service with a total too) to keep track of the overall installation would be great.
It could contain:
- Amount of servers
- Amount of suspended servers
- Online players total
- Resources assigned/in use (RAM, Diskspace)
- More things I currently can't think of.

stanjg:
Closes #156 

Screenshot

### DIFF
--- a/app/Http/Controllers/Admin/StatisticsController.php
+++ b/app/Http/Controllers/Admin/StatisticsController.php
@@ -11,18 +11,15 @@ use Pterodactyl\Contracts\Repository\NodeRepositoryInterface;
 use Pterodactyl\Contracts\Repository\ServerRepositoryInterface;
 use Pterodactyl\Contracts\Repository\UserRepositoryInterface;
 use Pterodactyl\Http\Controllers\Controller;
-use Pterodactyl\Services\DaemonKeys\DaemonKeyProviderService;
-use Pterodactyl\Traits\Controllers\JavascriptStatisticsInjection;
+use Pterodactyl\Traits\Controllers\PlainJavascriptInjection;
 
 class StatisticsController extends Controller
 {
-    use JavascriptStatisticsInjection;
+    use PlainJavascriptInjection;
 
     private $allocationRepository;
 
     private $databaseRepository;
-
-    private $keyProviderService;
 
     private $eggRepository;
 
@@ -35,7 +32,6 @@ class StatisticsController extends Controller
     function __construct(
         AllocationRepositoryInterface $allocationRepository,
         DatabaseRepositoryInterface $databaseRepository,
-        DaemonKeyProviderService $keyProviderService,
         EggRepositoryInterface $eggRepository,
         NodeRepositoryInterface $nodeRepository,
         ServerRepositoryInterface $serverRepository,
@@ -44,28 +40,33 @@ class StatisticsController extends Controller
     {
         $this->allocationRepository = $allocationRepository;
         $this->databaseRepository = $databaseRepository;
-        $this->keyProviderService = $keyProviderService;
         $this->eggRepository = $eggRepository;
         $this->nodeRepository = $nodeRepository;
         $this->serverRepository = $serverRepository;
         $this->userRepository = $userRepository;
     }
 
-    public function index(Request $request)
+    public function index()
     {
         $servers = $this->serverRepository->all();
-        $serversCount = count($servers);
         $nodes = $this->nodeRepository->all();
-        $nodesCount = count($nodes);
         $usersCount = $this->userRepository->count();
         $eggsCount = $this->eggRepository->count();
         $databasesCount = $this->databaseRepository->count();
-        $totalServerRam = DB::table('servers')->sum('memory');
-        $totalNodeRam = DB::table('nodes')->sum('memory');
-        $totalServerDisk = DB::table('servers')->sum('disk');
-        $totalNodeDisk = DB::table('nodes')->sum('disk');
         $totalAllocations = $this->allocationRepository->count();
-        $suspendedServersCount = $this->serverRepository->getBuilder()->where('suspended', true)->count();
+        $suspendedServersCount = $this->serverRepository->getSuspendedServersCount();
+
+        $totalServerRam = 0;
+        $totalNodeRam = 0;
+        $totalServerDisk = 0;
+        $totalNodeDisk = 0;
+        foreach ($nodes as $node) {
+            $stats = $this->nodeRepository->getUsageStatsRaw($node);
+            $totalServerRam += $stats['memory']['value'];
+            $totalNodeRam += $stats['memory']['max'];
+            $totalServerDisk += $stats['disk']['value'];
+            $totalNodeDisk += $stats['disk']['max'];
+        }
 
         $tokens = [];
         foreach ($nodes as $node) {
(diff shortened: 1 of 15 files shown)


app/Http/Controllers/Admin/StatisticsController.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        $this->keyProviderService = $keyProviderService;
+    }
+
+    public function index(Request $request)
+    {
+        $servers = Server::all();
+        $nodes = Node::all();
+        $serversCount = count($servers);
+        $nodesCount = Node::count();
```
lancepioch: Can you just use the count function like the $servers?

app/Http/Controllers/Admin/StatisticsController.php
```diff
@@ -0,0 +1,75 @@
...
+use Pterodactyl\Http\Controllers\Controller;
+use Pterodactyl\Models\Allocation;
+use Pterodactyl\Models\Database;
+use Pterodactyl\Models\Egg;
+use Pterodactyl\Models\Node;
+use Pterodactyl\Models\Server;
+use Pterodactyl\Models\User;
+use JavaScript;
+use Illuminate\Support\Facades\DB;
+use Pterodactyl\Services\DaemonKeys\DaemonKeyProviderService;
```
lancepioch: IIRC this should be alphabetized

app/Http/Controllers/Admin/StatisticsController.php
```diff
@@ -0,0 +1,75 @@
...
+        $suspendedServersCount = Server::where('suspended', true)->count();
+
+        $tokens = [];
+        foreach ($nodes as $node) {
+            $tokens[$node->id] = $this->keyProviderService->handle($node->servers->get(0), $request->user());
+        }
+
+        Javascript::put([
+            'servers' => $servers,
+            'serverCount' => $serversCount,
```
lancepioch: Also for the counts, I feel like you can just count them in JS anyways if you're passing the entire collection into JS.

public/themes/pterodactyl/js/admin/statistics.js
```diff
@@ -0,0 +1,121 @@
...
+        statusChart.update();
+    });
+
+    index++;
+}
+
+function getNodeByID(id) {
+    for (var i = 0; i < nodes.length; i++) {
+        if (nodes[i].id === id)
+            return nodes[i];
```
lancepioch: {} {} for consistent styling

public/themes/pterodactyl/js/admin/statistics.js
```diff
@@ -0,0 +1,121 @@
...
+            }
+        ]
+    }
+});
+
+var servers = Pterodactyl.servers;
+var nodes = Pterodactyl.nodes;
+
+for (let i = 0; i < servers.length; i++) {
+    setTimeout(getStatus, 200 * i);
```
lancepioch: Pass the node (using server uuid) into the function instead of having a global index variable.

```javascript
for (let i = 0; i < servers.length; i++) {
    setTimeout(getStatus, 200 * i,  getNodeByID(servers[i].uuid));
}

function getStatus(node) {...}
```

lancepioch: Also I don't think it's even necessary to use setTimeout for this occasion because Javascript executes code sequentially. If you absolutely need them to execute in order, then make the ajax calls synchronous.

app/Http/Controllers/Admin/StatisticsController.php
stanjg: Yes actually

app/Http/Controllers/Admin/StatisticsController.php
stanjg: But I use them as a variable in the blade template aswell

public/themes/pterodactyl/js/admin/statistics.js
stanjg: The requests get sent too fast without a delay, resulting in 404 or 403 errors from the daemon

app/Http/Controllers/Admin/StatisticsController.php
lancepioch: Okay, it's not a big deal for this page.

app/Http/Controllers/Admin/StatisticsController.php
DaneEveritt: Should be sorted by length, and then alphabetically if anything is the same length. The .phpcs file should handle all of that if you can setup a linter locally.

app/Http/Controllers/Admin/StatisticsController.php
```diff
@@ -29,7 +29,7 @@ public function index(Request $request)
         $servers = Server::all();
```
DaneEveritt: I'm trying to avoid using any of the model direct eloquent calls in this fashion as they are difficult to unit test properly. There should be a repository call for all of these, its going to be quite a constructor most likely.

app/Http/Controllers/Admin/StatisticsController.php
```diff
@@ -29,7 +29,7 @@ public function index(Request $request)
...
         $nodes = Node::all();
         $serversCount = count($servers);
-        $nodesCount = Node::count();
+        $nodesCount = count($nodes);
```
DaneEveritt: I'm not sure these counts are necessary, you should be able to do `{{ count($servers) }}` in the blade templates, or `Pterodactyl.servers.length` in JS and keep things a little cleaner. :)

app/Http/Controllers/Admin/StatisticsController.php
```diff
@@ -43,7 +43,11 @@ public function index(Request $request)
 
         $tokens = [];
         foreach ($nodes as $node) {
-            $tokens[$node->id] = $this->keyProviderService->handle($node->servers->get(0), $request->user());
+            $server = Server::where('node_id', $node->id)->first();
+            if ($server == null)
```
DaneEveritt: Two things in here, the first is that this if check should use braces ( and you could do `is_null($server)`), but I would recommend just getting the count of servers from the database to check aganist.

There should also be a service that providers a node specific token, rather than a token for a specific server. All administrative users should theoretically have access to this page, and theres no admin specific permissions right now so you can expose the node daemon secret here without any issues.

Doing the per-server token will end up making a _large_ amount of DB calls if there are a lot of nodes. We can go back to this and adjust that once there are admin specific permissions in place.

app/Http/Controllers/Admin/StatisticsController.php
```diff
@@ -43,7 +43,11 @@ public function index(Request $request)
...
+                continue;
+
+            $tokens[$node->id] = $this->keyProviderService->handle($server, $request->user());
         }
 
         Javascript::put([
```
DaneEveritt: There should be an `InjectsJavascript` trait, or similarly named one that will allow you to make this call unit testable.

public/themes/pterodactyl/js/admin/statistics.js
```diff
@@ -0,0 +1,118 @@
...
+                label: 'RAM in MBs',
+                backgroundColor: ['#51B060', '#ff0000'],
+                data: [freeRam, Pterodactyl.totalServerRam]
+            }
+        ]
+    }
+});
+
+var activeServers = Pterodactyl.serverCount - Pterodactyl.suspendedServers;
+let serversChart = new Chart($('#servers_chart'), {
```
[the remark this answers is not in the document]
stanjg: It's working on all latest versions of major browsers. On older versions it's a bit tricky though. Let me know if it's okay to use `let`.

https://caniuse.com/#feat=let

PR #1146, opened 2018-05-13
Add support for Windows, replace all backslashes with forwardslashes

lancepioch:
This breaks support for filenames with backslashes in them in Unix, however there's not really any good reasons to have files with forward or back slashes in them either way on any OS.


lancepioch:
Let me know @DaneEveritt if you'd like to see some tests covering this.

PR #1147, opened 2018-05-13
Upgrade Laravel to 5.6

lancepioch:
Includes and Depends on #1143 being merged.


lancepioch:
@DaneEveritt I'd like to increase overall test coverage after my other prs are merged. Therefore this one is fine too wait until we can be more confident about the upgrade.

PR #1166, opened 2018-05-22
Fix mistakes in the German lang files and add missing translations

ZeroxTV:
I fixed some mistakes in the German localisation of the panel I noticed, after installing it on my server. I went through every single file in the /resource/lang/de directory and fixed every mistake or typo I came across.

I also added translations for some previously untranslated parts, while ignoring the untranslated parts of the /command directory, because I thought you would want to keep the command line tool in English, although I can translate it at any time, if you want me to.

Hope I could help,
ZeroxTV


resources/lang/de/auth.php
```diff
@@ -1,21 +1,21 @@
 <?php
 
 return [
-    'not_authorized' => 'Du bist nicht autorisiert diese Aktion auszuführen0.',
-    'auth_error' => 'Es gab ein Problem während du dich anmelden wolltest.',
+    'not_authorized' => 'Du bist nicht autorisiert diese Aktion auszuführen.',
+    'auth_error' => 'Es gab ein Problem während der Anmeldung wolltest.',
```
BlobCodes: "Es gab ein Problem während der Anmeldung wolltest." is not correct

resources/lang/de/command/messages.php
```diff
@@ -15,28 +15,28 @@
         'confirm_delete' => 'Bist du dir wirklich sicher?',
         'no_users_found' => 'Es wurden keine User gefunden.',
         'multiple_found' => 'Es wurden mehrere Accounts gefunden.',
-        'ask_admin' => 'Is this user an administrator?',
+        'ask_admin' => 'Ist dieser User ein Administrator?',
```
BlobCodes: I think Ist dieser Benutzer ein Administrator? would be better

resources/lang/de/command/messages.php
```diff
@@ -15,28 +15,28 @@
...
         'ask_email' => 'Email Adresse',
         'ask_username' => 'Username',
         'ask_name_first' => 'Vornamee',
```
BlobCodes: Should be 'Vorname' without double e

resources/lang/de/auth.php
BlobCodes: "Es gab ein Problem während der Anmeldung wolltest." is not correct

resources/lang/de/command/messages.php
BlobCodes: I think Ist dieser Benutzer ein Administrator? would be better

resources/lang/de/command/messages.php
```diff
@@ -15,28 +15,28 @@
...
-            'If this is not what you wanted to do, press CTRL+C to exit this process.',
+            'Dieser Befehl deaktiviert 2-Faktor-Authentifizierung für ein Benutzerkonto, falls es aktiviert ist. Dieser Befehl sollte nur zur Accountrettung verwendet werden, wenn sich ein Nutzer aus seinem Account ausgeschlossen hat.',
+            'Falls das nicht ist, was du erreichen wolltest, drücke Strg+C, um diesen Prozess zu beenden.',
         ],
-        '2fa_disabled' => '2-Factor authentication wurde für :email deaktivier.',
+        '2fa_disabled' => '2-Faktor-Authentifizierung wurde für :email deaktiviert.',
     ],
     'schedule' => [
-        'output_line' => 'Dispatching job for first task in `:schedule` (:hash).',
+        'output_line' => 'Erledigt den Job für die erste Aufgabe in `:schedule` (:hash).',
```
BlobCodes: Shouldn´t this start with "Erledige" ?
"Erledigt" is the adjective!

resources/lang/de/exceptions.php
```diff
@@ -1,55 +1,55 @@
...
     ],
     'subusers' => [
-        'editing_self' => 'Du darfst deinen eigenen SUbuser nicht bearbeiten.',
+        'editing_self' => 'Du darfst deinen eigenen Subuser nicht bearbeiten.',
         'user_is_owner' => 'Du kannst den Owner nicht als Subuser hinzufügen.',
         'subuser_exists' => 'Diese Email ist bereits registriert.',
     ],
     'databases' => [
-        'delete_has_databases' => 'Es kann keine Datenbank gelöscht werden die von einem aktivien Server gelöscht wird.',
+        'delete_has_databases' => 'Es kann keine Datenbank gelöscht werden, die von einem sktiven Server verwendet wird.',
```
BlobCodes: "sktiven"

BlobCodes:
I remade the resources/lang/de/command/messages.php as there are still a lot of english sentences.
https://github.com/pterodactyl/panel/pull/1167/files

BlobCodes:
Yes, I am also a native speaker

resources/lang/de/command/messages.php
ZeroxTV: Typo, will fix it

resources/lang/de/command/messages.php
ZeroxTV: They used "User" all over the place in the original German translation, so I thought that the term "User" is what they wanted to use throughout the Panel.

resources/lang/de/command/messages.php
ZeroxTV: Nope, "erledigt" is correct. You can use both depending on the perspective the message is sent from, either the tool "says" "I am dispatching...", then it's the way you said it, or the tool says "Some process is dispatching", then it's the way I phrased it.

ZeroxTV:
Can someone tell me, why the Travis-CI Build is failing? I only changed the language files.
It seems to be something about "DeleteUserCommandTest", or more specifically:

1) Tests\Unit\Commands\User\DeleteUserCommandTest::testCommandWithInvalidInitialSearch
Mockery\Exception\NoMatchingExpectationException: No matching handler found for Mockery_14_Pterodactyl_Contracts_Repository_UserRepositoryInterface::setSearchTerm('yes'). Either the method was unexpected or its arguments matched no expected argument list for this method

schrej:
Don't care about travis with this PR as long as you're only changing language files.

PR #1173, opened 2018-05-25
[WIP] Disallow changing an already created node to HTTP if connection is secure

jacobgc:
I'm not sure why this hasn't been done already. But hey, if it is intentionally not done, please let me know. :)

### DIFF
--- a/app/Http/Requests/Api/Application/Nodes/StoreNodeRequest.php
+++ b/app/Http/Requests/Api/Application/Nodes/StoreNodeRequest.php
@@ -74,7 +74,6 @@ class StoreNodeRequest extends ApplicationApiRequest
         $response = parent::validated();
         $response['daemonListen'] = $response['daemon_listen'];
         $response['daemonSFTP'] = $response['daemon_sftp'];
-        $response['daemonBase'] = $response['daemon_base'];
 
         unset($response['daemon_base'], $response['daemon_listen'], $response['daemon_sftp']);
 


rouing:
But why reduce functionality? What if I want to remove an SSL certificate as it expired and I need to purchase one further down the road.

jacobgc:
It only reduces the functionality if the panel is running with HTTPS. If you were to run the panel with HTTPS and the nodes on HTTP browsers will fail to make any requests to them as you cant load none HTTPS pages from a HTTPS origin.


> jquery.min.js?t=e9b8dc0ea7d1478fbd77a7cf03d2ee4e:4 Mixed Content: The page at 'https://panel.XXXXXXXX.com/admin/nodes' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'http://001-paris-nodes.XXXXXXXX.com:8080/v1'. This request has been blocked; the content must be served over HTTPS.

jacobgc:
Also just to add, when creating a new node, this functionality already exists. I simply copied it over to work for editing a node.

lancepioch:
Ignoring whether or not we would want to have this, it's not a good way to disable functionality.

Perhaps a warning below would suffice to let the user know what they might experience, because right now this can be easily overridden with an Inspect Element change and would only frustrate some users.

jacobgc:
I don't think you understand. This functionality already exists when you are creating a node. See the screenshot below:

capture

In regard to frustrating users, it would be way more frustrating to allow them to set it to HTTP only for it to fail?

lancepioch:
So right now, if in the node settings page you go and try to use an HTTP connection, it will currently fail and notify the user? If not, then it needs to in addition to this PR. There has to be backend logic to enforce the front end logic.

jacobgc:
@lancepioch Currently it will silently fail (unless they check their browsers' console) but for the back end validation, there is none for creating a new node either. It seems like someone has forgotten to implement that.

schrej:
Good Idea to enforce it when modifying nodes as well. We should actually enforce it in the backend logic as well as Lance said.

jacobgc:
I'm not too familiar with PHP anymore but I'll have a go at implementing the backend logic

DaneEveritt:
Where are we at on this? I'm fine with enforcing HTTPS on the nodes like this, but the code does not appear to be ready to handle it correctly and validate that logic.

DaneEveritt:
Closing due to inactivity.

PR #1178, opened 2018-05-28
Fixed 500 error on POST /api/application/nodes

jacobgc:
Line 77 is not needed as daemon_base does not exist because the API still takes the daemon base parameter as ```daemonBase``` NOT ```daemon_base```.


DaneEveritt:
Can you please show me the request you made that caused this error to occur?

DaneEveritt:
`daemon_base` is defaulted to `/srv/daemon-data` and therefore is only necessary if you plan on using a different path.

lancepioch:
@DaneEveritt We can just reopen this issue and use it to track that.

DaneEveritt:
The line of code I provided above is what is needed to fix it.

PR #1187, opened 2018-06-03
Fixed Search Bar Scaling in Firefox

JaydenDownes:
Fixed scaling issue with Firefox where it would scale the search bar to 30% of the width of the box-tools container instead of 30% of the row container. (Note this is not a fix for the issue with it appearing on a separate line on mobiles).


DaneEveritt:
This is used in multiple places throughout the panel, would you be able to fix all of those places? (Most admin CP listing pages for example)

JaydenDownes:
Sure, I will look into it today for you.

JaydenDownes:
Dane, I took a look at it and could only find one other div that uses box-tools and it looks better off without scaling it to 30% of the box-body. The previous changes I have made include adding an extra div to the stylesheet for just that one search bar and all the rest of the box-tool divs seem to be fine as they don't have inline styles and aren't changed in Firefox.

PR #1188, opened 2018-06-04
Update issue templates

lancepioch:
When creating an issue now:

image

The button will be green and say Get started instead.


schrej:
We should change it back to beeing commented to avoid having all the instructions in every issue..

PR #1198, opened 2018-06-12
Fixed Grammar and Spelling issues.

JaydenDownes:
Periods don't need to be doubled and I also fixed a few other minor issues.


JaydenDownes:
Leave this one open for now as I have a few more things to fix.

matthewpi:
I'd recommend changing errors.maintenance.desc to "This node is currently under maintenance, therefore you may not access your server.", it flows better and is shorter.

JaydenDownes:
Sure, I will change it now.

linktlh:
E-mail and Email are both grammatically correct I don't get why you changed it.

lancepioch:
Thank you for the contribution, but the grammar looks acceptable in these cases, plus you also switched from single to double quotes in one case.

I believe there are some more grammar mistakes in the untranslated hardcoded English in the app. We'd appreciate it if you could look through that.
Also it'd help if you were able to find and add those commonly used strings to the translation files.

PR #1206, opened 2018-06-20
Add support for authentication via Unix socket

tenten8401:
Just thought it'd be cool to be able to authenticate with mysql via a unix socket. A unix socket checks the user that's trying to access it and does authentication that way.

Laravel has support built in already, I just needed to define it here to get it working on my install. Seems to work as expected.


tenten8401:
Here's what it looks like in action:

```
[root@Isaac-Dedi:/var/www/pterodactyl]# sudo -H -u nginx bash -c 'php artisan migrate --seed'
**************************************
*     Application In Production!     *
**************************************

 Do you really wish to run this command? (yes/no) [no]:
 > yes

Nothing to migrate.
Seeding: NestSeeder
Seeding: EggSeeder
[...]
```

I have to specifically sudo the migrate commands under the nginx user to get it working. If you try and do it from another user, mysql will reject the connection:
```
[root@Isaac-Dedi:/var/www/pterodactyl]# php artisan migrate --seed
**************************************
*     Application In Production!     *
**************************************

 Do you really wish to run this command? (yes/no) [no]:
 > yes


In Connection.php line 664:
                                                                                                                                                                           
  SQLSTATE[28000] [1698] Access denied for user 'nginx'@'localhost'
```

Most probably won't use it but it'd be cool to have it for the people that do.

tenten8401:
I'd like to say it'd fail in the same way if someone gave it a bad socket location as it would if it got a bad hostname.

Only difference between using a socket and a direct connection is you can restrict access to processes running under specific users. For the NixOS module I'm writing, the php-fpm pool for pterodactyl runs under the "pterodactyl" user, which can freely authenticate without a password to the "pterodactyl" mysql account. If I were to try and log into MySQL from a process running under another user, it'd reject the connection. This makes it much more secure than storing a MySQL password in plaintext.

I haven't yet tried it with both a unix socket _and_ a valid host to see what happens, but I'm pretty sure it prefers the socket over any host specified, unless the socket fails.

I'm also pretty sure it can be used in conjunction with a password, but doing so would very well defeat the purpose of a unix socket.

DaneEveritt:
I guess my only concern is that we're defaulting to a location for the socket, and if thats different, will that break existing installations?

tenten8401:
The location is fairly universal between distros from what I understand, but just incase I asked for someone else to test it for me.
image
image
Looks like it fails if the socket doesn't exist in it's current state however.

I'll see if I can test env('DB_SOCKET') and env('DB_SOCKET', '') later.

DaneEveritt:
Yeah we can't roll out a change like that hope that everyone's sockets are in the correct place.

tenten8401:
Yeah, I'm working on testing an alternative approach.

Sergzy:
Why not default it to NULL and then whoever wants to add it will add it? But we can alsready do that since it's just a one line change, so maybe no additional benefit there, except raising awarness.

DaneEveritt:
@Sergzy that would work as well. 

https://github.com/illuminate/database/blob/8875397ff80728b1040f58840eb314494b432303/Connectors/MySqlConnector.php#L106

Looks like it only tries to use the socket if its set and not empty, so defaulting to null will avoid breaking any existing installations.

PR #1244, opened 2018-07-21
Create nests.php

patricknasci:


DaneEveritt:
see https://github.com/pterodactyl/panel/pull/1245#issuecomment-406820351

PR #1245, opened 2018-07-21
Create pack.php

patricknasci:

PR #1247, opened 2018-07-21
Create user.php

patricknasci:
Tradução PT


DaneEveritt:
https://github.com/pterodactyl/panel/pull/1245#issuecomment-406820351

PR #1263, opened 2018-08-02
Add Simplified Chinese translation for the panel

lunarbunnys:
Hi,I am a Chinese student,I have made a Chinese translation for this panel.I don't know if it helps.


lancepioch:
Adding foreign languages is easy because of the language files. However the big issue with Chinese is that it's also right to left writing which doesn't really work for our current themes. This is what I'm assuming the new theme is, essentially adding support for RTL.

The unfortunate part of this is I don't think we'll be able to support this PR with a new addition of the RTL theme. In any case, this needs to be discussed with @DaneEveritt

BirkhoffLee:
Chinese is LTR.

lancepioch:
@BirkhoffLee Sorry if I mixed up the writing orientation, I had been told otherwise. If the Chinese written here is LTR, then what is the pterodactyl_zh-cn theme for then?

BirkhoffLee:
pterodactyl_zh-cn?

lunarbunnys:
I commit pterodactyl_zh-cn theme because admin page has some texts do not contain in lang folder.I want to finish my work completely.

DaneEveritt:
I won't accept a PR that basically doubles the number of template files that would have to be maintained, especially since I do not speak Chinese, and therefore would be completely unable to (and frankly I'm not interested in) maintaining two templates for the same page.

I'm fine to accept a PR with just the language files, and then a second PR that introduces the language setting in the Admin CP, but I won't maintain two sets of files in two languages.

lunarbunnys:
OK,I will delete it and wait until language setting be introduced.

DaneEveritt:
@lunarbunnys thanks, can you confirm this shows up correctly when you view the available languages under the admin settings?

lunarbunnys:
Yes,I have tested before commit,and they working fine.

PR #1281, opened 2018-08-13
change smtp.google.com to smtp.gmail.com

Bl4ckSpr4y:
read the title

### DIFF
--- a/resources/lang/cn/command/messages.php
+++ b/resources/lang/cn/command/messages.php
@@ -49,7 +49,7 @@ return [
     ],
     'environment' => [
         'mail' => [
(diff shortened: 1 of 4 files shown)


DaneEveritt:
This is unnecessary, it is a valid SMTP server and is simply to point the user in the correct direction for what should be entered there.

Bl4ckSpr4y:
actually it isn't
C:\Users\mrkrabs>ping smtp.google.com
Ping request could not find host smtp.google.com. Please check the name and try again.
but I get it that it's supposed to point in the correct direction, just saw someone in support earlier using smtp.google.com which is incorrect

jacobgc:
NGL a ping is not the best way to see if the URI is valid, however, you are right. There are no DNS records for smtp.google.com

PR #1117, opened 2018-04-27
Feature/dockerfile with SSL creation completed

quikstrike:
This includes the previous commits that create the Dockerfile, it's build process, as well as adding automatic Lets Encrypt SSL certificate creation and renewal.

PR #1285, opened 2018-08-14
Feature/add dockerfile

parkervcp:
A fully tested and complete docker setup for the panel. Should work for all versions of the 0.7.X+ versions.


.dev/docker/README.md
```diff
@@ -0,0 +1,69 @@
+# Pterodactyl Panel - Docker Image
+This is a ready to use docker image for the panel.
+
+## Requirements
+This docker image requires some additional software to function. The software can either be provided in other containers (see the [docker-compose.yml](docker-compose.yml) as an example) or as existing instances.
+
+A mysql database is required. We recommend [this](quay.io/parkervcp/mariadb-alpine) image if you prefer to run it in a docker container. As a non-containerized option we recommend mariadb.
```
schrej: I can't open that link, 403 unauthorized

.dev/docker/entrypoint.sh
```diff
@@ -0,0 +1,48 @@
...
+  echo -e "Generating key"
+  APP_KEY=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
+  echo -e "Generated app key: $APP_KEY"
+  echo -e "APP_KEY=$APP_KEY" > /app/var/.env
+
+  ln -s /app/var/.env /app/
+fi
+
+## check for DB up before starting the panel
+zshoecho "Checking database status."
```
schrej: Is this an actual command or meant to be just `echo`?

docker-compose.example.yml
```diff
@@ -0,0 +1,72 @@
...
+      ## Email settings
+      - "MAIL_FROM=noreply@your.domain.here"
+      - "MAIL_DRIVER=smtp"
+      - "MAIL_HOST=mail"
+      - "MAIL_PORT=1025"
+      - "MAIL_USERNAME=''"
+      - "MAIL_PASSWORD=''"
+      - "MAIL_ENCRYPTION=true"
+
+networks:
```
schrej: do we need that in the example?

.dev/docker/README.md
parkervcp: I forgot about this. Moving to use the stock mariadb image anyways.

.dev/docker/entrypoint.sh
parkervcp: I may have been typing in another window and not noticed.

docker-compose.example.yml
parkervcp: I like to have options there even if not being used. It could be commented out instead.

.dev/docker/entrypoint.sh
```diff
@@ -0,0 +1,48 @@
...
+## Ensure we are in /app
+
+cd /app
+
+## check for .env file and generate app keys if missing
+if [ -f /app/var/.env ]; then
+  echo "external vars exist"
+  rm /app/.env
+
+  ln -s /app/var/.env /app/
```
quikstrike: I am running an earlier version of the dockerized panel that has a very similar entrypoint.sh file and have a small suggestion to make. This line will cause the container to fail to restart after you have created and run the container once before, my fix was to simply add a `-f` flag to the command to force it to update.

One additional question, was having Lets Encrypt support out of the box not a goal for the container?

parkervcp: add a `-`f to what line?

Also I was looking at auto let's encrypt but haen't worked it yet. My older images were using caddy, which has auto ssl.

quikstrike: Oops sorry, thought I had highlighted a single line. The change would be on line 11 and change it to `ln -sf /app/var/.env /app/` otherwise it tries to create the symlink to the file and errors out because it had already been created from the first the entrypoint.sh script had been run.

I had submitted a PR that added a dockerfile that add's Let's Encrypt support https://github.com/pterodactyl/panel/pull/1117 . Most of the changes were done in the Dockerfile, entrypoint.sh, and the README.

DaneEveritt: I agree with @quikstrike on the `-f`, that screws me up quite frequently with scripted deployments.

parkervcp: There is nothing for it to force overwrite. The container doesn't have that file to mess up linking it. It shouldn't be an issue.

DaneEveritt:
This is missing the queue workers?

parkervcp:
It is missing queue workers. I will look into adding them.

parkervcp:
@DaneEveritt Queue workers added now.

PR #1320, opened 2018-09-10
Added uMod variable and depreciation notice to OxideMod

tenten8401:
See https://github.com/pterodactyl/images/pull/26

### DIFF
--- a/database/seeds/eggs/rust/egg-rust.json
+++ b/database/seeds/eggs/rust/egg-rust.json
@@ -33,9 +33,18 @@
             "rules": "required|string|max:40"
         },
         {
-            "name": "OxideMod",
-            "description": "Set whether you want the server to use and auto update OxideMod or not. Valid options are \"1\" for true and \"0\" for false.",
+            "name": "OxideMod - DEPRECIATED",
+            "description": "DEPRECIATED - PLEASE USE \"UMOD\" VARIABLE INSTEAD",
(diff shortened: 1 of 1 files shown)


DaneEveritt:
This is just confusing because if you ignore the deprecated variable and set the umod one it still behaves inconsistently because of the way the docker image is checking the variables.

Honestly its probably fine to just leave the environment variable name how it is, and just change the naming scheme on that. Its that or completely remove that variable and just break a few setups unfortunately.

You could also leave it how it is in this PR, but change the logic in the docker image to properly use the umod environment variable if provided, and fall back to oxide if not present? This still leads to issues where as soon as you add this variable you're going to have issues because the moment that server gets rebuild the new umod default value is going to be passed through, even if it differs from the oxide mod setting.

We might just need to hold off on pushing this egg change into the repo as it is until a breaking release cycle, and just do what I mentioned at the start of this wall of text which is change the wording on the `OXIDE` one to reflect that it is now umod.

tenten8401:
I'll see if I'm able to change the docker image to properly use the uMod variable, unless you'd rather me just change the naming scheme.

DaneEveritt:
You have to just change the naming because of the backwards break. If you only add the UMOD variable you'll default everyone to the default value for it, rather than the value they have for oxide, and theres not really a way to know what they want to do.

Issue #220, opened 2016-12-18
Remove multiple ports at a time


parkervcp:
When removing ports they need to be removed one at a time.

The option to select more than one port and mass remove would be a good improvement

Edbtvplays:
maybe a mass add feature aswell so you can select a range of ports that you would like to be added and it will add all the ports in bettween saves time if you have 10 + ports you need to add

parkervcp:
You can already add a range of ports. The limit is something like 2000 at a time.

parkervcp:
This was resolved. I'll close this out.

PR #1322, opened 2018-09-11
Allow deletion of multiple allocations at once

ayan4m1:
Relates to #220 

I have added checkboxes to each item in the node allocation view. I have also added a Mass Actions dropdown patterned after the one in the file manager. You can check multiple allocations, then delete them all with one click (well, one click to activate one click to confirm).

[Here](https://imgur.com/XeL8Knf) are [some](https://imgur.com/GSFRo5b) screenshots.

I have tried to re-use code and style where possible.

Feedback is appreciated.


public/themes/pterodactyl/js/frontend/nodes/allocation.js
```diff
@@ -0,0 +1,115 @@
+class Allocation {
```
DaneEveritt: Can't use any of this unless you add it to babel to compile into ES2015. :(

ayan4m1: Should I do that or just inline the code (sans class) in `allocation.blade.php`?

DaneEveritt: Can you just write it using ES5?

ayan4m1: Should be done.

DaneEveritt:
What happens with that popup when you select a ton of ports?

PR #1333, opened 2018-09-16
Google Analytics Input

ayan4m1:
Adds a `APP_GOOGLE_ANALYTICS_ID` which is then passed to the views as `$appGoogleAnalyticsId` - tracking code embeds for each layout.


resources/themes/pterodactyl/layouts/admin.blade.php
```diff
@@ -40,6 +40,17 @@
             <![endif]-->
         @show
     </head>
+    @if(isset($appGoogleAnalyticsId))
```
DaneEveritt: Can we just plop this into another blade template that gets included in these files rather than repeating the same logic throughout them all?

app/Providers/AppServiceProvider.php
```diff
@@ -29,6 +29,11 @@ public function boot()
 
         View::share('appVersion', $this->versionData()['version'] ?? 'undefined');
         View::share('appIsGit', $this->versionData()['is_git'] ?? false);
+        $googleAnalyticsId = env('APP_GOOGLE_ANALYTICS_ID');
+        if ($googleAnalyticsId) {
+            View::share('appGoogleAnalyticsId', $googleAnalyticsId);
```
DaneEveritt: I'd be fine with `$analyticsId` as the variable name here, and you can just always pass it through and make the blade do a `!empty($analyticsId)` check.

.env.example
```diff
@@ -8,6 +8,7 @@ APP_DELETE_MINUTES=10
 APP_ENVIRONMENT_ONLY=true
 LOG_CHANNEL=daily
 APP_LOCALE=en
+APP_GOOGLE_ANALYTICS_ID=
```
DaneEveritt: This should be referenced in the `config/pterodactyl.php` configuration file, and used in that manner in the code base, rather than direct calls to `env()`.

Doing the the way you are currently leads to issues with a cached configuration.

PR #1334, opened 2018-09-16
Fix "installed" server update

ayan4m1:
Currently, the daemon POSTs "installed" or "error" to `/daemon/install` on the panel. However, for some reason the `$server->save()` in the current code does not cause a SQL update to occur, and thus the install is stuck as "installing" even though the daemon has called back to report success or failure. I issue the update a slightly different way which does cause a SQL update.

### DIFF
--- a/app/Http/Controllers/Daemon/ActionController.php
+++ b/app/Http/Controllers/Daemon/ActionController.php
@@ -53,7 +53,7 @@ class ActionController extends Controller
 
         Server::where('id', $server->id)
             ->update([
-                'installed' => ($status === 'installed') ? 1 : 2
+                'installed' => ($status === 'installed') ? 1 : 2,
             ]);
 
         // Only fire event if server installed successfully.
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Daemon/ActionController.php
```diff
@@ -51,8 +51,10 @@ public function markInstall(Request $request)
             ], 403);
         }
 
-        $server->installed = ($status === 'installed') ? 1 : 2;
-        $server->save();
+        Server::where('id', $server->id)
```
DaneEveritt: This change breaks the rest of the code pathway because the installed state is never set on the existing model. I'm not sure what is leading you to believe this is broken, I've had it work fine in every test I've ever done with it?

DaneEveritt: I should clarify, finding the actual cause of the failure for it to persist to the database would be more ideal than just changing the code here.

PR #1340, opened 2018-09-20
Implemented rar files decompression

KewaiiGamer:
It seems that this feature has been requested a couple times already. It doesn't seem to be something that hard to implement so I thought it would be cool to do this as my first PR

I also added a PR to the Daemon!


lancepioch:
Apply your commit (133032d293f2e8f20a203d9c7c9dcbf053088ef2) on top of the `develop` branch.

DaneEveritt:
I will not be adding this feature to the panel at this time due to the breaking changes needed on the Daemon (which I am not looking to do right now).

PR #1349, opened 2018-10-04
Adds start of a mod manager.

gabeoliver-bicode:
This mod manager is only client side currently, they will need to know some laravel to actually set the location of where the addon goes and to add the actual things inside the modmanager.blade.php.


victionn:
lmfao

PR #1350, opened 2018-10-04
adds join button

gabeoliver-bicode:
This will add a join button for users to connect to their server.

### DIFF
--- a/resources/themes/pterodactyl/server/index.blade.php
+++ b/resources/themes/pterodactyl/server/index.blade.php
@@ -43,7 +43,7 @@
                 @can('power-restart', $server)<button class="btn btn-primary disabled" data-attr="power" data-action="restart">Restart</button>@endcan
                 @can('power-stop', $server)<button class="btn btn-danger disabled" data-attr="power" data-action="stop">Stop</button>@endcan
                 @can('power-kill', $server)<button class="btn btn-danger disabled" data-attr="power" data-action="kill">Kill</button>@endcan
-                <button class="btn btn-danger disabled"><a href="steam://connect/{{ $allocation->ip }}>Join</a></button>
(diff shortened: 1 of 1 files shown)


gabeoliver-bicode:
I could make a new database table for steam games, have the shortid of the servers that use steam automatically get put into that table so, it will only show to servers in that table.

gabeoliver-bicode:
Do you have any ideas @Bl4ckSpr4y

tenten8401:
Possibly a configurable "Join URI" in the egg that allows for variables

gabeoliver-bicode:
That is a good idea I will try and do that. Thanks @tenten8401

tenten8401:
Also, I wanted to point out, the button shouldn't be disabled by default unless you add JS to un-disable it once it's got data from the daemon / server's online

gabeoliver-bicode:
So what you are saying for the configurable thing is that they (the admins) should be able to set the url?

tenten8401:
Yeah, like an egg property of some sort. That way I could set the join URI to say "steam://connect{whatever_variable}" or something sort've like how you can set the startup command.

gabeoliver-bicode:
Yeah, that's what I thought you meant. I was just confirming, okay I will get on that.

gabeoliver-bicode:
I think I am going to add a new page when editing an egg for like "Console Configurations" or something like that and it will route to a page similar to the startup parameters thing. Where then you can change that URL.

DaneEveritt:
Please feel free to open a new PR when you have this in a more completed state.

lancepioch:
It's definitely a neat option that I think should be in. Obviously the logistics with determining if it's a Steam or other online service will be tricky, but good luck!

gabeoliver-bicode:
thanks

PR #1354, opened 2018-10-13
Update to latest version

Sir3lit:
Use latest version to prevent issues with invalid licensekey.dat file


Sir3lit:
Some back story, TeamSpeak 3 has a free limited use without the key, 3.1.1 seems to have issues with it causing an immediate crash, and makes it useless to someone attempting to use the free version, using latest build, 3.4.0, it seems to fix itself and runs without the key like it should.

JanDevDE:
@Sir3lit Just FYI, 3.3.0 would be high enough to fix it. (Not that it matters, just saying.)

PR #1363, opened 2018-10-22
Remove phraseapp because we no longer use it

lancepioch:
I'm pretty sure we only use the native Laravel translations now.


DaneEveritt:
Theres a middleware and some views doing things with it as well that should be removed.

lancepioch:
I believe I've purged it finally all through the power of the GH editor

DaneEveritt:
@schrej can you double check and make sure I didn't miss anything since you added this?

PR #1367, opened 2018-10-28
Updated TS3 Server egg

fabm3n:
New minimal Script based on bash.
ts3server.ini is not needed, because its only used when the server get started with the inifile=ts3server.ini parameter.
Also the license aggrement moved to the parameter.


DaneEveritt:
I don't mind the startup change, but changing the script to ubuntu seems overkill. The advantage of those alpine containers is that they're small, and quick to download if missing on the system.

PR #1377, opened 2018-11-11
Things needed for srcds games in this panel

BlackZon77:
Make fastdownloads and players live count, map choose/live map, live player list to srcds games ( csgo )
And the fastdownloads for csgo go search for all files from models, materials, sound,map and their specific extension files, like .mdl .mp3 .jpg .vtf .vtm etc, and just compress them in bz2, and send them to the specific folders in the webhost Ex.: /var/www/fastdl.yourawesomesite.com/ip-port/csgo/ [ models; materials; maps; sound ]
Add to variables, game type, game mode, max players, and the necessary things


Sir3lit:
This needs to be opened as an issue the fast dl part would be impossible, unless you run a web server within the docker container as they don't have access to anything outside of its own container. This is also VERY game specific so i highly doubt it will be added

DaneEveritt:
Do not open an issue for this, it will be closed like the other few that have been opened. I'm not adding Steam specific functionality, nor do I wish to maintain such a complex system.

PR #1386, opened 2018-11-14
Escape the database name when granting permissions to it

itsgatto:
These should be escaped as MySQL may take them in a different way than it should, specially the underscore.


lancepioch:
A database name shouldn't have an underscore (or percentage sign) in its name in the first place. Can we make sure that is consistent?

itsgatto:
> A database name shouldn't have an underscore (or percentage sign) in its name in the first place. Can we make sure that is consistent?

The databases created by Pterodactyl always have an underscore because the panel appends the s{serverid}_ prefix to the entered name when creating them.

DaneEveritt:
In what case does MySQL ever take a database name with an underscore in an unexpected manner?

lancepioch:
@DaneEveritt Underscores is a special type of wildcard in mysql if unescaped. It's probably easier to just disallow it from being entered in as part of the name.

DaneEveritt:
@lancepioch I've literally never seen that happen before, and I use underscores in plenty of database names? Plus we've been making databases with underscores for close to a year now with no issues?

PR #1387, opened 2018-11-14
Match original database password length when doing a password reset

itsgatto:
When creating a new database the password length is 16, but when performing a password reset, the new password is created with a length of 20 characters. While I'm not sure if this was done on purpose, this PR matches the 16 character length on password resets for consistency.


lancepioch:
Thanks for the contribution but this doesn't look like an issue.

PR #1397, opened 2018-11-26
Upgrade Laravel to 5.7

zKoz210:


Yomanz:
Do the various .env files need to be adjusted to allow for https://got-your.bigno.se/b748adc.png ??

zKoz210:
What do you prefer? Use the old environment variable or use new one with backward compatibility?
```
'default' => env('QUEUE_CONNECTION', env('QUEUE_DRIVER', 'database')),
```

DaneEveritt:
@zKoz210 because I plan on rolling this into the 0.7.X release it needs to not be a breaking change, so yeah your suggestion is what would be used. I went ahead and committed this.

PR #1402, opened 2018-12-03
Adding bulk reinstall command

JamsheedMistri:


ZeProf2Code:
That's would be cool if we could only reinstall servers with the specified egg

JamsheedMistri:
> That's would be cool if we could only reinstall servers with the specified egg

Hey, it's a start at least!

PR #1407, opened 2018-12-06
More correct link to connect nodes

zKoz210:
Some developers can change the routes at their discretion, but the connection always goes to the root


DaneEveritt:
This was set specifically because the number of users who mess up setting the root URL is insanely high, and they often forgot to put the URL scheme on it.

Using the index route ensures that the URL is generated correctly, and in general people are not editing that routes definition, nor should they really do that (to this specific route at least).

zKoz210:
It looks like a very strange decision, all for the sake of fools.

PR #1417, opened 2018-12-17
make source eggs use SERVER_IP instead of 0.0.0.0

AusHick:
This should be applied to all source egg instances that use +ip 0.0.0.0

### DIFF
--- a/database/seeds/eggs/source-engine/egg-custom-source-engine-game.json
+++ b/database/seeds/eggs/source-engine/egg-custom-source-engine-game.json
@@ -8,7 +8,7 @@
     "author": "support@pterodactyl.io",
     "description": "This option allows modifying the startup arguments and other details to run a custom SRCDS based game on the panel.",
     "image": "quay.io\/pterodactyl\/core:source",
-    "startup": ".\/srcds_run -game {{SRCDS_GAME}} -console -port {{SERVER_PORT}} +map {{SRCDS_MAP}} +ip 0.0.0.0 -strictportbind -norestart",
+    "startup": ".\/srcds_run -game {{SRCDS_GAME}} -console -port {{SERVER_PORT}} +map {{SRCDS_MAP}} +ip {{SERVER_IP}} -strictportbind -norestart",
     "config": {
         "files": "{}",
         "startup": "{\r\n    \"done\": \"gameserver Steam ID\",\r\n    \"userInteraction\": []\r\n}",


Sir3lit:
Why exactly does this need to be changed?

Sir3lit:
That's how it should work, yes. If it's not then an issue needs to be opened.

Sir3lit:
I think you're not understanding how Docker works? Docker only allows the container to access the IP you give it. So if you have 10.0.0.1 and 10.0.0.2 on the same server, then you only allocate 10.0.0.2:27015 to the cs:go server it SHOULD only let the server bind to 10.0.0.2:27015, and nothing else. If it is binding to both then this might be more of a docker error then a issue with the panel / daemon 

Hope that clears some stuff up

Revenge282:
Either way, if a game server instance is assigned a specific IP allocation, it should *explicitly* use that IP, if allowed (in this case, srcds allows it through +ip).  There is zero functionality difference for the game server if it is being told to bind itself to a proper address, and in cases where the node has been set to pass-through host interfaces, this will prevent the server from binding to all available interfaces on the node.

The Pterodactyl documentation also specifies steps to allow a node to pass-through the host machines interfaces to its containers.  Therefore, this change will support that documentation, as well as not interfere with other users who opted otherwise.

Sir3lit:
Is this issue only with srcds? Almost all the eggs bind to 0.0.0.0

Revenge282:
> Is this issue only with srcds? Almost all the eggs bind to 0.0.0.0

As far as I am aware, yes.  I specified source eggs in my PR as that is the only set I have tested with so far.  I would imagine this could be extrapolated to other engines as well if they allow bindings.

kaganus:
for this change to actually work, you have to configure your daemon so that containers do not use docker interface, but the host interface. otherwise containers will not be able to bind that specific IP. that's probably why the default config is to use 0.0.0.0.

you can make it work for srcds servers by configuring daemon to use the host network interface and creating your own egg duplicate and modify it to use {{SERVER_IP}} instead of 0.0.0.0.

this panel was originally designed for Minecraft, and while it can be used for hosting other games' dedicated servers, you are somewhat limited in functionality since it utilizes docker containers. 

for example you just need ~50Mb of files to install and run a Minecraft server, but ~18Gb of files for CS:GO. since every server will be run as a docker container, you cannot reuse the same 18Gb of data for all servers, and this will limit your disk usage and network bandwidth while installing/updating.

since repo owners do not want to dive into specifics for every (any) game, and want the panel to be more of a generic docker container panel, which you can use to deploy game servers with somewhat limited configuration, you have to do your own modifications to it to have a better support for the games you want to host

in short; docker = easy to manage and more secure, but limits things you can normally do without it (hence the security and easiness)

this panel uses docker, so it inherits these properties from docker

schrej:
Have you actually tested this?
If you did, you would have noticed that it in fact does interfere when users aren't using the host network with docker. The external IP is invisible within docker containers when the network overlay is used.

We could argue about adding an option to start specific servers with network=host and having a new special variable that resolves to the external IP or 0.0.0.0 depending whether network=host is enabled or not.

Bucky420:
for any future go getters i use a ipip tunnel to protect my home server from attacks i didnt want to have to make a docker just to make another tunnel i would rather use the tunnel on the host so i change the config to use host witch i think u should add a check box for but i only use it on the bungge instance. i bind to just the tunnel making it so its isloated in the tunnel port fowerding is done on the vps side of the ipip tunnel. and on a side note u can get srcds servers down to 200mb if you use a base install and use simbolic links for all files but the .so files u must copy :P 

this will replace the ip properly with the assined one 

{{server.build.default.ip}}

PR #1431, opened 2018-12-30
Add Minetest 0.4.17.1 egg

niansa:


DaneEveritt:
This would fit better in https://github.com/parkervcp/eggs probably. :)

I would make the version an environment variable so its not a fix version egg as well before submitting there.

PR #1433, opened 2019-01-01
Created Egg for Rust Staging Branch

smcfreak:
Not too sure what to do with "exported_at" etc.
I will leave that up to you.


parkervcp:
@smcfreak I think this is better suited for places like my [eggs repo](https://github.com/parkervcp/eggs).

PR #1449, opened 2019-01-23
Fix Forge Install Container

parkervcp:
frolvlad's oraclejdk image died and the stock openjdk image works fine for this,


DaneEveritt:
We're gonna need to get a release out to address this, will attempt to get one out in the coming days with the other fixes that have gone into develop thus far.

PR #1460, opened 2019-01-31
Add russian translation

stashenko:


parkervcp:
@stashenko Whatever this is it should have been its own branch that you are making these PR's from.

PR #1468, opened 2019-02-05
Fixed error handling in beginTransaction()

zKoz210:
Fixed display of errors in server creation

image


DaneEveritt:
Maybe I'm being dense, but doesn't the connection roll back automatically if it isn't explicitly committed? Haven't worked in this area of the code recently, so just trying to understand what is going on here and why the rollback is fixing the issue?

zKoz210:
If we do not make connection rollback, errors from Validator will not get into the session and will not be displayed to the user.

DaneEveritt:
So I cannot reproduce this at all, making it really hard to test this code and understand what is happening. I know @TrixterTheTux ran into a similar issue with logins not showing error messages because the transaction was rolling back the session data, which I think is happening here, but I can't get it to happen.

DaneEveritt:
Okay, ate my own words there, forgot I was using the redis driver for sessions.

What is happening is that those opened transactions don't get closed when the exception is thrown, thus the session data is never saved since it gets added into the transaction.

DaneEveritt:
Approached this at a global level in https://github.com/pterodactyl/panel/commit/114afb86461b13d20c5bf5dcb3497721af43d475

PR #1469, opened 2019-02-05
Update to latest Alpine

Sir3lit:
Updated alpine version used for install containers, so its only one version and not 4


Sir3lit:
Yes i did test the ones i updated, and they all installed and started

PR #1478, opened 2019-02-11
Dirty fix to add ability to i18nize Admin Control panel.

f0rb1d:
Dirty fix to add ability to i18nize Admin Control panel.


ZeProf2Code:
This is a massive change :C

DaneEveritt:
Sorry for the long delay, I'm not looking to i18n-ize the admin portion of the control panel at this time.

PR #1488, opened 2019-02-17
Added OAuth2 Sign In support

jer3m01:
### Functionality 
Once enabled the route `<domain>/auth/login/oauth2` and `<domain>/auth/login/oauth2/callback` will be available.
When a user goes to that link it'll either redirect them to the OAuth2 authorization url or log them in he if already authorized.
This does not affect in any way the default sign in method.
### Features (disabled if OAuth2 is disabled)
* OAuth2 Sign In
* OAuth2 in depth configuration command `p:environment:oauth2`
* Create an account using oauth2
* Convert an account to OAuth2 or back to default
* Disables default login for OAuth2 users
* Disables 2FA for OAuth2 users (enable it on your OAuth2 server/provider instead)
#### User
login img  
account img  
security img  
Obvious password reminder for OAuth2 users
sftp notice
#### Admin
admin user list  
OAuth2 user  
admin user view oauth2  
Non Oauth2 user   
admin user view non oauth2  
admin new user  
### Demo 
If you would like to test yourself (only for project team members) send me a message on discord with your github username and i'll create you an account and give you admin credentials.
https://pterodactyl.amnesia-network.com

EDIT: Re enabled password for OAuth2 users
EDIT 2: SFTP notice


lancepioch:
Thanks for the contributions. I think it's a good feature and is close to being finished.

Some important things to note:

1. We would prefer for only the `thephpleague/oauth2-client` package to be included. We can't expect your package wrapper to remain up to date as well as the official package.

2. "Convert this account into an OAuth2 account" - Instead of being a separate "type of account", a user should be allowed to login with their 3rd party provider or their password normally. Also there shouldn't be any reason a user can't have multiple Oauth providers and be able to disconnect or reconnect them.

3. This looks to require Pt users to setup and maintain their own Oauth providers which can be useful, but since the installation process is already lengthy enough I think it'd be best to include some default providers like Google, Facebook, Steam, maybe Github even.

app/Http/Controllers/Admin/UserController.php
```diff
@@ -189,6 +208,39 @@ public function update(UserFormRequest $request, User $user)
...
+     * Change a user to OAuth2/Normal on the system.
+     *
+     * @param \Illuminate\Http\Request $request
+     * @param \Pterodactyl\Models\User $user
+     * @return \Illuminate\Http\RedirectResponse
+     *
+     * @throws \Exception
+     * @throws \Throwable
+     */
+    public function put(Request $request, User $user)
```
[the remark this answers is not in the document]
DaneEveritt: It looks like this function shouldn't even be called put, or should be in it's own controller under a different route because it is tied specifically to updating OAuth credentials, not the user.

app/Http/Controllers/Admin/UserController.php
```diff
@@ -189,6 +208,39 @@ public function update(UserFormRequest $request, User $user)
...
+    {
+        if (is_null(env('OAUTH2_CLIENT_ID'))) {
+            return redirect()->route('admin.users.view', $user->id);
+        }
+
+        if (isset($user->getAttributes()['oauth2_id'])) {
```
DaneEveritt: `->getAttribute('oauth2_id')`

app/Http/Controllers/Admin/UserController.php
```diff
@@ -189,6 +208,39 @@ public function update(UserFormRequest $request, User $user)
...
+            $user->update(['oauth2_id'  => null]);
```
DaneEveritt: Please do not use the models to modify and save the user, they're difficult to unit test properly. You should look at the UserUpdateService or use the user repository to make quick changes like this.

app/Http/Controllers/Admin/UserController.php
```diff
@@ -189,6 +208,39 @@ public function update(UserFormRequest $request, User $user)
...
+            $user->save();
+        } else {
+            $oauth2_id = $request->only('oauth2_id')['oauth2_id'];
```
DaneEveritt: PSR standards don't define a specific naming convention, but they should all match, so in this case please switch anywhere using underscores to be camelCased when naming variables.

app/Http/Controllers/Auth/ForgotPasswordController.php
```diff
@@ -13,6 +13,28 @@ class ForgotPasswordController extends Controller
 {
     use SendsPasswordResetEmails;
 
+    /**
+     * Send a reset link to the given user.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\RedirectResponse|\Illuminate\Http\JsonResponse
+     */
+    public function sendResetLinkEmail(Request $request)
```
DaneEveritt: Why is this here, it appears to be exactly same as the function defined in the trait?

app/Http/Controllers/Auth/OAuth2Controller.php
```diff
@@ -0,0 +1,191 @@
+<?php
+
+namespace Pterodactyl\Http\Controllers\Auth;
+
+use OAuth2;
```
DaneEveritt: What package is giving you this global `OAuth2` namespace like this? @lancepioch is this what you were mentioning?

app/Http/Controllers/Auth/OAuth2Controller.php
```diff
@@ -0,0 +1,191 @@
...
+                }
+
+                $name_last = $user->getAttributes()['name_last'];
+                if (env('OAUTH2_LAST_NAME_KEY') != null) {
+                    $name_last = $resources[env('OAUTH2_LAST_NAME_KEY')];
+                }
+                try {
+                    $this->updateService->handle($user, compact('email', 'username', 'name_first', 'name_last'));
+                } catch (\Exception $e) {
+                    abort(500, $e->getMessage());
```
DaneEveritt: This should return a response, not use the `abort()` helper.

app/Http/Controllers/Auth/OAuth2Controller.php
```diff
@@ -0,0 +1,203 @@
...
+     * OAuth2Controller constructor.
+     *
+     * @param \Pterodactyl\Services\Users\UserCreationService $creationService
+     * @param \Illuminate\Contracts\Hashing\Hasher $hasher
+     * @param \Illuminate\Auth\AuthManager $auth
+     * @param \Pterodactyl\Contracts\Repository\UserRepositoryInterface $repository
+     * @param \Pterodactyl\Services\Users\UserUpdateService $updateService
+     * @param AlertsMessageBag $alert
+     */
+    public function __construct(UserCreationService $creationService, Hasher $hasher, AuthManager $auth, UserRepositoryInterface $repository, UserUpdateService $updateService, AlertsMessageBag $alert)
```
DaneEveritt: This should be multiple lines.

app/Http/Controllers/Auth/OAuth2Controller.php
```diff
@@ -0,0 +1,203 @@
...
+        $this->auth = $auth;
+        $this->repository = $repository;
+        $this->updateService = $updateService;
+        $this->alert = $alert;
+    }
+
+    public function login()
+    {
+        // If this feature is not enabled or the user is already signed in and isn't linking a new account redirect or abort
+        if ($this->auth->guard()->check() and ! session()->has('link_oauth2_driver')) {
```
DaneEveritt: Please try to avoid using these global functions where possible. I know `view` and `redirect` get used a bunch, but you don't need `session()` or `abort()`. You already have the request object available here, and you can throw `NotFoundHttpException`s to cause 404s. 👍

app/Http/Controllers/Auth/OAuth2Controller.php
```diff
@@ -0,0 +1,203 @@
...
+
+                return redirect()->route('auth.login')
+                    ->withErrors($errors);
+            }
+
+            $user = $this->auth->guard()->user();
+
+            // Get the user info from OAuth2
+            try {
+                $socialite_user = Socialite::driver($session_driver)->user();
```
DaneEveritt: Per PSR-2, variables should be camelCase'd to match the rest of the application.

app/Http/Controllers/Auth/OAuth2Controller.php
```diff
@@ -0,0 +1,203 @@
...
+        if (session()->has('link_oauth2_driver')) {
+            $session_driver = session()->get('link_oauth2_driver');
+            session()->forget('link_oauth2_driver');
+
+            // Check if authenticated
+            if (! $this->auth->guard()->check()) {
+                $errors = new MessageBag(['user' => [__('auth.failed')]]);
+
+                return redirect()->route('auth.login')
+                    ->withErrors($errors);
```
DaneEveritt: `->withErrors(new MessageBag(['user' => ...]))`

app/Http/Controllers/Auth/OAuth2Controller.php
```diff
@@ -0,0 +1,203 @@
...
+
+            $oauth2_socialite_id = $socialite_user->getId();
+
+            $new_ids = [];
+            $done = false;
+            // Replace if already exists
+            foreach (preg_split('~,~', $user->getAttributes()['oauth2_id']) as $id) {
+                if (Str::startsWith($id, $session_driver)) {
+                    $id = $session_driver . ':' . $oauth2_socialite_id;
+                    $done = true;
```
DaneEveritt: Should can break out of the loop at this point most likely, or is there a reason to continue through the remaining?

app/Http/Controllers/Auth/OAuth2Controller.php
```diff
@@ -0,0 +1,203 @@
...
+        try {
+            // Try to get the user
+            $user = $this->repository->findFirstWhere([['oauth2_id', 'LIKE', '%' . $session_driver . ':' . $oauth2_id . '%']]);
+
+            // Login
+            $this->auth->guard()->login($user);
+            if ($this->auth->guard()->check()) {
+                return redirect()->route('index');
+            }
+        } catch (\Exception $e) {
```
DaneEveritt: There should never be a blind catch, and in the rare case it is necessary a comment needs to be included as to why it is there.

app/Http/Controllers/Auth/OAuth2Controller.php
```diff
@@ -0,0 +1,203 @@
...
+
+                return redirect()->route('account');
+            }
+
+            $oauth2_socialite_id = $socialite_user->getId();
+
+            $new_ids = [];
+            $done = false;
+            // Replace if already exists
+            foreach (preg_split('~,~', $user->getAttributes()['oauth2_id']) as $id) {
```
DaneEveritt: `$user->oauth2_id` is going to be cleaner to understand for most people.

app/Http/Controllers/Base/AccountController.php
```diff
@@ -2,18 +2,21 @@
 
 namespace Pterodactyl\Http\Controllers\Base;
 
+use Illuminate\Support\Arr;
+use Illuminate\Support\Str;
 use Pterodactyl\Models\User;
 use Illuminate\Auth\AuthManager;
 use Prologue\Alerts\AlertsMessageBag;
-use Illuminate\Contracts\Session\Session;
+use Laravel\Socialite\Facades\Socialite;
```
DaneEveritt: No facade usage please.

app/Http/Controllers/Base/AccountController.php
```diff
@@ -51,8 +54,20 @@ public function __construct(AlertsMessageBag $alert, AuthManager $authManager, U
...
+                $split = preg_split('~:~', $id);
+                if (! empty($split[1])) {
+                    $oauth2_ids = Arr::add($oauth2_ids, $split[0], $split[1]);
+                }
+            }
+        }
+
         return view('base.account', [
             'languages' => $this->getAvailableLanguages(true),
+            'enabled_providers' => config('oauth2.enabled') == true ? $this->getEnabledProviderSettings() : '',
```
DaneEveritt: Strict comparisons only, but this can also completely skip the boolean checking, its either going to be a truthly value or evaluate to false anyways.

DaneEveritt:
As we discussed on Discord, this is just too big of a change for me to feel okay merging in right now, especially with all of the massive changes in the development branch right now. Let's have a discussion and come back to this once 0.8 is released and we have a better idea of how things should work.

VibeGAMESNL:
Shame we need this!

FoksVHox:
@VibeGAMESNL Look at #1964

PR #1501, opened 2019-02-24
Replace Spigot with Paper

parkervcp:
Replace Spigot as the default server with plugin support with paper.

Paper can be freely downloaded without the need to build the server jar every time.

This egg still allows for the download like to be supplied to download whatever jar a user wants.


aikar:
:+1: :taco:

parkervcp:
It's been requested we don't integrate this quite yet due to changes on the Paper side.

parkervcp:
@DaneEveritt I have updated this PR to use the new paper API to get versions. Feel free to review.

aikar:
https://raw.githubusercontent.com/parkervcp/eggs/master/minecraft_java/server.properties

Why this file? Should just let mojang add all the defaults, and only set ones that you really intend to override Mojangs default on (but none of them stand out to me?)

parkervcp:
@aikar The panel will edit the file on startup. Without that it needs to be restarted a few times to set the port correctly.

PR #1509, opened 2019-03-06
Match original database password length when doing a password reset

ZeProf2Code:
When creating a new database the password length is 16, but when performing a password reset, the new password is created with a length of 20 characters. While I'm not sure if this was done on purpose, this PR matches the 16 character length on password resets for consistency.

---
This is a duplicate of #1387


LeCodeCo:
Instead of making the password weaker, I think that both should be set to 20, rather than 16.

ZeProf2Code:
@DaneEveritt Please review when you have the time :)

parkervcp:
Pretty simple increase in security. LGTM.

PR #1512, opened 2019-03-07
Allow daemon to be configured on privileged ports

### LINKED ISSUE #1511: Allow Daemon Port Proxying
I want to proxy pass the daemon to be on a subdomain on port 443/80, instead of exposing a random port on the daemon server. 

Currently you can't set the listen port of the daemon and the port the panel will actually connect to different values.
Either add an option to add a port in the FQDN field in the node settings, or remove the lower bound restriction on the daemon port field (then the user would change the listen port manually in the daemon configuration).

tenten8401:
Closes #1511

I don't know if this is the only thing that needs to be changed or not


DaneEveritt:
But, why?

DaneEveritt:
Can't wait for people to screw this up 👍

DaneEveritt:
Please fix this PR, you did it against master

tenten8401:
Welp now I broke it and can't re-open it... I clearly need to learn to use git a little better.

PR #1516, opened 2019-03-08
Allow users to reinstall their servers

fuwwy:
This is a feature that allows users to reinstall their servers manually through a "Server Settings" page, adapted and renamed from the "Server Name" which is now part of this new page. I'm kinda noob at PHP so please let me know if I did anything wrong or if I need to change anything. I tested the feature and the panel in general and wasn't able to find any issues.

image


routes/server.php
```diff
@@ -18,13 +18,14 @@
 |
 */
 Route::group(['prefix' => 'settings'], function () {
+    Route::get('/', 'Settings\SettingsController@index')->name('server.settings');
     Route::get('/allocation', 'Settings\AllocationController@index')->name('server.settings.allocation');
-    Route::get('/name', 'Settings\NameController@index')->name('server.settings.name');
```
DaneEveritt: This is a breaking change to anyone who might be reliant on that name elsewhere in their code. Please include backwards support for this, and update the readme to include a deprecation notice on that name. :)

ZeProf2Code:
Maybe just add it in "Startup Parameters" ?

routes/server.php
checkium: I've pushed it. I'm not sure if I included the deprecation notice correctly so please review that 👍

checkium:
@ZeProf2Code I think the Server Settings page is a better option because it fits better and opens room for further additions. This is an example of itself: Sure, I could put it in Startup Parameters, but it would be "hacky" since reinstalling the server is not related to the startup parameters, however, a settings page is much more general and can include practically anything that involves managing and configuring the server but doesn't fit on any of the other pages.

ZeProf2Code:
https://hastebin.com/wiyezocumu.js I've added `@can('reinstall-server', $server)`

mayed505:
@checkium & @ZeProf2Code How are you adding `@can('reinstall-server', $server)` when that doesn't exist in the sub-users permission list? You need to create Actions for `reinstall-server? Anyways, once that's done add it in there (example):

https://github.com/pterodactyl/panel/blob/ad9ed5ea0049e9fd267648b5cafddd3b7123a6f5/app/Models/Permission.php

DaneEveritt:
I can't merge this into `pterodactyl:develop`, please switch the base to `pterodactyl:0.7-develop`

checkium:
> I can't merge this into `pterodactyl:develop`, please switch the base to `pterodactyl:0.7-develop`

Done

DaneEveritt:
Thank you for taking the time to create this feature, unfortunately I seriously dropped the ball and did not merge this in at any reasonable pace. Because 0.7 is in maintenance only mode I'm closing this PR in favor of features added to 0.8.

schrej:
I guess the server-side changes of this could be merged into 0.8, it would just require UI (and potentially an api endpoint) to go with it?

IIPoliII:
Any news on this feature ?

justdeleteaccount:
Is it planned to implement the function to allow users to reinstall the server?

PR #1520, opened 2019-03-13
External modding support

robertdrakedennis:
Pterodactyl currently doesn't provide any easy way to add modding support or extending without editing core files. This feature gives users their own namespace, provider, and route examples. This PR gives the user options to make future mods for their own projects while still being able to keep up to date with Pterodactyl with little to no hassle.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,8 @@
             "app/helpers.php"
         ],
         "psr-4": {
-            "Pterodactyl\\": "app/"
+            "Pterodactyl\\": "app/",
+            "Entrypoint\\": "entrypoint/"
         }
     },
     "autoload-dev": {
--- a/config/app.php
+++ b/config/app.php
@@ -184,6 +184,11 @@ return [
         Pterodactyl\Providers\ViewComposerServiceProvider::class,
 
         /*
+         * Add Entrypoint provider
+         */
+        Entrypoint\Providers\EntrypointServiceProvider::class,
+
+        /*
          * Additional Dependencies
          */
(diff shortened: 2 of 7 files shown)


schrej:
Is there a realistic way to just specify a "plugins" folder where panel extensions can be added in subfolders and get automatically loaded?

DaneEveritt:
@schrej most likely yes, assuming you write the logic in such a way that it just looks for every folder and a specific entry in said folder. Similar to how Laravel handles automatically loading service providers from composer deps. @arcdigital might know more on that front?

DaneEveritt:
Also this PR makes no sense, its just adding code that someone would need to manual edit anyways, and is using a non-Pterodactyl namespace as well?

arcdigital:
Yeah, I don't think we should merge this either. We appreciate the effort, but it doesn't belong in the core. I'm going to put together an "Example" package that we can host in a separate repo and show how you can use that to add functionality.

schrej:
We should try to make it possible to add modifications without touching the panel source, so it's easier to handle upgrades. Not sure how much would be necessary. Probably some hooks to override services, routes, templates etc. - or is that possible with laravel already?

DaneEveritt:
Its somewhat possible, but we would need more hooks or template sections most likely to make things easier. I know a lot of people just want to add things to the sidebar, so maybe we start with an example project that makes that possible, and we just see whats needed to get that working.

arcdigital:
yea, I added the menu hooks in a hacky way. will clean it up and PR that along with my example package.

PR #1527, opened 2019-03-19
Fix tooltips

mayed505:
This should fix the bootstrap's tooltip since it doesn't show properly for the user-end.

From: https://i.imgur.com/vBg3F9j.png
To: https://i.imgur.com/0zHf8IP.png


Sir3lit:
Wouldn't this be pointless as everything is changing in 0.8?

mayed505:
@Sir3lit You're right, I don't believe 0.8 had a specific date for release. It's there just in case v0.7.14 was released for example.
or
For users who are staying with the old design before 0.8 becomes official maybe? Not sure.

DaneEveritt:
Closing since develop is no longer the same branch as when you made this PR. Not that worried about it in the current version either. :)

PR #1535, opened 2019-03-26
Change Console Chart yAxes to start at 0 and calculate maximum depending on server config

MadddinTribleD:
Neither the cpu chart nor the memory chart use a minimum/maximum for the yAxes values.  
This results in the memory chart having +/-2MB of the servers memory and the cpu chart having negative values.


ZeProf2Code:
@DaneEveritt Can you review these changes?

ET-Bent:
image

+1 from my side. Works as expected and makes the graphs a lot more useful.

PR #1542, opened 2019-03-30
Changing recommended app

Ethereal739:
Changing recommended session driver from Redis to MySQL as Session management only works with MySQL.

### DIFF
--- a/app/Console/Commands/Environment/AppSettingsCommand.php
+++ b/app/Console/Commands/Environment/AppSettingsCommand.php
@@ -26,9 +26,9 @@ class AppSettingsCommand extends Command
     ];
 
     const ALLOWED_SESSION_DRIVERS = [
-        'redis' => 'Redis (recommended)',
+        'redis' => 'Redis',
         'memcached' => 'Memcached',
-        'database' => 'MySQL Database',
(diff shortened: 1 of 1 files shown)


Lunaversitay:
Reason they probably have redis as recommended is because it's way faster than MySQL in some cases and generally faster in almost all cases depending on the setup, I assume it's an attempt at a wide scale recommendation so people don't complain about their panel being slow.

Ethereal739:
In my opinion, if someone is using large scale deployment that MySQL can't handle, they should have a different setup at that point. For an average user, they often get confused as of why session management doesn't work and have to ask support to enable it. I think MySQL is more than good enough for most people.

Lunaversitay:
Wide scale as in they're trying to recommend what's best for the biggest margin of people using the app, people will good and bad databases, not how big x company is.

Redis is generally a better option because people are always doing weird stuff with their mysql servers like putting them on external machines and what not. If your MySQL server is on an external machine it's going to cause panel loading increases compared to Redis.

DaneEveritt:
Session management barely even works as is it, and is completely removed from 0.8. I'm not enabling MySQL as the default when redis works fine and is the recommended approach.

PR #1548, opened 2019-04-02
Replace 2FA QR Code generation

### LINKED ISSUE #1528: 2FA Image Doesn't Load, 502 From Google Charts API
* Panel or Daemon: Panel
* Version of Panel/Daemon: 0.7.13
* Server's OS: Ubuntu 18.04.2
* Your Computer's OS & Browser: Win 10, Chrome 73

When enabling 2FA from the /account/security page, the QR code image never loads. It sticks on "Loading QR Code". Upon doing some digging, the failure seems to be a 502 returned by Google's Charts API.

Related issue seems to indicate they've possibly EoL'd the service all together?! https://github.com/google/google-visualization-issues/issues/2721

**To Reproduce**
1. Go to /account/security
2. Click on Enable 2-Factor Authentication
3. See error

**Expected behavior**
I expect the QR code to display for scanning with my authenticator app.

saibotk:
Hello,
with this change i fixed #1528 

This is my first contribution / PR to a bigger Laravel / PHP project, so you're welcome to give me some tips for possible future commits or on how to improve this one.

The changes i made were pretty small, but there is something to discuss. The BaconQR package needs the imagick extension (also the default iconv extension has to be enabled) to render the image, so this would be a new requirement for all Pterodactyl instances. Also what would you suggest to handle installations missing the library, since i did not know what your convention is on missing php extensions (codewise).

Thanks
saibotk


composer.json
```diff
@@ -29,7 +29,8 @@
         "lord/laroute": "^2.4",
         "matriphe/iso-639": "^1.2",
         "nesbot/carbon": "^1.22",
-        "pragmarx/google2fa": "^2.0",
+        "pragmarx/google2fa": "^5.0",
+		"pragmarx/google2fa-qrcode": "^1.0.3",
```
DaneEveritt: Please fix the spacing on this line, not sure why it is off by one.

saibotk:
Done, the reason for the build failing should not be about my change i suppose.

PR #1573, opened 2019-04-29
Update ServersController.php

Nobody196:
(!$server->installed) echoes false which won't save to the database

### DIFF
--- a/app/Http/Controllers/Admin/ServersController.php
+++ b/app/Http/Controllers/Admin/ServersController.php
@@ -438,7 +438,7 @@ class ServersController extends Controller
         }
 
         $this->repository->update($server->id, [
-            'installed' => ! $server->installed,
+            'installed' => (!$server->installed ? 1 : 0),
         ], true, true);
 
         $this->alert->success(trans('admin/server.alerts.install_toggled'))->flash();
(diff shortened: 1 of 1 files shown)


ZeProf2Code:
This resolved my bug that I had for so long where I can't change the install state!

Nobody196:
> This resolved my bug that I had for so long where I can't change the install state!

🙏 thanks for catching that

lancepioch:
The code doesn't seem correct, but the fix doesn't quite fit either I think. Take a look: https://github.com/pterodactyl/panel/blob/0f8dcabb45aa80e1659bf35874fc048e6d0e7170/app/Models/Server.php#L90

It can be 0-2

Nobody196:
> The code doesn't seem correct, but the fix doesn't quite fit either I think. Take a look:
> 
> https://github.com/pterodactyl/panel/blob/0f8dcabb45aa80e1659bf35874fc048e6d0e7170/app/Models/Server.php#L90
> 
> It can be 0-2

Maybe I'm a bit confused but the fix can only echo 0 or 1 based on true of false

lancepioch:
Whoops, didn't even notice that. Looks good then.

DaneEveritt:
This change doesn't make sense, you can only get here if the server is installed or not, so all we're doing is toggling the state between true & false, which the existing code does just fine?

PR #1569, opened 2019-04-27
Added Disk to overview (issue:1547)

Sir3lit:
As requested, storage information has been added to the overview

https://github.com/pterodactyl/panel/issues/1547

image

Issue #1547, opened 2019-04-01
Show storage usage in the panel


ntimo:
**Is your feature request related to a problem? Please describe.**
Currently the only way to get the storage used by a server is to make a GET request to the `https://pterodactyl.app/api/client/servers/<ID>/utilization` API endpoint.

**Describe the solution you'd like**
The storage usage with the storage limit should be shown in the panel. This would make it super easy to know how much storage space is used. 

**Screenshot current interface**
<img width="1444" alt="Screenshot 2019-04-01 at 20 57 52" src="https://user-images.githubusercontent.com/6145026/55352444-25bbd980-54c1-11e9-9a0a-06627c1b1b17.png">

ntimo:
That sounds interesting, but from a consumer point of view that's a bit techy. I am personally fine opening Postman and asking the API. But it would still be nice to simply take a look at the awesome UI and see it :)

Sir3lit:
![image](https://justmyrandomstuff.com/img/IGMvxrn.jpg)

Good?

#1569

ntimo:
@Sir3lit This looks fantastic 🙌

PR #1578, opened 2019-05-02
Added Disk usage to admin/node server viewer

Sir3lit:
Goes with Issue (https://github.com/pterodactyl/panel/issues/1547)
Goes with PR (https://github.com/pterodactyl/panel/pull/1569)

Sorry I'm horrible at pr's and attempting, and failing, to use the GitHub desktop program


DaneEveritt:
This file is deprecated and needs to be removed from this branch. It is not used, so this PR is pointless (unfortunately). Please adjust the other be be based against `0.7-develop` if you'd like it merged into that branch.

PR #1597, opened 2019-05-31
Adding Navigation for Addons

MatthewSH:
With @arcdigital releasing his Pterodactyl mobile app, we got to talking about navigation and the best way to do it.

It kind of sparked this idea for me to add the ability for addons, that don't modify core files but instead hook into Pterodactyl through the standard service provider, to add navigation items.

After thinking through a few different ways, I think Arc and I both kind of were both on the same wavelength, in my opinion, about addons registering their routes with navigation when they register the routes themselves. This provides an easier way to do that while giving themers ability to still theme accordingly. It should be good as-is. If any adjustments or changes need to be made, then I can.

Adding a route to the navigation would be as easy as:
```php
resolve('UserNavigation')->add('Route Name', 'my-custom-route', 'fa fa-servers');
```
Last argument for icon is optional, if none is passed through it will default to cogs. I used route names instead of paths to ensure consistency across the application, plus...if you're not naming routes then you should be.

All methods in the `Navigation` class are documented with an example array commented out in the `$items` variable. I've aliased some other method names as well to ensure that even if you're coming from other patterns you're set. `find()` is aliased to `get()` so both work exactly the same.

EDIT: This is also for me too so I can hook into it in my own addons for the community. So it's not just for Arc. It can help encourage less core modification too and hopefully by providing little things like this we can encourage users to edit core files less and opt for composer based addons.

EDIT after 28cc40c90373dc4b23962f323284ed06c742aba8: I added the ability for tree items as well. To create a tree, you give it a name, id, and optional icon (defaults to gears). To add children is the same as adding an item except you give it a tree id first:
```php
resolve('UserNavigation')->addTree('My Tree', 'my-tree-id');
resolve('UserNavigation')->addChild('my-tree-id', 'Tree Route', 'my-tree-route');
```

### DIFF
--- a/app/Navigation.php
+++ b/app/Navigation.php
@@ -63,7 +63,7 @@ class Navigation
      */
     public function addTree(string $name, string $id, string $icon = null)
     {
-        if (!$this->has($id)) {
+        if (! $this->has($id)) {
             array_push($this->items, [
                 'id' => $id,
                 'name' => $name,
@@ -171,10 +171,10 @@ class Navigation
     private function setTreeItem(array $item)
     {
         $currentTrees = Arr::where($this->items, function ($v, $k) use ($item) {
-            return (array_key_exists('id', $v) && $v['id'] === $item['id']);
+            return array_key_exists('id', $v) && $v['id'] === $item['id'];
         });
 
-        foreach($currentTrees as $key => $tree) {
+        foreach ($currentTrees as $key => $tree) {
             $this->items[$key] = $item;
         }
     }
--- a/app/Navigation.php
+++ b/app/Navigation.php
@@ -1,0 +1,118 @@
+<?php
+
+namespace Pterodactyl;
+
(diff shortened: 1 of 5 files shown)


MatthewSH:
I went ahead and added ability to have tree items. This was a bit wonky and done in 28cc40c90373dc4b23962f323284ed06c742aba8. I had a quick issue with pushing items to the children array and it not persisting. I don't assign keys or anything to the arrays as navigation can have multiple of the same link. I can go back and make the array of items each have it's own reference key and it could make tree items a bit easier...but I don't think it would be worth the added complexity and possible issue with duplicate items with the same keys.

DaneEveritt:
I'm not merging a major new feature into the 0.7 branch, its just there so I can continue patching bugs that might arise, it is feature complete as it stands.

DaneEveritt:
This is better handled by making the templates hookable and using an external package to do things.

Issue #1568, opened 2019-04-27
POST /api/application/nodes/*/allocations returns an empty string, doesn't work without alias


Kushfield:
The API endpoint /api/application/nodes/*/allocations seems to have some issues for POST requests.

Not including an alias should be an option, but it results in a 500 error:
```
-bash-4.2$ curl -X "POST" "https://pterodactyl.my-servers.com/api/application/nodes/1/allocations" \
>      -H "Content-Type: application/json" \
>      -H "Authorization: Bearer key" \
>      -H "Accept: application/vnd.pterodactyl.v1+json" \
>      -d '{"ip": "123.123.123.123", "ports": ["10059"]}'
{
    "errors": [
        {
            "code": "ErrorException",
            "status": "500",
            "detail": "An error was encountered while processing this request."
        }
    ]
}
-bash-4.2$ 
```
> production.ERROR: ErrorException: Undefined index: alias in /var/www/pterodactyl/app/Http/Requests/Api/Application/Allocations/StoreAllocationRequest.php:43

If an alias is included, the allocation seems to be created fine, but the request doesn't return anything:
```
-bash-4.2$ curl -X "POST" "https://pterodactyl.my-servers.com/api/application/nodes/1/allocations" \
>      -H "Content-Type: application/json" \
>      -H "Authorization: Bearer key" \
>      -H "Accept: application/vnd.pterodactyl.v1+json" \
>      -d '{"ip": "123.123.123.123", "alias": "my-server.my-servers.com", "ports": ["10059"]}'
-bash-4.2$ 
```

JanDevDE:
Is an alias set within the panel? (Reason I ask that is because I wonder if this error will only be thrown if an alias is set.)

Kushfield:
> 
> 
> Is an alias set within the panel? (Reason I ask that is because I wonder if this error will only be thrown if an alias is set.)

I'm not sure what exactly you mean, but:
If I include an alias in the request, then the new allocation is created fine, and the alias is also shown in the panel. However, the response for the request is just an empty string (I would expect it to return information about the new allocation).
If I do not include an alias in the request, then it results in an error.

parkervcp:
It looks like an alias is required. The panel should be responding with a "code": "required" error and not a 500.

DaneEveritt:
The alias isn't actually required though, so... :|

DaneEveritt:
[...]

PR #1599, opened 2019-05-31
Possible fix for allocation creation issue (#1568).

MatthewSH:
Now, it's only a hunch...so if someone could replicate it that would be wonderful since I don't have a full setup with the panel/daemon that I could test on.

I noticed that when validating the request it doesn't actually require the IP alias, but it's setting the allocation alias when it's passing it along. Based on the log, it's failing at the request point because it's trying to set the variable `allocation_alias` with a value from the data array when it doesn't exist. So I've added a quick check with the request and a default on the service as well. So if something fails anywhere...it's default to `null`.

Reference issue #1568

### DIFF
--- a/app/Http/Requests/Api/Application/Allocations/StoreAllocationRequest.php
+++ b/app/Http/Requests/Api/Application/Allocations/StoreAllocationRequest.php
@@ -40,7 +40,7 @@ class StoreAllocationRequest extends ApplicationApiRequest
         return [
             'allocation_ip' => $data['ip'],
             'allocation_ports' => $data['ports'],
-            'allocation_alias' => $data['alias'],
+            'allocation_alias' => $data['alias'] !== null ? $data['alias'] : null,
         ];
     }
 }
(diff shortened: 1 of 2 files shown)


app/Http/Requests/Api/Application/Allocations/StoreAllocationRequest.php
```diff
@@ -40,7 +40,7 @@ public function validated()
         return [
             'allocation_ip' => $data['ip'],
             'allocation_ports' => $data['ports'],
-            'allocation_alias' => $data['alias'],
+            'allocation_alias' => $data['alias'] !== null ? $data['alias'] : null,
```
DaneEveritt: What is the point of this logic, its literally just the same as using just `$data['alias']`

MatthewSH: Honest to god, I really don't know/can't remember. It was like 3 months ago. I must've been just expanding the logic for testing and just ended up replacing the test with null and didn't shorten it. I'll commit a change for that. I don't recall there being any specific issue since I made the changes elsewhere.

Maybe it could've been because somewhere the logic was messing up? I'll see if I can pull up my project and double check.

DaneEveritt:
Fixed in https://github.com/pterodactyl/panel/commit/2198269a65456079e0ef54b919b2efba1c0a75ed

PR #1612, opened 2019-06-20
Fixed issue with gmod server

rtm516:
If the workshop collection is specified after the map and gamemode it can occasionally cause the server not to start. This moves the argument and fixes the issue.


DaneEveritt:
Please base this against `pterodactyl:0.7-develop`.

arcdigital:
Thanks for the PR, we've merged it in #1664 and gave you credit.

Issue #1613, opened 2019-06-20
Strange 404 on file edit


rtm516:
**Background (please complete the following information):**
* Panel or Daemon: Panel
* Version of Panel/Daemon: 0.7.13
* Server's OS: Ubuntu 18.04.2 LTS
* Your Computer's OS & Browser: Chrome 75.0.3770.100 on Windows 10

**Describe the bug**
I get an error 404
Please provide additional information too, depending on what you have issues with:
Panel: `php -v` (the php version in use). - PHP 7.2.19-0ubuntu0.18.04.1 

**To Reproduce**
1. Go to one of the servers
2. Choose file management
3. Pick a folder
4. Pick a sub-folder
5. Pick a file to edit
4. See error 404

**Expected behavior**
The file shows up when you click and you don't get an error 404.

**Further info**
The url that the page takes me to is `https://pterodactyl.url/server/server.id/files/edit/garrysmod%2Fcfg/server.cfg` but if i ajust this to `https://pterodactyl.url/server/server.id/files/edit/garrysmod/cfg/server.cfg` it works as expected.
After enabling `APP_REPORT_ALL_EXCEPTIONS=true` in the .env file I get 
```
[2019-06-20 22:44:41] production.ERROR: Symfony\Component\HttpKernel\Exception\NotFoundHttpException in /var/www/panel/public_html/vendor/laravel/framework/src/Illuminate/Routing/RouteCollection.php:179
Stack trace:
#0 /var/www/panel/public_html/vendor/laravel/framework/src/Illuminate/Routing/Router.php(636): Illuminate\Routing\RouteCollection->match(Object(Illuminate\Http\Request))
#1 /var/www/panel/public_html/vendor/laravel/framework/src/Illuminate/Routing/Router.php(625): Illuminate\Routing\Router->findRoute(Object(Illuminate\Http\Request))
#2 /var/www/panel/public_html/vendor/laravel/framework/src/Illuminate/Routing/Router.php(614): Illuminate\Routing\Router->dispatchToRoute(Object(Illuminate\Http\Request))
#3 /var/www/panel/public_html/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php(176): Illuminate\Routing\Router->dispatch(Object(Illuminate\Http\Request))
[...]

PR #1617, opened 2019-06-23
Fixes file 404 error, closes pterodactyl#1613

rtm516:
Fixes the issue referenced in #1613


MatthewSH:
Have you tested this for any vulnerabilities considering you're removing the url encode method?

DaneEveritt:
The urlencode isn't for security, its to prevent browsers from struggling to correctly handle files with weird characters. The output is technically already escaped by blade's render engine.

schrej:
The script Matthew suggested wouldn't get escaped by blade. Those are all regular characters that can easily be present in a html document. Escaping mostly boils down to `<` and `>`.

For javascript to work in the `href` it needs to *start* with `javascript:`, i've quickly tested the string below and that doesn't show the alert.
```html
<a href="/test;javascript:alert('Test')">test</a>
```
As we start the `href` part with a path, we should be fine, `javascript:` can't appear at the beginning.

lancepioch:
Blade already escapes everything by default, so the question is does it escape too much and break at all? Or does the average filename still work correctly.

DaneEveritt:
@rtm516 please switch the base for this PR from pterodactyl:develop to `pterodactyl:0.7-develop`

PR #1622, opened 2019-06-24
Dutch Translation

Kudjoch:
When i was re-installing the Pterodactyl panel, i find out the Dutch translation was out-dated.
Somethings where in Dutch, something where in english.

So, here it is, the new dutch translation
[nl.zip](https://github.com/pterodactyl/panel/files/3322308/nl.zip)


DaneEveritt:
This is not the proper way to open a PR and we don't accept patches in the form of unknown zip files.

PR #1627, opened 2019-07-05
Update egg-teamspeak3-server.json

robingroppe:
- Change to official Teamspeak repo 
- Change to alpine specific tarballs
- Update to latest Teamspeak Server version

### DIFF
--- a/database/seeds/eggs/voice-servers/egg-teamspeak3-server.json
+++ b/database/seeds/eggs/voice-servers/egg-teamspeak3-server.json
@@ -17,7 +17,7 @@
     },
     "scripts": {
         "installation": {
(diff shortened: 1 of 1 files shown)


parkervcp:
Also seeing as they now have a server.json file that will tell you what releases are out.

```
apk add --no-cache tar curl jq

if [ -z ${TS_VERSION} ] || [ ${TS_VERSION} == latest ]; then
    TS_VERSION=$(wget https://teamspeak.com/versions/server.json -qO - | jq -r '.linux.x86_64.version')
fi

cd /mnt/server


echo -e "getting files from http://files.teamspeak-services.com/releases/server/${TS_VERSION}/teamspeak3-server_linux_amd64-${TS_VERSION}.tar.bz2"
curl http://files.teamspeak-services.com/releases/server/${TS_VERSION}/teamspeak3-server_linux_amd64-${TS_VERSION}.tar.bz2 | tar xj --strip-components=1
```

robingroppe:
I have to say that I don't like the approach to switch to a image with a larger footprint instead of grabbing the correct binary.
The alpine build is there for a reason... ;P

Nice to know there ist a version grabber, i was about to so the same this evening

parkervcp:
I have updated the image not knowing they pushed an alpine version. But I will also call out that while it's 5 times larger it's still under 100mb, and arguably easier to maintain.

The other java images moved to a debian base image as well a while ago.

DaneEveritt:
Closing this in favor of #1628

PR #1628, opened 2019-07-05
update teamspeak egg

parkervcp:
Install Script
    Update to use --no-cache for apk and add jq.
    Update to use the teamspeak file servers.
    New handler for latest release version using the teamspeak json available publicly.
New Variables
    Variable for setting the FileTransfer port
New Start command
    Use ts3server executable instead of the shell script.

This would supersede #1627 as it would replace getting versions with getting the latest.

### DIFF
--- a/database/seeds/eggs/voice-servers/egg-teamspeak3-server.json
+++ b/database/seeds/eggs/voice-servers/egg-teamspeak3-server.json
@@ -3,21 +3,21 @@
     "meta": {
         "version": "PTDL_v1"
     },
-    "exported_at": "2019-05-01T16:35:15+00:00",
+    "exported_at": "2019-07-05T11:59:29-04:00",
     "name": "Teamspeak3 Server",
     "author": "support@pterodactyl.io",
     "description": "VoIP software designed with security in mind, featuring crystal clear voice quality, endless customization options, and scalabilty up to thousands of simultaneous users.",
-    "image": "quay.io/parkervcp/pterodactyl-images:base_debian",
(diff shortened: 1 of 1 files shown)


robingroppe:
I think as this egg still runs alpine, the specific teamspeak alpine bin is required to avoid the glibc error, isn't it?

parkervcp:
Please re-check the image that is used.

It only uses alpine to download the files.

is uses `quay.io\/parkervcp\/pterodactyl-images:base_debian` to run

JanDevDE:
@parkervcp @robingroppe is partially right. New servers will be installed and used with the new image, however, to update an already existing TeamSpeak server, you usually click on the "reinstall" button with the new version written in the server settings. In that situation, the old image is going to be used and the server will crash on startup.

image

parkervcp:
While I i understand that old servers will still have the issue. We should be looking forward to using waht is most likely going to be supported longer for new server configs.

We will just need to let people know that they will need to update the iamge used to run their servers when updating the server.

JanDevDE:
> While I i understand that old servers will still have the issue. We should be looking forward to using waht is most likely going to be supported longer for new server configs.
> 
> We will just need to let people know that they will need to update the iamge used to run their servers when updating the server.

I mean the best option would be to modify the panel source code to automatically update build configurations of existing servers, but because I doubt that we're doing that, I can agree with you! :)

arcdigital:
due to some annoying git stuff we had to re-do this update in #1652 which is now merged and will go out in the v0.7.15 release. I gave you credit on the commit.

PR #1633, opened 2019-07-10
Update teamspeak egg to newest version with new download server

JanDevDE:
TeamSpeak has now their own download server. Dl4players is no longer available for new server versions.


JanDevDE:
Duplicate of #1628

PR #1643, opened 2019-07-18
Fix #1526 - Bulk Power

### LINKED ISSUE #1526: Bulk Power fails with multiple nodes
* Panel or Daemon: Panel
* Version of Panel/Daemon: 0.7.13 
* Server's OS:  Ubuntu 18.04.2 LTS 4.15.0-45-generic
* Your Computer's OS & Browser: Unrelated (Artisan issue)

**Describe the bug**
Bulk power command failing with multiple nodes
Please provide additional information too, depending on what you have issues with:
PHP 7.2.15-1+ubuntu16.04.1+deb.sury.org+1

**To Reproduce**
1. Have multiple daemon nodes
2. Have at least 1 server on each
3. Try to bulk power with artisan with no nodes / servers specified, i.e p:bulk-power:restart
4. Observe as the first node is contacted for all servers, resulting in a few 404s

**Expected behavior**
The bulk power command should properly determine which node a server belongs to and use the correct URL for API calls

MatthewSH:
Upon noticing a pattern in Discord support and further conversations with Trixter in #supporters. We determined this would be a good fix for some of the issues happening when re-using the power repository over multiple nodes.

This PR makes sure that the node id that's current set is the same node id of the server being set. If it's not, it will update it. This ensure that the proper HTTP client is being passed back, which should fix #1526.

### DIFF
--- a/app/Repositories/Daemon/BaseRepository.php
+++ b/app/Repositories/Daemon/BaseRepository.php
@@ -81,6 +81,10 @@ abstract class BaseRepository implements BaseRepositoryInterface
     public function setServer(Server $server)
     {
         $this->server = $server;
+        
+        if ($this->getNode()->id !== $server->node_id) {
+            $this->setNode($this->getServer()->getRelation('node'));
+        }
 
         return $this;
     }
--- a/app/Repositories/Daemon/BaseRepository.php
+++ b/app/Repositories/Daemon/BaseRepository.php
@@ -81,7 +81,7 @@ abstract class BaseRepository implements BaseRepositoryInterface
     public function setServer(Server $server)
     {
         $this->server = $server;
-        
+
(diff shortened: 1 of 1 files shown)


app/Repositories/Daemon/BaseRepository.php
```diff
@@ -82,6 +82,10 @@ public function setServer(Server $server)
     {
         $this->server = $server;
 
+        if ($this->getNode()->id !== $server->node_id) {
```
DaneEveritt: This should be happening in the `getHttpClient` function I think, otherwise you can still chain this backwards and defeat the purpose of this check.

DaneEveritt: Also can you make sure to get a test case added that catches this.

PR #1645, opened 2019-07-19
Add ALLOCATION_LIMIT to env

Kurounin:
Add ALLOCATION_LIMIT to environment variables and use it in servers where players/slots is usually restricted directly by host and not by allocated resources (like in the case of minecraft or rust)

### DIFF
--- a/database/seeds/eggs/source-engine/egg-garrys-mod.json
+++ b/database/seeds/eggs/source-engine/egg-garrys-mod.json
@@ -8,7 +8,7 @@
     "author": "support@pterodactyl.io",
     "description": "Garrys Mod, is a sandbox physics game created by Garry Newman, and developed by his company, Facepunch Studios.",
     "image": "quay.io\/pterodactyl\/core:source",
-    "startup": ".\/srcds_run -game garrysmod -console -port {{SERVER_PORT}} +ip 0.0.0.0 +map {{SRCDS_MAP}} +gamemode {{GAMEMODE}} -strictportbind -norestart +sv_setsteamaccount {{STEAM_ACC}} +host_workshop_collection {{WORKSHOP_ID}} +maxplayers {{MAX_PLAYERS}}  -tickrate {{TICKRATE}}",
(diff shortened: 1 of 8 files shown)


database/seeds/eggs/source-engine/egg-garrys-mod.json
```diff
@@ -8,7 +8,7 @@
     "author": "support@pterodactyl.io",
     "description": "Garrys Mod, is a sandbox physics game created by Garry Newman, and developed by his company, Facepunch Studios.",
     "image": "quay.io\/pterodactyl\/core:source",
-    "startup": ".\/srcds_run -game garrysmod -console -port {{SERVER_PORT}} +ip 0.0.0.0 +map {{SRCDS_MAP}} +gamemode {{GAMEMODE}} -strictportbind -norestart +sv_setsteamaccount {{STEAM_ACC}} +host_workshop_collection {{WORKSHOP_ID}} +maxplayers {{MAX_PLAYERS}}  -tickrate {{TICKRATE}}",
```
Nobody196: Why did you remove +maxplayers {{MAX_PLAYERS}} to add allocation limit to do the same thing?

DaneEveritt:
You do realize allocation limit is not a player count, its literally the number of ports that a user is allowed to assign to their services.

parkervcp:
I was wondering this whole time what they were trying to do here when this could have all been done by just adding a variable to their egg...

database/seeds/eggs/source-engine/egg-garrys-mod.json
Kurounin: To have a consistent way of managing the maximum allowed players on a server, instead of using allocation_limit on a server, max_players on another, maximum_slots on another and so on.

Kurounin:
> is not a player count, its literally the number of ports that a user is allowed to assign to their services.
In that case I misunderstood this variable. I thought it was supposed to provide a common way of managing the number of slots/players/users allowed on a specific server.
Then should I just add separate variables for each type of server to manage this?

parkervcp:
Yes. This is handled at the egg variable level and not the allocations level.

Kurounin:
Closing as per the discussion this was not the intended use of allocation limit.

PR #1665, opened 2019-08-01
remove paper query port again

parkervcp:
I realized after it was accepted I left in the server query.

This is to remove it.


DaneEveritt:
🐛

PR #1667, opened 2019-08-03
Fixed RegEx validate database password

zKoz210:
Failed asserting that [aIIXVo8WqAQ4uhvN.!uZtMeQ] contains 2 to 6 special characters.

Test:
```
<?php

$pattern = preg_quote(' !"#$%&\'()*+,-./:;<=>?@[\]^_`/{}|\|~', '#');

preg_match_all("#[{$pattern}]#", "aIIXVo8WqAQ4uhvN.!uZtMeQ", $matches, PREG_SET_ORDER);
var_dump(count($matches));
var_dump(count($matches) >= 2 && count($matches) <= 6);

preg_match_all("#[{$pattern}]#", 'he||o wor|d', $matches, PREG_SET_ORDER);
var_dump(count($matches));
var_dump(count($matches) >= 2 && count($matches) <= 6);

preg_match_all("#[{$pattern}]#", 'djsjKjffEf,mmddmd', $matches, PREG_SET_ORDER);
var_dump(count($matches));
var_dump(count($matches) >= 2 && count($matches) <= 6);
```

Result:
```
int(2)
bool(true)
int(4)
bool(true)
int(1)
bool(false)
```

### DIFF
--- a/tests/Unit/Services/Databases/DatabasePasswordServiceTest.php
+++ b/tests/Unit/Services/Databases/DatabasePasswordServiceTest.php
@@ -60,7 +60,7 @@ class DatabasePasswordServiceTest extends TestCase
         $this->dynamic->shouldReceive('set')->with('dynamic', $model->database_host_id)->once()->andReturnNull();
 
         $this->encrypter->expects('encrypt')->with(m::on(function ($string) {
-            preg_match_all('/[!@+=^-]/', $string, $matches, PREG_SET_ORDER);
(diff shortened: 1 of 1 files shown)


lancepioch:
Thank you for the contribution, but periods aren't special characters. Here's the list: https://www.owasp.org/index.php/Password_special_characters

I think it'd be fine if you wanted to cram all those in.

DaneEveritt:
This actually makes the test incorrect, there are only specific special characters I'm using in the passwords, all that was missing was a period in the list.

PR #1668, opened 2019-08-04
Fixed TooManyLoginAttempts not work correctly

zKoz210:
We use username(), so the cache key will be created correctly

image

Variables have also changed:
image

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -20,8 +20,6 @@ class LoginController extends Controller
 {
     use AuthenticatesUsers;
 
-    const USER_INPUT_FIELD = 'user';
-
     /**
      * @var \Illuminate\Auth\AuthManager
      */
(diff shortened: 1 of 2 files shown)


zKoz210:
@DaneEveritt It's about security, I think it's important.

DaneEveritt:
Yes, but it isn't actually causing a security issue, its just using Laravel's (sane) defaults right now rather than our higher ones. I'll review this when I have time, but it isn't urgent.

zKoz210:
It doesn't work at all because username cannot be obtained.

DaneEveritt:
You'll get a null value, thus making the key `|<ip>` which will still work, albeit locking all users and not just a single one on that IP address.

lancepioch:
* Why did you switch the constant to a function?
* Not sure if it's safe for an unsanitized string to make up the key in the cache
* Why change the lockout time from 120 to 2?

zKoz210:
1) Because the function takes data from username()
2) default by Laravel 
3) Now time in minute, not seconds

lancepioch:
This looks good to me after reviewing the traits. Magic and Laravel together

https://github.com/laravel/framework/blob/54bff8bd156c7817d5f32a4696474c72a2154f45/src/Illuminate/Foundation/Auth/ThrottlesLogins.php#L23-L25
https://github.com/laravel/framework/blob/3500e77d6ae85953e0b94b78530239a45231cc61/src/Illuminate/Foundation/Auth/AuthenticatesUsers.php#L94

PR #1674, opened 2019-08-11
Fixes file 404 error, closes pterodactyl#1613

rtm516:
Fixes the issue referenced in #1613

Recreation of pull request https://github.com/pterodactyl/panel/pull/1617 on correct branch


lancepioch:
Haha wow, I think we should have a function for this in the model or something like that?

It's not your fault rtm, but it shouldn't be this difficult to follow along with all this cramped logic directly in the view.

DaneEveritt:
@lancepioch its pointless to change anything at this point, its all gone in 0.8 anyways.

rtm516:
@DaneEveritt is 0.8 a complete rewrite? and if so what branch? and is there an estimated release date?

DaneEveritt:
Its in develop, and there is no set release date. I'll still merge this in, I was more speaking to making any major changes.

PR #1699, opened 2019-09-08
Allow alerts to be dismissed

lancepioch:


ZeProf2Code:
Allow validations alert to be dismissed too would be good no?

PR #1707, opened 2019-09-23
Request for username validation rule change.

Azoraqua:


DaneEveritt:
Please explain your reasoning for this change.

DaneEveritt:
The username field is set as a requirement like this because we use it to identify SFTP users who are logging in which are seperated with a period. You've also made usernames even more restrictive by enforcing a length limit in the regex.

Theres no reason a username should need to start with a non-alpha-numeric character.

P.S. - In the future, you should provide your reasoning for a PR, leaving it empty leaves me to guess at your intentions, and is likely to just be flat denied.

PR #1709, opened 2019-09-28
Fix french translation

Sykursen:


DaneEveritt:
Thank you for taking the time to create this feature, unfortunately I seriously dropped the ball and did not merge this in at any reasonable pace. Because 0.7 is in maintenance only mode I'm closing this PR in favor of features added to 0.8.

PR #1713, opened 2019-09-30
Dockerfile: Switch to php:7.2-fpm-alpine

BirkhoffLee:
It's usually a better practice to use an official supported Docker image (in this case, php) so it's best optimized and supported.

Not yet tested, but it should work.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,19 @@
-FROM alpine:3.8
+FROM php:7.2-fpm-alpine
 
 WORKDIR /app
 
-RUN apk add --no-cache --update ca-certificates certbot nginx dcron curl tini php7 php7-bcmath php7-common php7-dom php7-fpm php7-gd php7-mbstring php7-openssl php7-zip php7-pdo php7-phar php7-json php7-pdo_mysql php7-session php7-ctype php7-tokenizer php7-zlib php7-simplexml php7-fileinfo supervisor \
-    && curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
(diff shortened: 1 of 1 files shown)


parkervcp:
I have already been working on this change in my personal repo

DaneEveritt:
@parkervcp awesome, but can you work with @BirkhoffLee here since he has opened the PR already? No sense in closing it out or anything if its already been opened, plus I agree with this change.

Were there any specific "gotchas" we need to address before merging this?

parkervcp:
Let me finish my updated config files that I have been working on for some new support I am adding.

Then I can build test images based on this dockerfile.

DaneEveritt:
@parkervcp @BirkhoffLee any status updates here?

saibotk:
I can confirm that this image works, as i got my own fork on @BirkhoffLee 's pterodactyl-panel-docker repository, which also uses this Dockerfile. I am already using this for months now, without any problems and also with the newest panel version (here: https://github.com/saibotk/pterodactyl-panel-docker/blob/master/services/php/Dockerfile). I hope this information helps you :+1:

parkervcp:
I have already started on a dockerfile for 0.8 that will use the php standalone containers. I am looking at other possible things to add as well.

DaneEveritt:
Well, I'm not going to block a valid PR for that reason.

gpmidi:
Too bad this PR broken the docker builds for the panel

TrixterTheTux:
Then you'd want to open an issue with the details [here](https://github.com/pterodactyl/panel/issues/new/choose).

PR #1723, opened 2019-10-15
Fix for MySQL 8

Sir3lit:
So those running MySQL 8 don't get the syntex error.


Sir3lit:
God im horrible at life.

PR #1725, opened 2019-10-16
Update .travis.yml

andreybolonin:


DaneEveritt:
As a general repository policy I will not accept PR's that contain no supporting context. Please re-open this PR when you have provided that context -- including why you are opening it and what it will accomplish.

PR #1727, opened 2019-10-20
Added a file manager build script for Windows.

ZeProf2Code:


ZeProf2Code:
Yes, this was a mistake. I fixed it :D

DaneEveritt:
The file manager and entire build process has been rebuilt in 0.8, so I won't be merging this as it is unnecessary for 0.7. Thanks for looking into it though.

We also generally do not support development of any type on windows, and provide a vagrant environment specifically to have a consistent development environment for our products.

Issue #1731, opened 2019-10-21
Can't edit Daemon Server File Directory from Android device


WGOS:
**Overall information**
* Panel
* Version: 0.7.15
* Server's OS: Ubuntu 18.04.3
* Your Computer's OS & Browser: Android 9 ( w/ Samsung's OneUI 1.0), Chrome/Firefox

**Describe the bug**
On Android device (Samsung Galaxy S9+ in my case), I can't edit 'Daemon Server File Directory' while creating new node even in PC mode

**To Reproduce**
1. Go to 'Admin panel'
2. Then go to 'Nodes'
3. Click 'Create new'
4. Try to change 'Daemon Server File Directory' now

**Expected behavior**
Expected ability to change default value
Video with bug recorded: https://youtu.be/_r8Fjul1YLc

Sir3lit:
It should be noted that you can't edit it on any device.

Sir3lit:
Reproducible, Note 10+

devil38:
I have same issue too

PR #1732, opened 2019-10-21
Fix for new node on mobile 

Sir3lit:
Fixes https://github.com/pterodactyl/panel/issues/1731


devil38:
Screenshot_2019-10-22-00-59-44

PR #1743, opened 2019-11-11
i18nize add db card

ZeProf2Code:


ZeProf2Code:
it was already so why not using it haha

DaneEveritt:
0.7 is deprecated and not accepting any new PRs aside from critical changes. Thanks, but won't be merging this.

PR #1752, opened 2019-11-20
Add Player Slots to Vanilla config

chronicals:
Enforce the set player slots for the server.
As of right now, that setting is being ignored for the server.


parkervcp:
1. Learn how to use git properly please.
2. There is no variable for MAX_PLAYERS for you trying to set the value isn't helping things.
3. I don't feel like we should be forcing those settings on hosts. (I defend this in my own eggs repo)

PR #1763, opened 2019-12-08
Update steamcmd and copy steamclient.so for x64 as well

Kurounin:
Update all source-engine eggs to:
- download a newer **steamcmd** from 2018, instead of the old 2013 version
- copy **steamclient.so** from linux64 directory to fix the following error when starting the service `[S_API FAIL] SteamAPI_Init() failed; unable to locate a running instance of Steam, or a local steamclient.so.`


parkervcp:
I might propose a different solution for this.

Give me a bit to look into something.

DaneEveritt:
closed in favor of #1766.

PR #1766, opened 2019-12-08
Update steamcmd install scripts

parkervcp:
This is to consolidate the install script that is used on all the steamcmd servers to use the srcds_appid

See the script here - https://github.com/parkervcp/eggs/blob/master/scripts/steamcmd_installer.sh

This takes the changes in #1763 and expands on that organizing all the scripts the same way.

Changes made.
Update steamcmd install scripts to use new install script
Update steamcmd install image to Ubuntu 18.04
Add extra ARK variables


database/seeds/eggs/source-engine/egg-ark--survival-evolved.json
```diff
@@ -3,26 +3,35 @@
     "meta": {
         "version": "PTDL_v1"
     },
-    "exported_at": "2019-02-13T12:30:54-05:00",
+    "exported_at": "2018-10-29T20:51:32+01:00",
     "name": "Ark: Survival Evolved",
-    "author": "support@pterodactyl.io",
+    "author": "dev@shepper.fr",
```
DaneEveritt: Author of these shouldn't change.

parkervcp: My bad. I swear I caught all of those.

ZeProf2Code:
@parkervcp You should change the base branch to `0.7-develop` no?

parkervcp:
Eggs aren't just 0.7 I am fairly certain.

ZeProf2Code:
@parkervcp Yeah but 0.8 isn't out yet so I think it's better to first update 0.7 :D

parkervcp:
0.7 is security patch only currently.

This will just make it so 0.8 will have up to date versions of these. It also makes it easy to find them when people go looking for the current egg files as develop is the master branch.

PR #1825, opened 2020-02-08
Fix 2FA sizing issue, add support for copying text from xterm.js

### LINKED ISSUE #1812: 2FA Input, Stretchs to window width.
* Panel or Daemon: Panel
* Version of Panel/Daemon: Alpha 2
* Server's OS: Ubuntu 18.04
* Your Computer's OS & Browser: Win10 / Chrome

**Describe the bug**
2FA screen stretches to the width of the window

Why You Read this??

**To Reproduce**
Login with a 2FA account...

**Expected behavior**
The input box be the same size as the first login box

### LINKED ISSUE #1813: Unable to CTRL-C selected console text
* Panel or Daemon: Panel
* Version of Panel/Daemon: Alpha 2
* Server's OS: Ubuntu 18.04
* Your Computer's OS & Browser: Win10 / Chrome

**Describe the bug**
When highlighting text in the console, CTRL-C does nothing, so you have to right-click and copy, to copy the selected text. 

**To Reproduce**
1. Start Server
2. Select text from the console
3. CTRL-C
4. Try to paste the "copied" text

**Expected behavior**
... To Copy

matthewpi:
Fixes #1812 (The sizing issue on the 2FA page after entering a username and password)
Fixes #1813 (The inability to copy text from xterm.js on the server console page)

### DIFF
--- a/resources/scripts/components/server/Console.tsx
+++ b/resources/scripts/components/server/Console.tsx
@@ -81,6 +81,17 @@ export default () => {
             // @see https://github.com/xtermjs/xterm.js/issues/2265
             // @see https://github.com/xtermjs/xterm.js/issues/2230
             TerminalFit.fit(terminal);
+
+            // Add support for copying terminal text.
+            terminal.attachCustomKeyEventHandler((e: KeyboardEvent) => {
+                // Ctrl + C
+                if (e.ctrlKey && (e.key === 'c')) {
+                    document.execCommand('copy');
+                    return false;
+                }
+
+                return true;
+            });
         }
     }, [ terminal, connected, terminalElement ]);
 
--- a/resources/scripts/components/auth/LoginCheckpointContainer.tsx
+++ b/resources/scripts/components/auth/LoginCheckpointContainer.tsx
@@ -8,6 +8,27 @@ import { StaticContext } from 'react-router';
 import FlashMessageRender from '@/components/FlashMessageRender';
 import { ApplicationStore } from '@/state';
 import Spinner from '@/components/elements/Spinner';
+import styled from 'styled-components';
+import { breakpoint } from 'styled-components-breakpoint';
+
+const Container = styled.div`
+    ${breakpoint('sm')`
+        ${tw`w-4/5 mx-auto`}
+    `};
+
+    ${breakpoint('md')`
+        ${tw`p-10`}
+    `};
+
+    ${breakpoint('lg')`
+        ${tw`w-3/5`}
+    `};
+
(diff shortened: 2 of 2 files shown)


resources/scripts/components/server/Console.tsx
```diff
@@ -81,6 +81,17 @@ export default () => {
             // @see https://github.com/xtermjs/xterm.js/issues/2265
             // @see https://github.com/xtermjs/xterm.js/issues/2230
             TerminalFit.fit(terminal);
+
+            // Add support for copying terminal text.
+            terminal.attachCustomKeyEventHandler((e: KeyboardEvent) => {
+                // Ctrl + C
+                if (e.ctrlKey && (e.key === 'c')) {
+                    document.execCommand('copy');
+                    return false;
```
schrej: isn't just the `return false` sufficient?

schrej:
Also commit messages where you actually say what you did and just reference the commits in the body are much more useful. Same applies for PRs.

resources/scripts/components/server/Console.tsx
matthewpi: From what I understand, the `return false` prevents xterm.js from processing the event, while `return true` allows all additional events to be handled.  So while the extra return might not be necessary because xterm.js does not do anything observable with `Ctrl+C`, it should probably be left there.

https://xtermjs.org/docs/api/terminal/classes/terminal/#attachcustomkeyeventhandler

schrej: I'd assume if you prevent xterm from processing it by returning false, it shouldn't consume the event. It therefore propagates until the browser itself consumes it and copies the text. Just listening for keyboard events doesn't disable all of them, does it?

schrej:
Also, for the future, please separate fixes into individual pull requests. This way, in case one of them is unwanted or not ready yet, they can be merged individually.

matthewpi:
I could separate the commits into two pull requests if you don’t want to merge both.

DaneEveritt:
@schrej I told him he could just do a single PR for this since we're just trying to knock out bugs.

schrej:
Ah ok. I also meant it in a general sense, no need to split them up now. It's just a little cleaner that way, and most other projects would also prefer it like that I assume.

DaneEveritt:
> Ah ok. I also meant it in a general sense, no need to split them up now. It's just a little cleaner that way, and most other projects would also prefer it like that I assume.

Ah, yes, that is generally how I prefer them. :)

PR #1876, opened 2020-03-27
0.7 japanese language

PJZ9n:
Hi!
I created a Japanese(日本語, にほんご) language file for Pterodactyl 0.7.

It works as follows:
screenshot_20200328_003437


DaneEveritt:
Thank you for your contributions, however at this time 0.7 is in security-fix only mode, and therefore we are not accepting additional contributions to its codebase.

PR #1882, opened 2020-03-29
Add CPU Thread assignments

Sir3lit:
Added CPU Thread assignments for each server


Sir3lit:
Should be all set, ToDo is the rule check for the string.

PR #1907, opened 2020-04-10
Fix server-description colspan

Spirit55555:
Make sure the server-description row is as long as the server row above.

### DIFF
--- a/resources/themes/pterodactyl/base/index.blade.php
+++ b/resources/themes/pterodactyl/base/index.blade.php
@@ -78,7 +78,7 @@
                             </tr>
                             @if (! empty($server->description))
                                 <tr class="server-description">
-                                    <td colspan="7"><p class="text-muted small no-margin">{{ str_limit($server->description, 400) }}</p></td>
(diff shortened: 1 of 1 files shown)


resources/themes/pterodactyl/base/index.blade.php
```diff
@@ -78,7 +78,7 @@
                             </tr>
                             @if (! empty($server->description))
                                 <tr class="server-description">
-                                    <td colspan="7"><p class="text-muted small no-margin">{{ str_limit($server->description, 400) }}</p></td>
+                                    <td colspan="8"><p class="text-muted small no-margin">{{ str_limit($server->description, 400) }}</p></td>
```
lancepioch, line 81:
```suggestion
                                    <td colspan="9"><p class="text-muted small no-margin">{{ str_limit($server->description, 400) }}</p></td>
```

There are 9 columns above.

Spirit55555: The first column has a rowspan of 2, so that should not be counted.

PR #1911, opened 2020-04-12
Fix encryption issue when creating new nodes

matthewpi:
Fixes the encryption issue when creating a new node (encrypt the token, not the token_id) 🤦‍♂️ 

https://bin.ptdl.co/4cxz5/


DaneEveritt:
Whoops, I fixed that somewhere else too. The copy/paste game was strong.

PR #1914, opened 2020-04-12
add trailing / for folder

parkervcp:
adds a trailing `/` on the daemon base folder /srv/daemon-data/.

This is a fix for wings for another PR to the wings repo.


DaneEveritt:
Whats the wings PR this is about? I'd rather just handle the slash in wings alone if possible and not have to deal with validating the slash here.

parkervcp:
Yeah it looks like handling the / in wings is probably the best answer.

PR #1915, opened 2020-04-12
Add configurable server mounts

### LINKED ISSUE #552: Host shared folders on servers
Hi, I have 8 Garry's Mod servers, and I'm thinking to switch to Pterodactyl to manage my servers, this panel is amazing! Excellent work!. But I need one last thing, shared folders, typically in Garry's Mod you can mount other source games content with the file `garrysmod/cfg/mount.cfg`

```
"mountcfg"
{
        "cstrike"       "/home/container/SHARED_SRCDS_CONTENT/cstrike"
        "tf"            "/home/container/SHARED_SRCDS_CONTENT/tf"
... etc
```

For all contents i have a 27G folder, which solution i have to allow server's docker's to access that folder without copying it every time ?

Thank you !

matthewpi:
This adds the ability for admins to create server mounts for external directories.  This will allow specific servers to have access to shared directories or directories stored outside of where the data directory is located.

Closes #552


DaneEveritt:
Do we want to make it so a mount is assigned to a node specifically, and then servers can be assigned those mounts if they exist on that node?

This would make more sense to me, as I don't think doing arbitrary mounts for a server is necessarily the best idea. This also means the daemon can get a list of all these mounts when it boots, and do the necessary steps to ensure the paths exist to avoid containers exploding if they're not there. It also makes the API easier to use, since you can just pass along the specific mounts to include.

I'd also recommend tagging the mounts with a UUID internally since I get the feeling we'll end up exposing this to users at _some_ point and I'd like there to be a UUID in those cases.

This makes the end result being a `node_mounts` table, and having a `server_node_mount_pivot` table (or however Laravel names them). We can then get a little more flexible and make the node_mounts have a default mount type (read, read/write) and allow individual servers to have a different mount type if needed.

matthewpi:
I see the point in having mounts assignable to individual nodes as the files would have to exist on the nodes but this would make it really difficult for allowing easy integration of this with other features as every individual node could have different mounts.

Would there be a point to have mounts able to be created on eggs? (The mounts should not be exported with the egg unless an option is checked and should not get overwritten with a panel update)

Having them on eggs would allow the egg to determine what mounts a server can assign, for example adding mounts to a Garry's Mod egg for CS:S or whatever else.

But this might be a better thing to integrate into a generic mod manager as it should most likely be a user assignable option to enable/disable specific mounts depending on their needs.

Or even better, having an option to have mounts assignable to multiple eggs and to nodes, allowing the display of mounts based off of the server's egg and assigned node rather users can enable them, they are required, or require admin permissions to add them.

* These are just a bunch of thoughts I had and some better ideas came along while I was writing this, if it looks like I was all over the place it is because I was 🙂

matthewpi:
Once I fix my development environment I'm going to try implementing this in the way I described.
> Having an option to have mounts assignable to multiple eggs and to nodes, allowing the display of mounts based off of the server's egg and assigned node so users can enable them if they are required, or require admin permissions to add them.

The `Mount` model would look something like
```json
{
    "uuid": "<randomly generated uuidv4, primary key>",
    "name": "<admin choosen name for the mount, unique>",
    "source": "<source folder for the mounted content, unique>",
    "target": "<where the source will be accessible in the container, unique>",
    "read_only": "<boolean if the mount is read only>",
    "user_assignable": "<boolean if the mount can be added by users>"
}
```

With two [many to many](https://laravel.com/docs/6.x/eloquent-relationships#many-to-many) relationships, one named `MountNodes`
```json
{
    "mount_id": "<mount.id>",
    "node_id": "<node.id>"
}
```

The other named `MountEggs`
```json
{
    "mount_id": "<mount.id>",
    "egg_id": "<egg.id>"
}
```

mattmalec:
If I understand this correctly, this is similar to symlinks, but with more control? For example, you’re only able to mount whitelisted folders (added in a "Mounts" section on the Node) and these folders can be attached to an Egg?

matthewpi:
> If I understand this correctly, this is similar to symlinks, but with more control? For example, you’re only able to mount whitelisted folders (added in a "Mounts" section on the Node) and these folders can be attached to an Egg?

So you would create a mount in the admin panel (like you would a database host), you have the option to link it to multiple nodes and multiple eggs.  Then depending on if `user_assignable` (might rename to `user_mountable`) is true or false it can be attached by users who have a server on one of those nodes with one of those eggs or by an admin (if it meets the same requirements).

This uses the docker mounts feature already used to mount `/home/container` in-case you were wondering.

The reasoning for this is so only one mount object needs to be created for multiple nodes or eggs, instead of 20.  Having them link to a node is important because each node would have to have the mount source on them, it will not be fetched from the panel or anything automatically (which would be pretty cool).  Linking to multiple eggs is just for the freedom to have multiple eggs with the same mount.

Justinborzi:
This feature would be a beautiful addition in my opinion, I’m guessing this would also solve the “gmod css” issue a lot of people want/complain about(if I’m correct) but would love to see this implmented

matthewpi:
Cool story bro, force-pushing the branch closed the pull request 🙃

DaneEveritt:
You uh, broke it.

matthewpi:
This is basically ready, the only other thing to be added would be the ability for user's to attach/detach mounts that are `user_mountable` (also adding subuser permissions for this).  But it is functional in it's current state and admins are able to attach/detach mounts.

f0rmatme:
Very useful, wishing this was merged aaaa

jewome62:
I confirm, have external mountpoint is useful.
In my case, i have two real case :

- First, share the schematics into Minecraft Creative Server between many servers.
- Secondly, Just write file somewhere else, As Dynmap Standalone file for a web server.

For these servers, i use currently Multicraft because i can make it with Pterodactly
For the first, currently this is a symlink to another directory for the two servers to another directory
For the second, this is just a config into dynmap + setfactl the destination directory.

I think, for the first, a docker named volume is enough.
For the second, this is a docker volume with host mount point

PR #1917, opened 2020-04-12
add docker configs for 0.8

parkervcp:
This is a push for dockerfiles for pterodactyl 0.8


DaneEveritt:
As discussed in Discord, let's move it all under the `docker` folder, and put the `Dockerfile` in the root directory. :)

Casper042:
Apologies for the Newb questions.

1) Does this mean the next version (0.8) will be able to easily run the Ptero back end on a container itself so it doesn't have to be installed direct on a host/vm?

2) Will Ptero running as a Container be able to use the same Container host it's running on to host Games?  Will it just use a pointer to the main IP or the API access like Portainer for example.

3) You have 2 references to the Compose file that need updating here: https://github.com/pterodactyl/panel/blob/develop/docker/README.md
Currently they are all 404 because they still have /docker/ in the path.

Thanks

parkervcp:
@Casper042 you commented on a PR that has been merged for 3+ months.

PR #1924, opened 2020-04-13
Update Admin Overview for server

Sir3lit:
Changes CPU Threads show Not Set instead of n/a
Changes 0% -> Unlimited, when CPU is set to 0(unlimited)
Changes 0MB -> Unlimited, when memory is set to 0(unlimited)
Changes 0MB Swap -> Not Set
Changes -1 Swap -> Unlimited
Changes Name -> Server Name
Changes Service -> Current Egg

Mostly aesthetic / uniformity changes 
Before
Before

After: 
After

### DIFF
--- a/resources/views/admin/servers/view/index.blade.php
+++ b/resources/views/admin/servers/view/index.blade.php
@@ -47,19 +47,25 @@
                                 <td><code>{{ $server->uuid }}</code></td>
                             </tr>
                             <tr>
-                                <td>Service</td>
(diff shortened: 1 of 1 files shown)


Sir3lit:
Couldn't decide between `Current Egg` or just `Egg` or anything diff, advise if you'd like changed

parkervcp:
Current Egg is probably the best option.

Not like you would ever list past eggs.

PR #1926, opened 2020-04-14
Fix some pages

synocx:


DiscoverSquishy:
This actually seems useful in adding more info when something goes wrong with the panel. I guess it'd help people immediately spamming you asking what had happened 👍

DaneEveritt:
I'd rather we just remove the translation strings entirely from these pages and use hard-coded strings. I have no interest in maintaining languages for the admin area, and the frontend will get fixed to support them eventually.

DaneEveritt:
Closing in favor of the above comment.

PR #1937, opened 2020-04-17
Added german language files to pterodactyl 1.0.0 beta.

vanlueckn:
Added german language files.


DaneEveritt:
Thanks, however I regret supporting languages to that degree in the admin area, and won't be continuing with them in 1.0.

PR #1963, opened 2020-04-22
Add Max Concurrent Connections for database users

### LINKED ISSUE #1849: Allow max connections to be set for user databases
Add the ability to limit database connections for user databases.  This can prevent a user with a bad plugin from consuming all of the available database connections for a database host.

At a quick glance I expect this would be done at:
https://github.com/pterodactyl/panel/blob/7543ef085d1b466536e37efc6c4d679fb82797ae/app/Repositories/Eloquent/DatabaseRepository.php#L153-L161

This could be updated to specify max_user_connections if configured in the database host settings.

Sir3lit:
Closes #1849

Allows database users to be limited to a number of concurrent connections to prevent one user from connecting hundreds of times and bottlenecking the MySQL server.

### DIFF
--- a/app/Models/Database.php
+++ b/app/Models/Database.php
@@ -41,6 +41,7 @@ class Database extends Model
     protected $casts = [
         'server_id' => 'integer',
         'database_host_id' => 'integer',
+        'max_connections' => 'integer',
     ];
 
     /**
@@ -51,7 +52,7 @@ class Database extends Model
         'database_host_id' => 'required|exists:database_hosts,id',
         'database' => 'required|string|alpha_dash|between:3,100',
         'username' => 'string|alpha_dash|between:3,100',
-        'max_connections' => 'nullable',
+        'max_connections' => 'nullable|integer',
         'remote' => 'required|string|regex:/^[0-9%.]{1,15}$/',
         'password' => 'string',
     ];
--- a/app/Models/Database.php
+++ b/app/Models/Database.php
@@ -51,7 +51,7 @@ class Database extends Model
         'database_host_id' => 'required|exists:database_hosts,id',
         'database' => 'required|string|alpha_dash|between:3,100',
         'username' => 'string|alpha_dash|between:3,100',
-        'max_connections' => 'string',
+        'max_connections' => 'nullable',
         'remote' => 'required|string|regex:/^[0-9%.]{1,15}$/',
         'password' => 'string',
     ];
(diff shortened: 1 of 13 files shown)


Sir3lit:
Not sure what to make default, it's set at 150. Should it be 0? Or more of a limit of 15 or so?

DaneEveritt:
I think a default value of null makes more sense for "unlimited". Then in the UI they can pass blank or 0 to do that, and we go from there.

Sir3lit:
So allow null and make it 0 by default in the database? As iirc it needs a value or else the SQL statement will fail

DaneEveritt:
I mean make it null, and store null not 0 in the database for unlimited. I've discovered using 0 for unlimited makes it very confusing to keep track of everywhere.

For the statement, check if it is not empty (`!empty()`) to determine if you should set the limit.

Sir3lit:
```php
    public function createUser(string $username, string $remote, string $password, $max_connections): bool
    {
        if (!$max_connections) {
            return $this->run(sprintf('CREATE USER `%s`@`%s` IDENTIFIED BY \'%s\'', $username, $remote, $password));
        } else {
            return $this->run(sprintf('CREATE USER `%s`@`%s` IDENTIFIED BY \'%s\' WITH MAX_USER_CONNECTIONS %s', $username, $remote, $password, $max_connections));
        }
    }
```

Somthing like this in mind?

Sir3lit:
I made a commit hours ago and it never added updated?

https://github.com/Sir3lit/panel/commit/82dd7dc8e3b913839685728fb41f34139ee1d871

app/Models/Database.php
```diff
@@ -51,6 +51,7 @@ class Database extends Model
         'database_host_id' => 'required|exists:database_hosts,id',
         'database' => 'required|string|alpha_dash|between:3,100',
         'username' => 'string|alpha_dash|between:3,100',
+        'max_connections' => 'string',
```
matthewpi, line 54: Is there a reason why you are using `string` instead of `sometimes|integer` or `nullable|integer` as the validation rule?

Sir3lit: As referred to my last comment, I made a commit hours ago and it was changed to `nullable`

https://github.com/Sir3lit/panel/blob/82dd7dc8e3b913839685728fb41f34139ee1d871/app/Models/Database.php#L54

DaneEveritt:
I don't think showing max connections to the user is really important.

Sir3lit:
Removed from end-user, but kept it in the API call.

PR #1971, opened 2020-04-25
Add support for Google Analytics 

### LINKED ISSUE #1331: Google Analytics Input
Well, after every Pterodactyl update, I have to go about editing templates to install Google Analytics to get some statistics on panel usage, etc.

I think adding a Google Analytics code input box on settings area which adds an analytics script if there's an analytics tracking code put within it, and if the input is blank, the script is not rendered on the page to speed loading time.

Editing a template after updates is doable but I think this small feature could be very useful to quite a few people without affecting any core functionality of the panel.

**Additional context**
Here's a screenshot of what this could look like to configure. https://file.properties/host/fZZPtdKH.png

Sir3lit:
Adds to settings page to be configurable 
Adds to siteSettings output for the front end to be used by react

Can be used with `react-ga` on the front end 

Closes #1331


resources/scripts/components/dashboard/AccountOverviewContainer.tsx
```diff
@@ -1,4 +1,6 @@
 import * as React from 'react';
+import { useEffect } from 'react';
```
Sir3lit, line 2: @DaneEveritt Not sure if this is right, still trying to learn what I'm exactly doing in react.

Sir3lit:
In its current state

Sir3lit:
85% sure the migration isn't needed, as it is just making a row in the settings table... Could be removed

resources/scripts/components/dashboard/AccountOverviewContainer.tsx
DaneEveritt: You can actually just do `import React, { useEffect } from 'react';` which will do the same thing. The `import * as React from 'react';` was from before I made some changes in typescript that remove the need for that `*` syntax.

You could also use `React.useEffect()` in this case, but I don't do that anywhere, so my first example in this response is the best answer.

DaneEveritt:
You should be able to do a `useEffect()` in a root component for the router stuff. And then just do something like...

```
useEffect(() => {
    ReactGA.whatever(location.pagename);
}, [ location.pagename ]);
```

This would avoid a ton of code duplication and issues when it is inevitably forgotten in templates.

resources/scripts/components/dashboard/AccountOverviewContainer.tsx
Sir3lit: Changed app.tsx to use `import React, { useEffect } from 'react';`

Sir3lit:
It works but complains that the module isn't loaded? but still, reports to Google Analytics? Not sure what's going on as my debugging skills are not up to par with react.

resources/scripts/routers/DashboardRouter.tsx
```diff
@@ -9,6 +10,10 @@ import AccountApiContainer from '@/components/dashboard/AccountApiContainer';
 import NotFound from '@/components/screens/NotFound';
 
 export default ({ location }: RouteComponentProps) => (
```
DaneEveritt, line 12: What the heck is this formatting. 🤔 

Should be
```
=> {

useEffect(...);

return (<div/>);
```

Sir3lit: Changed

Sir3lit:
When the next version is pushed i will re-visit this when I have some free time.

PR #1981, opened 2020-04-27
Create auth.php

stashenko:


DaneEveritt:
I am not currently accepting PRs for language changes as it is not fully implemented, and I do not know what strings I actually need at this time.

PR #1997, opened 2020-05-05
Update configuration path

synocx:


parkervcp:
this would go hand in hand with https://github.com/pterodactyl/wings/pull/30

Also you could add a description as to why this would be warranted @synocx ...

DaneEveritt:
This is not correct based on pterodactyl/wings#30, and yes, if you're not going to include a description I'm not going to take the time to look at the PR.

PR #2002, opened 2020-05-07
Fix docker entrypoint database check

features-not-bugs:
The Database port was hard coded to 3306 and not taking into consideration the DB_PORT environment variable

This was causing an endless loop when starting.

### DIFF
--- a/.dev/docker/entrypoint.sh
+++ b/.dev/docker/entrypoint.sh
@@ -49,7 +49,7 @@ fi
 
 ## check for DB up before starting the panel
 echo "Checking database status."
-until nc -z -v -w30 $DB_HOST 3306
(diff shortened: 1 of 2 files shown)


parkervcp:
My bad. When I wrote that I only considered running with the DB in docker as well...

features-not-bugs:
haha whoops!

I wasn't sure what branch / where to direct this PR, so I've slapped it against master, let me know if you want me to target a different branch.

features-not-bugs:
I just absentmindedly committed to my fork again with additional changes, completely forgot it would auto add commits to this PR.

The additional commit allows for connecting to redis over ssl.

parkervcp:
should be develop

features-not-bugs:
I'll create a new pr shortly

PR #2031, opened 2020-05-16
Show "unlimited" not 0

Sir3lit:
Shows unlimited for resources if they're set to "0"
I'm sure we could also just remove the "of {limit}" part if its unlimited? Dane Advise

I'm sure there is a 100x better way of doing this? Dane, please advise.

### DIFF
--- a/resources/scripts/components/dashboard/ServerRow.tsx
+++ b/resources/scripts/components/dashboard/ServerRow.tsx
@@ -53,6 +53,21 @@ export default ({ server, className }: { server: Server; className: string | und
         alarms.disk = server.limits.disk === 0 ? false : isAlarmState(stats.diskUsageInBytes, server.limits.disk);
     }
 
+    let disklimit;
+    let memorylimit;
+
+    if(server.limits.disk != 0) {
+        disklimit = bytesToHuman(server.limits.disk * 1000 * 1000);
+    } else {
+        disklimit = "Unlimited";
+    };
+
+    if(server.limits.memory != 0) {
+        memorylimit = bytesToHuman(server.limits.memory * 1000 * 1000);
+    } else {
+        memorylimit = "Unlimited";
+    };
+
     return (
         <Link to={`/server/${server.id}`} className={`grey-row-box cursor-pointer ${className}`}>
             <div className={'icon'}>
@@ -127,7 +142,7 @@ export default ({ server, className }: { server: Server; className: string | und
                                     {bytesToHuman(stats.memoryUsageInBytes)}
                                 </p>
                             </div>
-                            <p className={'text-xs text-neutral-600 text-center mt-1'}>of {bytesToHuman(server.limits.memory * 1000 * 1000)}</p>
+                            <p className={'text-xs text-neutral-600 text-center mt-1'}>of {memorylimit}</p>
                         </div>
                         <div className={'flex-1 ml-4'}>
                             <div className={'flex justify-center'}>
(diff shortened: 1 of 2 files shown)


alhassanv:
Is Unmetered a better term to use here than Unlimited?

Sir3lit:
It's referred to as unlimited throughout the panel.

resources/scripts/components/dashboard/ServerRow.tsx
```diff
@@ -53,6 +53,21 @@ export default ({ server, className }: { server: Server; className: string | und
...
     }
 
+    let disklimit;
+    let memorylimit;
+
+    if(server.limits.disk != 0) {
+        disklimit = bytesToHuman(server.limits.disk * 1000 * 1000);
+    } else {
+        disklimit = "Unlimited";
+    };
```
schrej, line 63: You have the same code four times, why not wrap it into a function?
Or alternatively, maybe do the following as it's a lot shorter.
```js
const disklimit = server.limits.disk != 0 ? bytesToHuman(server.limits.disk * 1000 * 1000) : "unlimited";
```

Sir3lit: Used the shorter one.

schrej:
Less lines are not always better, but in this case they are imo. But let's wait for @DaneEveritt to tell me otherwise 😄

resources/scripts/components/server/ServerConsole.tsx
```diff
@@ -121,7 +124,7 @@ export default () => {
                             className={'mr-1'}
                         />
                         &nbsp;{bytesToHuman(disk)}
-                        <span className={'text-neutral-500'}> / {server.limits.disk} MB</span>
```
schrej, line 124: any reason you removed the styling here?

Sir3lit: I originally moved it to the if statement in the first pr...... Ahh I see you..... Yeh I'll fix it when I get home.

Sir3lit: I'm going to have to play with it after work, as we don't need "MB" at the end of Unlimited, one of the other reasons I moved it to the IF statement.

Sir3lit: Fixed.

Sir3lit:
Fixed conflict

resources/scripts/components/dashboard/ServerRow.tsx
```diff
@@ -52,6 +52,8 @@ export default ({ server, className }: { server: Server; className: string | und
         alarms.memory = isAlarmState(stats.memoryUsageInBytes, server.limits.memory);
         alarms.disk = server.limits.disk === 0 ? false : isAlarmState(stats.diskUsageInBytes, server.limits.disk);
     }
+    const disklimit = server.limits.disk != 0 ? bytesToHuman(server.limits.disk * 1000 * 1000) : "Unlimited";
```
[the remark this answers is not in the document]
Sir3lit: Correct, you would have to refresh the page to see the new value.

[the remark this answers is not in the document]
schrej: Doesn't that just depend on whether `server` updates automatically?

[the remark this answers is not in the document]
DaneEveritt: @schrej Yes, but it is possible server could update automatically, especially from the websocket (since I plan on sending even more events over it). I like to build with the assumption the data could change on the fly later on, rather than have some things working and some things not.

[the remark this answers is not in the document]
schrej: And why would that be an issue here? Because of const? The whole function is executed on each render, don't forget that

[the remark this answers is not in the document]
DaneEveritt: Oh right, totally forgot server was coming from a hook here and will re-render when updated. Nvm.

PR #2042, opened 2020-05-19
Allow changing back to unrestricted CPU quota

lflare:
Found a little bug trying to set one of my testing servers back to unlimited quota. `0` does not work as unlimited as Docker appears to only interpret `-1` as unlimited.


lflare:
Note: This will not work, I thought I had the smartest fix in the book. I have not woken up yet.

PR #2043, opened 2020-05-19
Allow changing back to unlimited CPU quota

lflare:
Noticed this bug while trying to play around with CPU quotas in the panel on my testing servers. Setting the quota to anything initially works, but when setting it back to zero to remove the quota, the setting does not apply. Docker only accepts `-1` as an unlimited parameter for CPU quota.

This PR fixes this misunderstanding and allows for setting of CPU quotas back to unlimited.

### DIFF
--- a/app/Models/Server.php
+++ b/app/Models/Server.php
@@ -113,7 +113,7 @@ class Server extends Model
         'memory' => 'required|numeric|min:0',
         'swap' => 'required|numeric|min:-1',
         'io' => 'required|numeric|between:10,1000',
-        'cpu' => 'required|numeric|min:0',
+        'cpu' => 'required|numeric|min:-1',
         'threads' => 'nullable|regex:/^[0-9-,]+$/',
         'oom_disabled' => 'sometimes|boolean',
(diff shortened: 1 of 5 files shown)


Sir3lit:
Its been "0" since 0.7? Was this something that docker changed recently? 

https://docs.docker.com/config/containers/resource_constraints/#cpu

Docker only states that "-1" will allow for unlimited memory and has no documentation of using "-1" for CPU limit

It it changed to -1 in wings if its set to 0 

https://github.com/pterodactyl/wings/blob/ffd7357a1cb9df050f487dc541e9412fe6ea2672/server/server.go#L108-L111

lflare:
This is weird behavior. Hang on.

The default CPU quota of a server created via normal docker commands is `0`. I might have made a huge mistake here. However when trying to reset the quota for a server, putting `0` in a panel does not seem to actually reset it.

Sir3lit:
Okay. If you'd like to come to discord we can look into it with ya when we have some time. 

~@AreYouScared

PR #2063, opened 2020-05-24
Show Unlimited instead of 0MB for disk space on about page ( Admin Side )

Sir3lit:
I guess I missed this one when I set everything to show unlimited when its 0.

Shows Unlimited disk space instead of 0 on admin about page.

### DIFF
--- a/resources/views/admin/servers/view/index.blade.php
+++ b/resources/views/admin/servers/view/index.blade.php
@@ -97,7 +97,13 @@
                             </tr>
                             <tr>
                                 <td>Disk Space</td>
-                                <td><code>{{ $server->disk }}MB</code></td>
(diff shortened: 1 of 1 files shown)


schrej:
Is it 0 or -1?

fabm3n:
If i try set it to -1 the panel throws a 500 (Should not happen will open issue soon).
If i set the disk space to 0 the server can't start, because the disk is full.

So it looks like there isn't any unlimited disk space available right?

Sir3lit:
@schrej some things like memory are -1, Disk space is 0 for unlimited. 

As for the -1, 500 error, it's just failing to pass the validation error to the panel, and it just throws a 500 instead.

PR #2091, opened 2020-06-05
Add 'db:backup' command to artisan command line

Sir3lit:
Allows users to export the panel's database to an SQL file using `mysqldump`, uses the config values in the .env, This is reliant on MySQL / MariaDB being installed on the machine the panel is running on.

If there is a better way of doing this, please do show.

### DIFF
--- a/app/Console/Commands/BackupDatabase.php
+++ b/app/Console/Commands/BackupDatabase.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Pterodactyl\Console\Commands;
+
+use Illuminate\Console\Command;
+use Symfony\Component\Process\Process;
+use Illuminate\Contracts\Config\Repository as ConfigRepository;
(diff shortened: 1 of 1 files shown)


app/Console/Commands/BackupDatabase.php
```diff
@@ -0,0 +1,48 @@
...
+class BackupDatabase extends Command
+{
+    protected $signature = 'db:backup';
+
+    protected $description = 'Provides an exported sql file of the panels current database.';
+
+    /**
+     * @param \Illuminate\Contracts\Config\Repository $config
+     */
+    public function __construct(ConfigRepository $config)
```
DaneEveritt, line 19: This should be done in the `handle()` function. This constructor is call whenever a command is executed, so all of this code will be executed (even though technically yes, it does nothing).

Makes more sense to just do it all in the handle. Also, I've moved away from the verbose way of getting configuration values, you're welcome to just use `config()` as needed.

PR #2105, opened 2020-06-10
Fix improper byte conversions

matthewpi:
Use 1024 instead of 1000 for byte conversions (which were added by #2071), rename gloabl.d.ts to global.d.ts.


DaneEveritt:
@matthewpi got some conflicts.

PR #2118, opened 2020-06-14
Add scrollbar styling + change syntax background color

Sir3lit:
Adds styling to the scrollbar using the WebKit in the CSS, WebKit is widely supported by most modern browsers. 

Anyone have a better idea of colors / looks for it please let me know, I was at the point of anything is better then what it currently is. 

Before(already has the syntax background change) :
image
image

After:
image
image

The last one shows the mouse hover color.

I'm open to changes... wasn't sure if it should be a color for the hover over or not?


Sir3lit:
For some reason...
image

that refuses to use the scrollbar style.... everything else on the front end uses it... BUT THAT

Lunaversitay:
> 
> 
> For some reason...
> image
> 
> that refuses to use the scrollbar style.... everything else on the front end uses it... BUT THAT

https://gist.github.com/maxcraftsman/f10bf7a47f3dd02ce8c7844937dc3a6a

Sir3lit:
Ohhh

Sir3lit:
@Lunaversitay No matter what I do... I can't get that one to change...

PR #2125, opened 2020-06-17
Show correct action name and icon in schedule tast view

Sir3lit:
When backups were added as a task for schedules the name and icon were not added to properly support it. Causing it to show up as a power action and have "send command" as the name.


Sir3lit:
Closing as it was already fixed in https://github.com/pterodactyl/panel/commit/693b9eab0c5913cde6908697455cdbb3e7bc39ae

PR #2128, opened 2020-06-17
Fix #terminal css line formatting

galeapatrik:
Add missing [white-space: pre-wrap] to [#terminal > .cmd]

Before
before

After
after

### DIFF
--- a/public/themes/pterodactyl/css/terminal.css
+++ b/public/themes/pterodactyl/css/terminal.css
@@ -27,6 +27,7 @@
 #terminal > .cmd {
     padding: 1px 0;
     word-wrap: break-word;
+    white-space: pre-wrap;
 }
 
 #terminal_input {


parkervcp:
The other way was formatted for a reason. Cutting all the whitespace will break formatting and will make it hard to read in several applications.

DaneEveritt:
@parkervcp do you have an example of ourput that is made difficult by the change?

parkervcp:
I swear the screen shots were the other way around when I commented on this.

I agree though as it does look like it's broken on the default config.

parkervcp:
This does fix some formatting I have that has been broken as well.

galeapatrik:
> I swear the screen shots were the other way around when I commented on this.
Yep, I edited it right after posting, but you were too fast. :D

PR #2179, opened 2020-07-16
Add fastdl option for csgo,and source based games.

delije23:
Games like 1.6 and css use SRCDS that only allows up to 20kbps.Imagine someone downloading a 1mb map.The fastdl allows the client to download custom server content (maps, materials, models, particles, sounds, fonts, images) from a web server. It is used to offload content downloads to a web server instead of relying on source servers.


matthewpi:
Not sure why you opened a pull request for something that is a WIP and you didn't write.  Also this feature doesn't do what you think it does.

DaneEveritt:
This has been brought up numerous times and the answer remains no. I don't have the time or energy to be building out tons of he specific functionality right now and it goes against my long standing approach to this panel, which is keeping it take agnostic.

Additionally, pull requests are for code changes, not making a feature request (which also can be searched here and discord to see the hundreds of times we've said no). 👍🏼

PR #2301, opened 2020-09-01
Update loading status of file upload via d&d (progress bar)

msharp19:
If you drag and drop large files via the file manager, you are only presented with a loading cursor. This is fine for small file transfer but when it comes to large file transfer (which takes a varying amount of time) it would be nice to see the upload progress. If you ask me it would be good to see the progress bar regardless of the file size but that's up to you!

Sorry if this is not the correct pace to log requests but I have found myself here because of - https://steamuserimages-a.akamaihd.net/ugc/1595918438106867352/02736C976F64F6B91F35DEC6971EBC6A6EEEE986/

Thanks!

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -9,7 +9,7 @@ return [
     | change this value if you are not maintaining your own internal versions.
     */
 
-    'version' => 'canary',
+    'version' => '1.0.0-rc.3',
 
     /*
     |--------------------------------------------------------------------------


fabm3n:
This is definitely the wrong place. You need to create a issue and select Festure Request as template

msharp19:
Thought so lol thanks!

PR #2327, opened 2020-09-12
fix: remove ./ before wings for auto-configuration

ZeProf2Code:
on the docs, you're downloading it to /usr/local/bin so it's easier like this


DaneEveritt:
@Sir3lit `/etc/pterodactyl` is correct since that is where we store the configuration stuff for wings. The binary itself is just in `/usr/local/bin`. 👍

PR #2339, opened 2020-09-14
Use YAML for eggs

### LINKED ISSUE #2193: Use YAML for eggs
Currently, eggs are exported in json format. Eggs can contain a lot of multiline strings, which makes it very hard to read them in json format.

YAML has much better support for multiline strings (see [here](https://yaml-multiline.info)) which would make it much easier to modify eggs as code and see changes in Pull Requests.
In addition YAML is easier to read and we've converted wings to use yaml as it's configuration file format as well.

As YAML is basically an extension of JSON, it should be easy to be backwards compatible.

occanowey:
Eggs are now exported and imported as YAML and use YAML for their configs.
This also has a fallback for importing JSON eggs and updating them to PTDL_v2.

Closes #2193


database/seeds/eggs/minecraft/egg-bungeecord.yml
```diff
@@ -0,0 +1,67 @@
...
+meta:
+  version: PTDL_v1
+exported_at: '2020-04-12T16:00:51-07:00'
+name: Bungeecord
+author: support@pterodactyl.io
+description: 'For a long time, Minecraft server owners have had a dream that encompasses a free, easy, and reliable way to connect multiple Minecraft servers together. BungeeCord is the answer to said dream. Whether you are a small server wishing to string multiple game-modes together, or the owner of the ShotBow Network, BungeeCord is the ideal solution for you. With the help of BungeeCord, you will be able to unlock your community''s full potential.'
+image: 'quay.io/pterodactyl/core:java'
+startup: 'java -Xms128M -Xmx{{SERVER_MEMORY}}M -jar {{SERVER_JARFILE}}'
+config:
+  files: |
```
schrej: Have you verified whether the daemon can still handle that? Or does that get converted back to JSON?
If we do that as YAML from now on I also see no reason to have it as nested plain text. That would require to adapt the daemon though.

app/Services/Eggs/Sharing/EggExporterService.php
```diff
@@ -42,36 +54,40 @@ public function handle(int $egg): string
         $egg = $this->repository->getWithExportAttributes($egg);
 
         $struct = [
-            '_comment' => 'DO NOT EDIT: FILE GENERATED AUTOMATICALLY BY PTERODACTYL PANEL - PTERODACTYL.IO',
             'meta' => [
-                'version' => 'PTDL_v1',
+                'version' => 'PTDL_v2',
```
schrej, line 50: Apart from switching from JSON to YAML, this didn't really change the format, did it? Considering JSON is valid YAML, do we even have to change the version here?

The only change I can think of right now are the fields that used to json strings.

database/seeds/eggs/minecraft/egg-bungeecord.yml
parkervcp: One thing to remember is that you need to escape characters that would parse weirdly in yaml. Like these.`*&#` plus others.

occanowey: All the egg configs get transformed back into json when used in the api, but I've also verified it works with a live daemon.
I just kept the configs as strings because that's how they were in json and also they're stored as raw yaml in the database so it does make some sense to store them as plain values.

In block scalars no printable characters need to be escaped but that also comes with the downside of not being able to insert raw invisible characters.

app/Services/Eggs/Sharing/EggExporterService.php
occanowey: The json fields are the only reason I changed it. Because you can still upload old json eggs I wanted to have support for migrating the fields to yaml and decided on a version change, but we could also just transform the fields to yaml when a json egg is uploaded.

DaneEveritt:
Just wanted to mention I'll be looking at this hopefully this evening, didn't want to leave ya hanging. 👍

app/Rules/Yaml.php
```diff
@@ -0,0 +1,38 @@
+<?php
```
DaneEveritt: I think this file could probably be renamed to `IsYaml` which is more correct for what it is validating, and would certainly reduce the confusion with the Symfony YAML package.

app/Services/Eggs/Sharing/EggExporterService.php
DaneEveritt: I think v2 is fine here, it isn't used for much and might be useful down the road.

app/Services/Eggs/Sharing/EggExporterService.php
```diff
@@ -42,36 +54,40 @@ public function handle(int $egg): string
...
-                    'script' => $egg->copy_script_install,
+                    'script' => $this->crlfToLf($egg->copy_script_install),
                     'container' => $egg->copy_script_container,
                     'entrypoint' => $egg->copy_script_entry,
                 ],
             ],
             'variables' => $egg->variables->transform(function ($item) {
-                return collect($item->toArray())->except([
-                    'id', 'egg_id', 'created_at', 'updated_at',
+                $exportItem = collect($item->toArray())->except([
```
DaneEveritt: `$item->except()->merge(['description' => ''])->toArray();` should work.

database/migrations/2020_09_13_230104_replace_egg_json_with_yaml.php
```diff
@@ -0,0 +1,63 @@
...
+class ReplaceEggJsonWithYaml extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        $eggs = DB::select('SELECT id, config_files, config_startup, config_logs FROM eggs;');
```
DaneEveritt, line 16: Please import the DB facade here so we're not using the globally namespaced ones. 👍

database/seeds/eggs/minecraft/egg-bungeecord.yml
DaneEveritt: I feel like that was probably a design mistake on my part. This seems like a really weird way to export files, we can probably just export each of these sections as their own valid YAML that way its not a complete disaster to manually edit these files?

It parses out the same anyways.

DaneEveritt (changes requested):

fadedmax:
I think a few of the problems that i now see are how, somthing like this:
Old json message: "The JSON file you submitted was either blah blah blah"
Newer message: "The _file_ You put in was blah blah blah blah"
I think it would be better to specify YAML since people may not be getting used to the change.

DaneEveritt:
@occanowey Is it alright with you if I back-burner this for a little bit? Overall it is fine, but we're incredibly close to having a 1.0 final release build, and I'd like to feature freeze the codebase for a little bit so I'm not breaking things so close to launch. :)

gOOvER:
sry, when i ask, because @DaneEveritt answers this:

```
DaneEveritt commented on 17 Sep 2020
@occanowey Is it alright with you if I back-burner this for a little bit? Overall it is fine, but we're incredibly close to having a 1.0 final release build, and I'd like to feature freeze the codebase for a little bit so I'm not breaking things so close to launch. :)
```

When it will be implemented? @occanowey does great work and now it's forgotten? Can you pls start working on Egg System, because Eggs are  the "heart" of Ptero.

And there are a lot of open Tickets of broken parsing System

DaneEveritt:
There are three open issues about the egg system. Switching to a `yaml` format will fix exactly 0 of them.

[Yours](https://github.com/pterodactyl/panel/issues/2533), which is incredibly low priority and a very non-standard config format. [A unicode parsing issue](https://github.com/pterodactyl/panel/issues/3009), which will get fixed eventually but hasn't caused extreme pain at this point. [And a parsing issue](https://github.com/pterodactyl/panel/issues/2693) that is not a trivial fix most likely, and really just has more to do with comments and spacing issues, less to do with the actual underlying logic. Theres 130+ open issues about a variety of bugs and feature requests. Two uncommon parsing issues does not suddenly make the egg system broken or a high priority item for me.

gOOvER:
> There are three open issues about the egg system. Switching to a `yaml` format will fix exactly 0 of them.
> 
> [Yours](https://github.com/pterodactyl/panel/issues/2533), which is incredibly low priority and a very non-standard config format. [A unicode parsing issue](https://github.com/pterodactyl/panel/issues/3009), which will get fixed eventually but hasn't caused extreme pain at this point. [And a parsing issue](https://github.com/pterodactyl/panel/issues/2693) that is not a trivial fix most likely, and really just has more to do with comments and spacing issues, less to do with the actual underlying logic. Theres 130+ open issues about a variety of bugs and feature requests. Two uncommon parsing issues does not suddenly make the egg system broken or a high priority item for me.

Hmmm; i understand Bug, but i don't understand Feature Requests as High Prio. Are the Feature Requests paid?Then it makes sense, that they have High prio and brings you money. 

 MB Then you should tell the Author, that u do not want such an PR and not that, what you wrote him.


This should also a High Prio Bug (https://github.com/pterodactyl/panel/issues/2533). and the rest, as you said, you can fix easily

lancepioch:
@occanowey

If you would like to fix the conflicts because of the version 1 release, I can help test this week or next and confirm the changes.

Although my only thought here is that because json is valid yaml, we shouldn't really need a converter.

PR #2340, opened 2020-09-14
Add Redis UNIX socket support besides to TCP

kodehat:
Hi,

I'm currently in a scenario where I'm forced to connect to my Redis instance with an UNIX socket. As this is currently supported by Predis (as seen [here](https://github.com/predis/predis/wiki/Connection-Parameters)), it wasn't directly supported in Pterodacyl. So I've added it as configuration options during installation. Of course `TCP` stays the default to be backward compatible.

If there's something that I missed, I am looking forward to your comments. 😃

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -96,6 +96,8 @@ return [
         'client' => 'predis',
 
         'default' => [
+            'scheme' => env('REDIS_SCHEME', 'tcp'),
+            'path' => env('REDIS_PATH', '/run/redis/redis.sock'),
             'host' => env('REDIS_HOST', 'localhost'),
             'password' => env('REDIS_PASSWORD', null),
             'port' => env('REDIS_PORT', 6379),
@@ -103,6 +105,8 @@ return [
         ],
 
         'sessions' => [
+            'scheme' => env('REDIS_SCHEME', 'tcp'),
+            'path' => env('REDIS_PATH', '/run/redis/redis.sock'),
             'host' => env('REDIS_HOST', 'localhost'),
             'password' => env('REDIS_PASSWORD', null),
             'port' => env('REDIS_PORT', 6379),


DaneEveritt:
I think this would be better handled by just adding the normal configuration keys for it, and excluding it from the environment setup command. There are enough struggles with people inputting garbage values or completely incorrect information (regardless of how much guidance we give), and I'm worried this is just going to increase support volume.

Thoughts?

config/database.php
```diff
@@ -96,13 +96,17 @@
...
+            'path' => env('REDIS_PATH', '/var/run/redis/redis.sock'),
             'host' => env('REDIS_HOST', 'localhost'),
             'password' => env('REDIS_PASSWORD', null),
             'port' => env('REDIS_PORT', 6379),
             'database' => env('REDIS_DATABASE', 0),
         ],
 
         'sessions' => [
+            'scheme' => env('REDIS_SCHEME', 'tcp'),
+            'path' => env('REDIS_PATH', '/var/run/redis/redis.sock'),
```
matthewpi, line 109: This should probably be changed to `/run/redis/redis.sock` as most modern OSes symlink `/var/run` to `/run`.

config/database.php
```diff
@@ -96,13 +96,17 @@
         'client' => 'predis',
 
         'default' => [
+            'scheme' => env('REDIS_SCHEME', 'tcp'),
+            'path' => env('REDIS_PATH', '/var/run/redis/redis.sock'),
```
matthewpi, line 100: This should probably be changed to `/run/redis/redis.sock` as most modern OSes symlink `/var/run` to `/run`.

matthewpi (changes requested):

kodehat:
I can understand both proposals and have updated the PR accordingly.

DaneEveritt (approved):

Issue #2365, opened 2020-09-18
Mobile responsiveness is broken between lg and md sized screen sizes.


gabefraser:
DO NOT REPORT ISSUES CONFIGURING: SSL, PHP, APACHE, NGINX, YOUR MACHINE, SSH, SFTP, ETC. ON THIS GITHUB TRACKER.

For assistance installating this software, as well as debugging issues with dependencies, please use our discord server: https://discord.gg/pterodactyl

You MUST complete all of the below information when reporting a bug, failure to do so will result in closure of your issue. PLEASE stop spamming our tracker with "bugs" that are not related to this project.

**STOP: READ FIRST, AND THEN DELETE THE ABOVE LINES**

**Background (please complete the following information):**
* Panel or Daemon: Panel
* Version of Panel/Daemon:  1.0.0-rc6
* Server's OS: Ubuntu 18.04
* Your Computer's OS & Browser: Chrome 85.0.4183.102

**Describe the bug**
- Text isn't being shrunk to the proper screen size / it's causing an overflow with large titles.
- Text is overflowing on the console.
- Breakpoints aren't wrapping correctly.

![image](https://user-images.githubusercontent.com/10922294/93587186-d76c0480-f9ec-11ea-84fa-c5638cf71d53.png)
![image](https://user-images.githubusercontent.com/10922294/93587199-ddfa7c00-f9ec-11ea-88fd-d52043c856f4.png)
![image](https://user-images.githubusercontent.com/10922294/93587209-e357c680-f9ec-11ea-962b-8b278fe31516.png)

**To Reproduce**
Shrink the panel to the certain sizes.

**Expected behavior**
Proper flow of mobile responsiveness.

zKoz210:
![image](https://user-images.githubusercontent.com/28255085/93708375-50c73c80-fb3e-11ea-81e1-d1e77c81e6ed.png)
![image](https://user-images.githubusercontent.com/28255085/93708379-5755b400-fb3e-11ea-8e70-62d085b5444a.png)
![image](https://user-images.githubusercontent.com/28255085/93708410-810edb00-fb3e-11ea-8996-1e90775f89d6.png)

LeCodeCo:
So me and @AreYouRlyScared were testing it out and we located the issue. It seems to only exist on Chrome and it's caused by the graphs when the server is running. The moment the server stops and the graphs disappear, everything works again.

We need to identify why the graphs stay really big.

AreYouRlyScared:
[...]

PR #2411, opened 2020-09-23
Partial fix for mobile responsiveness

zKoz210:
Fixed console size on lg
Fixed the power button panel on lg
Fixed graphs that were huge in size

Partial fixed apply to #2365. Need fix console text overflow (I have no idea how to do this)

image


zKoz210:
@DaneEveritt can review?

PR #2434, opened 2020-09-29
Added Autoallocation Button

alliraine:
This will add a new allocation to a server at the user's request. It uses the already existing setting for how many allocations a user can have. It also allows you to enable a feature that will auto-create a new allocation if there isn't one available. If this enabled, it will pick a random port from the given range if there are more ports available in that range if there is no more allocation available.


DaneEveritt:
As mentioned in Discord, just waiting on 1.0 final to release before looking into these new features. 👍

matthewpi:
I'm not sure exactly what you did but it seems you merged from the `v1.0.1` tag (or `release/v1.0.1` branch) instead of `develop` which caused the version information in the code to be updated, this should be reverted as the version should only be set on the release branches themselves and not `develop`.

alliraine:
Yup yup. So I did. I'll get it sorted tomorrow. *facepalms*

matthewpi (dismissed):
Other than the v1.0.1 version information being merged and weird code-styling (not sure why StyleCI isn't running), this looks RTM (ready to merge).

PR #2454, opened 2020-10-05
Fixed phantom shadow when opening modal window

zKoz210:
1. Fixed phantom shadow when opening modal window
2. Fixed a bug with an incorrect margin modal
2. Fixed overflow API keys description
3. Fixed incorrect button appearance

image
image

image
image

image
image

image

image


DaneEveritt:
Please use `break-words` and not `break-all` when handling breaks. Using `break-all` makes things impossible to read if a sentence spans multiple lines as it will break in the middle of a word rather than just wrapping the entire word to the next line.

PR #2466, opened 2020-10-06
Rename CPU Threads for CPU Pinning

### LINKED ISSUE #2462: Rename CPU Threads to CPU Pinning
The name "CPU Threads" has lead to confusion twice today, and it actually is confusing. It sounds like it's the amount of CPU threads the server can use, not which specific (multi threaded) cores the server will use.

I suggest renaming it to "CPU Pinning", as that's more correct, and less likely to fall victim to "oh, I don't really know what I'm doing but my server needs all the performance it can get". While we're at it, we should probably make the "advanced" warning more prominent.

LeJeanbono:
Fix #2462

### DIFF
--- a/resources/views/admin/servers/new.blade.php
+++ b/resources/views/admin/servers/new.blade.php
@@ -158,7 +158,7 @@
                     </div>
 
                     <div class="form-group col-xs-6">
-                        <label for="pThreads">CPU Threads</label>
+                        <label for="pThreads">CPU Pinning</label>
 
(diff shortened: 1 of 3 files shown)


schrej (approved):

matthewpi:
Do we want to rename the actual variables or columns as well?  I know that can become a problematic change but it might prevent some confusion in the future.

fadedmax (approved):

PR #2471, opened 2020-10-08
Display empty lines (linebreaks) in terminal.

Patrick2562:
Linebreaks were not visible in live terminal, unless at page-refresh in loaded log.
- Also need to remove whitespace replace in daemon, here is the PR: https://github.com/pterodactyl/daemon/pull/118.

### Images
Before
Before image
After
After image

### Example code in NodeJS
```js
setInterval(console.log, 1000, "something\n");
```

### DIFF
--- a/public/themes/pterodactyl/css/terminal.css
+++ b/public/themes/pterodactyl/css/terminal.css
@@ -25,6 +25,7 @@
 }
 
 #terminal > .cmd {
+    height: 14px;
     padding: 1px 0;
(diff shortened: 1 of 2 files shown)


DaneEveritt:
Thank you for taking the time to open a PR for Pterodactyl. However, this version of the software is in security fixes only mode as we approach the release of 1.0.0. Due to this, I will not be merging in this PR at this time.

PR #2474, opened 2020-10-10
[Spelling, Minor] Disk/Mem/Swap is in MiB, not MB.

LilSizzurp:


DaneEveritt:
This is incorrect, we're multiplying by `1000` in Wings for all of this, so these suffixes are already correct.

PR #2503, opened 2020-10-12
create ua lang

Yarynka:


DaneEveritt:
Thank you but we specifically are not supporting multiple languages at this time, so this PR is fairly pointless until we have a proper frontend language system built out.

PR #2512, opened 2020-10-14
Upgrade Xterm to v4.9, Add Search

notAreYouScared:
Changes:
Added ` xterm-addon-search ` v0.7.0
Added ` xterm-addon-search-bar ` v0.2.0
Updated ` xterm `  v3.14.4 -> v4.9.0
Updated ` xterm-addon-fit `  v0.1.0 -> v0.7.0
Updated ` xterm-addon-attach ` v0.1.0 -> v0.4.0
Changed selection color to be yellowish instead of white so it would standout better, If you have another color to use please advise.

With the added packages above, when a user does Ctrl + F, a search box will appear within the console for them to search what's in the console.

This was requested in discord to allow the lines in the console to be searchable.

tUjfWQ3


resources/scripts/components/server/Console.tsx
```diff
@@ -55,6 +58,9 @@ export default () => {
     const TERMINAL_PRELUDE = '\u001b[1m\u001b[33mcontainer@pterodactyl~ \u001b[0m';
     const ref = useRef<HTMLDivElement>(null);
     const terminal = useMemo(() => new Terminal({ ...terminalProps }), []);
+    const fitAddon = new FitAddon();
+    const searchAddon = new SearchAddon();
+    const searchAddonBar = new SearchBarAddon({ searchAddon });
```
schrej: should be called `searchBarAddon`

resources/scripts/components/server/Console.tsx
```diff
@@ -82,26 +88,34 @@ export default () => {
...
+            // Add support for capturing keys
             terminal.attachCustomKeyEventHandler((e: KeyboardEvent) => {
-                // Ctrl + C
+            // Ctrl + C ( Copy )
                 if (e.ctrlKey && (e.key === 'c')) {
                     document.execCommand('copy');
                     return false;
                 }
 
+                if (e.ctrlKey && (e.key === 'f')) {
```
schrej: those parentheses are unnecessary, I'd remove them unless the linter complains otherwise.

resources/scripts/components/server/Console.tsx
```diff
@@ -82,26 +88,34 @@ export default () => {
...
+                    searchAddonBar.hidden();
+                }
                 return true;
             });
         }
     }, [ terminal, connected ]);
 
     const fit = debounce(() => {
-        TerminalFit.fit(terminal);
+        fitAddon.fit();
```
schrej, line 118: If we have this, do we need the previous call to `fit()`? Or does it lead to the console taking 100ms to adapt if the first one isn't there?

schrej (changes requested):

resources/scripts/components/server/Console.tsx
notAreYouScared: I have no idea? it was already like that I just adapted it to the new way Xterm does addons in 4.x.
It resizes correctly every time you adjust the window.

resources/scripts/components/server/Console.tsx
notAreYouScared: I was just following what was already there, if you have an improvement, please do post.

resources/scripts/components/server/Console.tsx
notAreYouScared: Fixed.

resources/scripts/components/server/Console.tsx
matthewpi: Not quite sure why I put those there in the first place (for line 99), they can both be removed.

resources/scripts/components/server/Console.tsx
matthewpi: When initially rendering xterm you should call `fit()` to ensure that the terminal is properly sized on component load, otherwise you can end up with weird results when the resize event is called.  The code for the fitAddon seems fine to me.

resources/scripts/components/server/Console.tsx
notAreYouScared: The issue was you couldn't Ctrl-C with selected text in the console, as it was sending ctrl-c to the console itself.

notAreYouScared: https://github.com/pterodactyl/panel/commit/b05048871c65141b64df0912024f1bfef4c3527b

resources/scripts/components/server/Console.tsx
notAreYouScared: This was added by Dane to resize the terminal when the panel was being made responsive.

resources/scripts/components/server/Console.tsx
schrej: I was just referring to code style, the following should be equivalent:
```suggestion
                if (e.ctrlKey && e.key === 'f') {
```

resources/scripts/components/server/Console.tsx
schrej: Nevermind, I misunderstood that piece of code, sorry 😄

resources/scripts/components/server/Console.tsx
notAreYouScared: Correction, it cannot be called `searchBarAddon` as that's also the name of the package. I've named it SearchBar

matthewpi (dismissed):

resources/scripts/components/server/Console.tsx
```diff
@@ -60,7 +60,7 @@ export default () => {
     const terminal = useMemo(() => new Terminal({ ...terminalProps }), []);
     const fitAddon = new FitAddon();
     const searchAddon = new SearchAddon();
-    const searchBarAddon = new SearchBarAddon({ searchAddon });
+    const SearchBar = new SearchBarAddon({ searchAddon });
```
schrej: Should be `searchBar` then though.

schrej (dismissed):
LGTM apart from that :D

resources/scripts/components/server/Console.tsx
notAreYouScared: Done.

matthewpi (approved):

DaneEveritt (approved):

PR #2521, opened 2020-10-14
Theme System [WIP]

HavilahOG:
config/theme.php outlines the layout/requirements of a theme

A few things I need to sort out are
- Decide between leaving the tailwind config where it is in the root vs having it per theme which would also require me to edit .babel-plugin-macrosrc.js somehow
- Theme variable for components to use so I can set the svg path without writing the theme name per theme
- Better middleware positioning or an alternative


DaneEveritt:
I have strong feeling about this, especially with the current implementation. We have had _numerous_ floods of people thinking there are bugs with the software when we roll critical security updates or making any type of UI change that hasn't been reflected properly in the theme.

I would much prefer any theme system simply be an extension of the base theme appending any changes as necessary by using hooks, and making CSS changes through some other manner.

I am much more inclined to lean towards telling people to maintain their own fork of the software if they want to make dramatic changes to the design. I think I have wasted far more time supporting theming and dealing with the inevitable constant flood of support requests from it than any other issue in this project. I also think it truly detracts from the product itself -- rather than encouraging people to open well thought out issues detailing usability issues, and how they'd like it fixed, people just run off and find a theme.

Suffice to say I am strongly leaning away from this PR, but I am going to leave it open so that a.) I can continue reflecting on this and b.) others can chime in with ideas and input.

HavilahOG:
I think the flood of people not reading update notes and having a problem with outdated external mods (Themes in this case) shouldn’t be something the project team/staff uses as a reason to remove a theme system.

 It’s really something that should actually be documented for the users/creators and from there it’s out of your hands like every other feature should be imo.

I do like the idea of hooks etc but if every “theme” is going to be limited to the base design with slight text/css changes there’s no point.

I don’t think the majority of people picked up themes because of an actual issue they had with the panel design, they just wanted something different to stand out which is common.

redstonedesigner:
As an end user and system maintainer with a significant visual difficulty, this is something that would very much help me, along with the rest of my team.

I use Pterodactyl for a lot of things, and one of my main problems with it is that I find it difficult to access given that certain things are not very high in contrast (for example, the server resource limits on the front page and in the server console view page).  Having a theme system would allow for increased customisability of the themes to the point where users can create themes that work for them.

Another example is someone with dyslexia may find it easier to use a dyslexia-friendly font, however that would likely not need an entire panel reconfiguration but another theme to be created instead.

All-in-all, this would be a great feature for users who struggle with reading and/or viewing the software.

schrej:
@redstonedesigner Regardless of whether we add a theming system it would be good to make the panel more accessible. Contrast is a very valid point that often gets overlooked in favor of making it look good.

It would be cool if you could open an issue with all the stuff you're struggling with as there are probably others that have similar issues. Maybe we can find a way to improve contrast while still maintaining the nice looks. Alternatively maybe add a high contrast toggle that improves the contrast in some areas, that would be quite a bit of work though.

I don't know how good we're doing with accessibility right now, but if there are ways (especially easy ones) to improve it, I'd be all for that.

robertdrakedennis:
In terms of companies or paid platforms using pterodactyl, which is a large majority of where pterodactyl see's it's adoption, theming is something that should be reinstated at some point. There is no real way for brands or companies to separate themselves and for them this will be a huge loss and I would not be surprised to see 1.0 not being adopted for quite some time because of it. I'm honestly surprised to see 1.0 have less first class modding support after the huge market that came with it.

DaneEveritt:
@redstonedesigner As @schrej mentioned please open an issue with the specific areas that need improvement, we should be fixing those at a first-class level not relying on a third-party to fix them and distribute those fixes separately.

As for the theme system, after some discussions in a private channel I think there might be a good middle ground that can work. Rather than having to ship _all_ of the modified JS files and CSS, what if we build a simpler theme system that allows you to ship the final product of those files (e.g. the resulting `build:production` output) which can be uploaded and managed via the Panel.

* For a majority of users this is all they need. Just the ability to change appearances and be done.
* It saves people having to install all of the build dependencies and get them working correctly on their system. Imagine trying to explain that whole process to someone that just wants a different style.
* It avoids introducing additional complexity into the codebase, and avoids having a confusing folder structure and having to handle all of the potential edge cases and issues as a result.

A fully-fledged theme system is (IMO) not a simple change like many people would like it to be. There is a lot of underlying complexity when you account for how you handle themes that are only making slight modifications to a few files versus a theme modifying all of them. Keeping everything updated between them is just as much work, and introduces even more places for something to go wrong. It also introduces additional support overhead for us, regardless of how many warnings or help bubbles we put throughout the docs, panel, and Discord.

If at the end of the day a majority of users (and I recognize this doesn't solve it for everyone) just want a different theme I think my suggestion above would accomplish that. For those looking for far more control, perhaps I can write up some documentation for how to work with the build system, change colors, and what the file structure is like.

mistermantas:
Hi, I was discussing about whether to use v1.0 or v0.7 and the main issue with using the new version is that there's no (seemingly) straightforward way to look into changing layout, CSS, or anything like that. Do you know when such a feature is likely to be made available? Even in a beta (not dev build) setting?

edbrk:
From experience working on modifying the panel these last few days - At first I was finding it extremely confusing to navigate the basic file structure of the frontend and to understand where things were, keep in mind I have very little frontend experience with React and typescript which I am sure is relatable to many people who want a theme. It becomes workable after you start looking at a few main files that link everything together but at first its extremely confusing 

I think the most useful thing for theme developers and people wanting to change small aspects of the panel would be having Offical Documentation on how the frontend file structure is laid out as well as the build process and how someone would go about modifying something. (although it kind of already exists in .mds in this repo). I have already answered about 10 questions in the Discord which are asking the same thing - "why are my changes not showing when I modify something?". People just need some sort of guidance.

DaneEveritt:
Closing this PR as after additional discussion I've settled on supporting themes via allowing the installation of pre-bundled assets and just telling the Panel to load them from a specific directory. This avoids the complexity of making people run bundler commands and overall keeps the panel from going too far down a theme rabbit hole.

FoksVHox:
> Closing this PR as after additional discussion I've settled on supporting themes via allowing the installation of pre-bundled assets and just telling the Panel to load them from a specific directory. This avoids the complexity of making people run bundler commands and overall keeps the panel from going too far down a theme rabbit hole.

Will this support multiple themes? Let's that I want a Christmas theme, and a Halloween theme, as well as a normal theme that isn't the default Pterodactyl theme?

theMackabu:
sounds cool

PR #2525, opened 2020-10-16
Remove target="blank" from /admin link

Spirit55555:
This should be changed to make it consistent with the admin panel, where the link back to the user panel does not open in a new tab.


gOOvER:
ty for this PR. Admin in a new Window was annoying

PR #2558, opened 2020-10-19
Allow uuid to be inserted on the user model

HavilahOG:
Model creation using the create method can't be done if the uuid can't be filled out. User::create();

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -92,6 +92,7 @@ class User extends Model implements
         'totp_authenticated_at',
         'gravatar',
         'root_admin',
+        'uuid',
     ];
 
     /**


DaneEveritt:
What specifically is broken in the code with this @HavilaQQ? Or is this just a developer experience thing?

HavilahOG:
> What specifically is broken in the code with this @HavilaQQ? Or is this just a developer experience thing?

It was more of a dev experience thing, not really needed anymore since trixter brought up the point of validation holes might appear elsewhere to allow uuid to be filled. More work for something minor.

DaneEveritt:
Gotcha, yeah usually anywhere I know for sure I need to insert a protected record I just use forceFill() against well validated data.

PR #2587, opened 2020-10-23
fix install scripts

parkervcp:
missed replacing an apk with apt

also fixes vanilla mc


alkanex-xyz:
the Paper installscript has also a failure:
https://pastebin.com/ju1cPdsS

it(jq) won't find Paper 1.16.1, trough ist available: https://papermc.io/api/v1/paper

PS: I checked the egg form your repo, which works. Could you include this in your PR?

PR #2592, opened 2020-10-24
Change SameSite attribute on session cookies to "lax"

Spirit55555:
Browsers are in the process of changing their default to "lax", if it has not been set, but developers are still encouraged to set it.

"lax" is the current default in Laravel: https://github.com/laravel/laravel/blob/bec982b0a3962c8a3e1f665e987360bb8c056298/config/session.php#L199

This change also make it possible to change it via an env variable, if you desire.


Spirit55555:
I just added another cookie fix to the pull request, as I think it is in the same category of improving cookie security.

Issue #2572, opened 2020-10-21
IP address overlap with too long ip alias


EriksenMartin:
**Background (please complete the following information):**
* Panel or Daemon: Panel
* Version: 1.0.1
* Server's OS: Ubuntu 18.04
* Your Computer's OS & Browser: Windows 10, Chrome

**To Reproduce**
1. Go to any server
2. Click on Network
3. Have a ip alias that is bigger then the div width

**Expected behavior**
I wanted to show the whole IP and not let it overlap.

![https://eriksen.app/i/n7x9p.png](https://eriksen.app/i/n7x9p.png)

PR #2633, opened 2020-10-31
Network view changes

notAreYouScared:
Allows address field to be copied to the clipboard when clicked
If an alias is used it will display `Hostname` instead of `IP Address` ( might just make it say `Address` as that would cover both? )
Fixed overflow for allocations with a long alias

image

chrome_2020-10-31_16-18-36


DaneEveritt:
closes #2572

PR #2658, opened 2020-11-04
Update egg-teamspeak3-server.json

pefman:
missing packet wget, install fails


parkervcp:
The current version in the repo should be working fine.

1.0.3 missed my last egg update for mumble and teamspeak

PR #2666, opened 2020-11-06
Use app key from environment

Yomanz:
The entry point for docker generates its own APP_KEY even if their is one in scope.

The entrypoint now checks if there is an APP_KEY in scope and will use that instead of overwriting it.


parkervcp (approved):
This is something I hadn't considered and it a great change.

DaneEveritt (approved):

PR #2392, opened 2020-09-22
Root admin allowed to use all functions even server is suspended, close #1654

zKoz210:
Full access to the server is working, new message about suspended

image

PR #2677, opened 2020-11-08
Root admin allowed to use all functions even server is suspended, close #1654

### LINKED ISSUE #1654: Root Admin Should be Allowed to Use File/Database Management Even Server is Suspended
I would like to be able to manage a server which is suspended if I am the root admin account,

currently, when I click on a suspended server with the root admin account, it will says suspended and I can't do anything, I would like to be able to use the database management and file management with a root admin if the server is suspended

**Additional context**

zKoz210:
Full access to the server is working, new message about suspended. ref #2392

image


zKoz210:
@DaneEveritt you could consider the pull request, there are not so complex changes, but I think it will be very useful for project administrators

lancepioch:
@zKoz210 Is this still an issue? If so, would you be able to fix the conflicts? I can then verify it's working and we can get it merged in shortly.

broimsuperman:
@zKoz210 can you fix the conflicts as @lancepioch said? If not, thanks for at least giving it a shot!

JackW6809:
The issue here I see now, is that Wings checks if the server is suspended. If it is, it will not allow the server to be started. Regardless of user,

server.go
```
// Checks if the server is marked as being suspended or not on the system.
func (s *Server) IsSuspended() bool {
	return s.Config().Suspended
}
```
power.go
```
// Disallow start & restart if the server is suspended. Do this check after performing a sync
// action with the Panel to ensure that we have the most up-to-date information for that server.
if s.IsSuspended() {
	return ErrSuspended
}
```

trenutoo:
> The issue here I see now, is that Wings checks if the server is suspended. If it is, it will not allow the server to be started. Regardless of user,
> 
> server.go
> 
> ```
> // Checks if the server is marked as being suspended or not on the system.
> func (s *Server) IsSuspended() bool {
> 	return s.Config().Suspended
> }
> ```
> 
> power.go
> 
> ```
> // Disallow start & restart if the server is suspended. Do this check after performing a sync
> // action with the Panel to ensure that we have the most up-to-date information for that server.
> if s.IsSuspended() {
> 	return ErrSuspended
> }
> ```

The change here is to allow admins to access the server, such as using the file manager. As seen in the screenshot, the power actions are hidden to admins if the server isSuspended, and normal users would have the suspended page.

DaneEveritt:
Due to PR merge policy changes and the fact that I abandoned my reviews of PRs so long that this has become quite stale, I am closing out this PR. Thank you for putting time into the project, and I am sorry that it did not end up being merged in.

PR #2682, opened 2020-11-09
Minor changes

notAreYouScared:
Changes CopyOnClick to allow any.
Allows database information to be copied on click.
Changes layouts on database/backups to match the network tab.
Changes text to lighten it one level from 400 to 300 for easier visibility.
Moves database API endpoints to their own folder for some organization.


schrej:
"Minor changes" is a bad commit message, especially since you're actually adding a new feature. The main reason is that it's absolutely impossible to guess what it's about when reading the message. The same also applies to pull requests btw.

Rather do sth like "add copy on click for databases and further enhancements", or something along those lines.

DaneEveritt:
Good note for the future. For now I can just fix it when I merge this.

Issue #2536, opened 2020-10-17
Backup rotation if limit is reached


Malachiel87:
**Is your feature request related to a problem? Please describe.**
When you hit the backup limit, the scheduled backup will fail

**Describe the solution you'd like**
Make a way to overwrite the oldest backup instead to fail the backup

**Describe alternatives you've considered**
none

**Additional context**
none

VibeGAMESNL:
This would be a great feature!

Mathsterk:
Another thing that would be nice to have is the ability to delete backups with a task. And the ability to delete multiple backups at a time.

james090500:
Both of these are much needed features. Especially when a coupe of days ago, Paper released an update which started deleting chunks! Backups are of upmost importance.

lasa01:
I looked into implementing this since I needed it myself.

The easiest way would be to make the rotation non-configurable, but some people may want to disable it.

I think the best way would be to have a "Backup Rotation" checkbox on the Create Task page. The problem with this is that a task can only have a single payload, which is currently used for ignored files on the "Create backup" task. So we would need some changes to allow multiple payloads per task, or we could encode both ignored files and backup rotation into the payload using JSON, or we could get rid of the "ignored files" payload, since it's configurable in the .pteroignore file anyway.

Another option would be a per-server setting to enable/disable backup rotations. It could be user-configurable like the server name is currently. However, I'm not sure if this should be in the "Backups" tab or in the "Settings" tab in the server view.

A third option would be to add a separate task that deletes the oldest backup if the backups are full. This would be relatively easy to implement and wouldn't require changes in task payload handling. It could be called something like "Rotate backups". The payload could be left unused, or it could be used for the number of how many backups to make room for, but I'm not sure if anyone would find that useful.

james090500:
Such a shame not to see this feature in 1.1.0. This is very much needed as we do not have unlimited disk space and need a valid way of managing backups.

Phyremaster:
The above issue that this issue was mentioned in was mine, and I closed it, because while it's not quite a duplicate, it's close enough that I should present it here as an alternative solution. It might be harder to implement, but I believe that it would be more flexible:

**Describe the solution you'd like**
Ideally, I would like to be able to configure my schedule(s) to overwrite backups that they have created previously, perhaps with a simple toggle. I would imagine this could be done by parsing the names of all backups with the format already used by scheduled backups, and removing the newest backup whose date and time is reasonably close to the schedule's last run date and time.

[...]

PR #2691, opened 2020-11-09
Backup rotation for schedules.

GravityCube:
Posible fix for #2536


app/Services/Backups/InitiateBackupService.php
```diff
@@ -96,13 +105,8 @@ public function setIgnoredFiles(?array $ignored)
      * @throws \Pterodactyl\Exceptions\Service\Backup\TooManyBackupsException
      * @throws \Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException
      */
-    public function handle(Server $server, string $name = null): Backup
+    public function handle(Server $server, string $name = null, bool $override = null): Backup
```
matthewpi: `bool $override = false`, not `bool $override = null`

app/Services/Backups/InitiateBackupService.php
```diff
@@ -111,6 +115,17 @@ public function handle(Server $server, string $name = null): Backup
             );
         }
 
+        if (! $server->backup_limit || $server->backups()->where('is_successful', true)->count() >= $server->backup_limit) {
+            if($override){
```
matthewpi: Space after `if`

app/Services/Backups/InitiateBackupService.php
```diff
@@ -111,6 +115,17 @@ public function handle(Server $server, string $name = null): Backup
...
+                // Remove latest backup
+                $last_backup = $server->backups()->where('is_successful', true)->oldest()->first();
+                $this->deleteBackupService->handle($last_backup);
+            }else{
```
matthewpi: Spaces before and after `else`

resources/scripts/components/server/schedules/TaskDetailsModal.tsx
```diff
@@ -91,7 +91,7 @@ const TaskDetailsForm = ({ isEditingTask }: { isEditingTask: boolean }) => {
                             <Label>Ignored Files</Label>
                             <FormikFieldWrapper
                                 name={'payload'}
-                                description={'Optional. Include the files and folders to be excluded in this backup. By default, the contents of your .pteroignore file will be used.'}
+                                description={'Optional. Include the files and folders to be excluded in this backup. By default, the contents of your .pteroignore file will be used.\nIf you dont have more backups spaces it will OVERRIDE the oldest.'}
```
matthewpi: `\nIf you have reached your backup limit, the oldest backup will be rotated.`

app/Services/Backups/InitiateBackupService.php
```diff
@@ -121,9 +136,7 @@ public function handle(Server $server, string $name = null): Backup
                 'disk' => $this->backupManager->getDefaultAdapter(),
             ], true, true);
 
-            $this->daemonBackupRepository->setServer($server)
-                ->setBackupAdapter($this->backupManager->getDefaultAdapter())
-                ->backup($backup);
+            $this->daemonBackupRepository->setServer($server)->setBackupAdapter($this->backupManager->getDefaultAdapter())->backup($backup);
```
matthewpi: It would be preferable if each function call was on a newline like it was previously.

app/Services/Backups/InitiateBackupService.php
```diff
@@ -111,6 +115,17 @@ public function handle(Server $server, string $name = null): Backup
             );
         }
 
+        if (! $server->backup_limit || $server->backups()->where('is_successful', true)->count() >= $server->backup_limit) {
```
matthewpi: Might want to add a comment above this line saying `"Check if the server has reached or exceeded it's backup limit"`

matthewpi (changes requested):

GravityCube:
@matthewpi Thanks for the feedback!

app/Services/Backups/InitiateBackupService.php
```diff
@@ -111,6 +115,18 @@ public function handle(Server $server, string $name = null): Backup
             );
         }
 
+        // Check if the server has reached or exceeded it's backup limit
+        if (! $server->backup_limit || $server->backups()->where('is_successful', true)->count() >= $server->backup_limit) {
+            if($override) {
+                // Remove latest backup
+                $last_backup = $server->backups()->where('is_successful', true)->oldest()->first();
```
DaneEveritt: Variables should be `$camelCase` and not `$snake_case`.

You also need to double-check that you don't have a null value returned here, since first can return a model, or a null value. I'd also prefer `where()->orderByDesc('created_at')->first()` so it is clear what we're ordering by, and intuitive to non-Laravel people what we're doing.

app/Services/Backups/InitiateBackupService.php
```diff
@@ -111,6 +115,18 @@ public function handle(Server $server, string $name = null): Backup
             );
         }
 
+        // Check if the server has reached or exceeded it's backup limit
+        if (! $server->backup_limit || $server->backups()->where('is_successful', true)->count() >= $server->backup_limit) {
+            if($override) {
```
DaneEveritt: I try to avoid `if/else` blocks like this, and lean towards a cleaner negation method of:

```
if (!$override) {
    throw X
}

// Do the other thing here
```

This reads a little cleaner to me and makes it a little clearer that there is a single thing that will cause the process to exit.

DaneEveritt (changes requested):
I imagined this being a totally different approach but like the simplicity. We can build off this later to support having backups that don't get rotated out automatically and whatnot, but this should solve a lot of people's pain pretty easily.

resources/scripts/components/server/schedules/TaskDetailsModal.tsx
```diff
@@ -91,7 +91,7 @@ const TaskDetailsForm = ({ isEditingTask }: { isEditingTask: boolean }) => {
                             <Label>Ignored Files</Label>
                             <FormikFieldWrapper
                                 name={'payload'}
-                                description={'Optional. Include the files and folders to be excluded in this backup. By default, the contents of your .pteroignore file will be used.'}
+                                description={'Optional. Include the files and folders to be excluded in this backup. By default, the contents of your .pteroignore file will be used.\nIf you have reached your backup limit, the oldest backup will be rotated.'}
```
DaneEveritt: Does the `\n` work correctly here? I thought that would give you a string literal `\n` in the output, and not an actual newline? I don't even think it needs a newline honestly.

> Optional. Include the files and folders to be excluded in this backup. By default, the contents of your .pteroignore file will be used, and old backups will be automatically removed once your limit is reached.

app/Services/Backups/InitiateBackupService.php
GravityCube: About the null value I don't think its needed. I added a count check anyway. Its that ok?

app/Services/Backups/InitiateBackupService.php
```diff
@@ -111,6 +115,18 @@ public function handle(Server $server, string $name = null): Backup
...
         }
 
+        // Check if the server has reached or exceeded it's backup limit
+        if (!$server->backup_limit || $server->backups()->where('is_successful', true)->count() >= $server->backup_limit) {
+            // Do not allow the user to continue if this server is already at its limit and can't override.
+            if (!$override || $server->backup_limit <= 0) {
+                throw new TooManyBackupsException($server->backup_limit);
+            }
+
+            // Remove latest backup
```
matthewpi: Is this actually the latest backup or is it the oldest backup?

GravityCube: Oh right. Sorry.

DaneEveritt (approved):

PR #2697, opened 2020-11-10
Fix RebuildServerCommand

ArnaudLier:
I don't know if this will get merged as there is 1.0 now but it's technically a bug so 🤷


matthewpi:
Rebuilding a container has been removed in 1.0 as the container is recreated with the latest settings whenever the server is started.

EDIT: I see this is for 0.7 which is in a security patch/major bug fixes only state, I doubt this is going to be merged as 1.0 is released and people should be updating if possible.

FoksVHox:
> Rebuilding a container has been removed in 1.0 as the container is recreated with the latest settings whenever the server is started.

That is a very nice thing.

PR #2718, opened 2020-11-16
Update index.blade.php

joshcannotcode:
Just to avoid confusion


ClumsyAdmin:
Linux doesn't call them threads though, they are just cpu cores (even though they may not be physical).

joshcannotcode:
Yeah I'm aware, I just think it would avoid confusion with certain people who might think it's referring to physical cores.

revij:
While it's of course nice to be able to say "Hey, you have X threads or X cores assigned", if someone has disabled hyper threading then it may be confusing to them.

PR #2763, opened 2020-11-28
Hide delete button on own subuser if for the subuser looking at the page

notAreYouScared:
Don't show the delete button if a sub-user is looking at sub-users, as they cannot delete themselves. goes along with them not being able to edit themselves that's already a thing.


DaneEveritt:
@AreYouRlyScared can you put the `if` check outside the `<Can/>` component, just logically makes more sense to me and follows the design for other component stuff. 👍

notAreYouScared:
@DaneEveritt Done.

resources/scripts/components/server/users/UserRow.tsx
```diff
@@ -48,23 +48,23 @@ export default ({ subuser }: Props) => {
...
-                    <FontAwesomeIcon icon={faPencilAlt}/>
-                </button>
-                }
-            </Can>
-            <Can action={'user.delete'}>
-                {subuser.uuid !== uuid &&
-                <RemoveSubuserButton subuser={subuser}/>
-                }
-            </Can>
+            {subuser.uuid !== uuid &&
```
DaneEveritt, line 51:
```
{subuser.uuid !== uuid &&
    <>
        {the two <can> blocks}
    </>
}
```

DaneEveritt (changes requested):

notAreYouScared (commented):
Changed

PR #2780, opened 2020-12-04
Allow 0 in first octet of database client IP address Fixes #2779

### LINKED ISSUE #2779: Database Connections from field does not accept 10.0.0.0/8 range addresses
When creating a new database on a server the connections from field does not allow you to use 10.0.0.0/8 range IPs (for example 10.10.2.%). It instead gives "A valid connection address must be provided" error. 
These IPs are reserved for use in private networks so it should be valid.

danny6167:
Allow user to enter database connection IP address that contains a a 0 in the first octet. Fixes #2779

With the current regex, any IP address that contains a 0 anywhere in the first octet of the IP is rejected by the front end.

### DIFF
--- a/resources/scripts/components/server/databases/CreateDatabaseButton.tsx
+++ b/resources/scripts/components/server/databases/CreateDatabaseButton.tsx
@@ -24,7 +24,7 @@ const schema = object().shape({
         .matches(/^[A-Za-z0-9_\-.]{3,48}$/, 'Database name should only contain alphanumeric characters, underscores, dashes, and/or periods.'),
     connectionsFrom: string()
         .required('A connection value must be provided.')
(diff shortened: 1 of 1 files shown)


parkervcp:
Each octet should only allow 0-255. Technically 1-255 as you can't send to 0.0.0.0.

schrej:
It's actually a little more complicated than that: https://dev.mysql.com/doc/refman/8.0/en/account-names.html

DaneEveritt:
Can someone provide a valid example of allowing `0[anything]` as the IP address here or is this just being pendantic?

DaneEveritt:
Should this be `[1-9][0-9]{0,2}` in that case, rather than just `[0-9]`? But thats just getting overly specific I suppose. LGTM 👍

PR #2792, opened 2020-12-06
Handle CompleteMultipartUpload and AbortMultipartUpload on the Panel

matthewpi:
This moves the `CompleteMultipartUpload` and `AbortMultipartUpload` S3 requests onto the panel instead of in wings which helps avoid any problems with S3 urls expiring.

I also have added the `BACKUP_PRESIGNED_URL_LIFESPAN` environment variable which sets the lifespan of the `UploadPart` S3 urls in minutes, it currently defaults to `60` (1 hour).

https://github.com/pterodactyl/wings/pull/73

### DIFF
--- a/app/Http/Controllers/Api/Remote/Backups/BackupRemoteUploadController.php
+++ b/app/Http/Controllers/Api/Remote/Backups/BackupRemoteUploadController.php
@@ -6,7 +6,9 @@ use Carbon\CarbonImmutable;
 use Illuminate\Http\Request;
 use Pterodactyl\Models\Backup;
 use Illuminate\Http\JsonResponse;
+use Illuminate\Foundation\Application;
 use League\Flysystem\AwsS3v3\AwsS3Adapter;
+use Illuminate\Contracts\Config\Repository;
 use Pterodactyl\Http\Controllers\Controller;
 use Pterodactyl\Extensions\Backups\BackupManager;
 use Pterodactyl\Repositories\Eloquent\BackupRepository;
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/Api/Remote/Backups/BackupRemoteUploadController.php
```diff
@@ -69,15 +78,15 @@ public function __invoke(Request $request, string $backup)
...
+            throw new BadRequestHttpException('The configured backup adapter is not an S3 compatible adapter.');
         }
 
         // The path where backup will be uploaded to
         $path = sprintf('%s/%s.tar.gz', $backup->server->uuid, $backup->uuid);
 
         // Get the S3 client
         $client = $adapter->getClient();
-        $expires = CarbonImmutable::now()->addMinutes(30);
+        $expires = CarbonImmutable::now()->addMinutes($this->config->get('backups.presigned_url_lifespan', 60));
```
DaneEveritt: Just use `config()`, we don't need the repository for this.

app/Http/Controllers/Api/Remote/Backups/BackupStatusController.php
```diff
@@ -47,13 +59,42 @@ public function __invoke(ReportBackupCompleteRequest $request, string $backup)
...
+            $params = [
+                'Bucket' => $adapter->getBucket(),
+                'Key' => sprintf('%s/%s.tar.gz', $backup->server->uuid, $backup->uuid),
+                'UploadId' => $request->input('upload_id'),
+            ];
+
+            // If the backup was not successful, send an AbortMultipartUpload request.
+            if (! $successful) {
+                $client->execute($client->getCommand('AbortMultipartUpload', $params));
+                return new JsonResponse([], JsonResponse::HTTP_NO_CONTENT);
```
DaneEveritt: Blank line before returns.

DaneEveritt (changes requested):
see comments

DaneEveritt (approved):

PR #2858, opened 2020-12-11
Add linting to the file editor

### LINKED ISSUE #2853: Linter
It's frustrating to mess up on a config file then look through it manually / use an external editor to see where the error is.

Add a linter, similar in manner to [this](https://codemirror.net/demo/lint.html)

Button that lints it instead of having it show up in the side, like [this](http://www.yamllint.com/)

**Additional context**
not much

matthewpi:
Resolves #2853 

Editor bundle size is now `1.26 MiB` in a production build.


resources/scripts/components/elements/CodemirrorEditor.tsx
```diff
@@ -79,6 +78,31 @@ require('codemirror/mode/vue/vue');
 require('codemirror/mode/xml/xml');
 require('codemirror/mode/yaml/yaml');
 
+require('codemirror/addon/hint/css-hint');
+require('codemirror/addon/hint/html-hint');
```
schrej, line 82: I don't think we really need css and html, do we? I'd limit it to configuration and scripting languages common in available eggs.

matthewpi: We supported those languages already so I added the linting for them, some people probably edit those files but there are probably very few.

PR #2859, opened 2020-12-11
Add confirm alert before restart and stop server

Aldiwildan77:
I think this is important because sometimes people accidentally press the restart & stop button.


DaneEveritt:
I disagree with this, notably because it has been like this for five years and infrequently been an issue. Additionally, these buttons are off to the left side out of the way of normal console interaction, and therefore unlikely to be in a location that would be accidentally clicked.

People can accidentally press all sorts of things. Putting a blocking popup on every such action would just get tedious very quickly, especially for people that want to be able to press a stop button and have their server stop without interaction.

Finally, this is not in line with the rest of the design of this software, we don't use generic browser popup actions for alerts.

PR #2871, opened 2020-12-13
Added translation for AutoAllocationNotEnabledException

lukedaviskzn:
The exception was in plain text, so I added a translation for it.


DaneEveritt:
Thanks, but right now translations are in a messy limbo and I'd rather not make things even more difficult on myself, so the hard-coded string is intentional.

PR #2879, opened 2020-12-16
Allow changing the prune age for backups

matthewpi:
Adds the `BACKUP_PRUNE_AGE` environment variable (default value is `60`, value is in minutes).  This value is used for determining if a backup has taken too long and needs to be deleted automatically.

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -25,11 +25,13 @@ class Kernel extends ConsoleKernel
         // Execute scheduled commands for servers every minute, as if there was a normal cron running.
         $schedule->command('p:schedule:process')->everyMinute()->withoutOverlapping();
 
-        // Every 30 minutes, run the backup pruning command so that any abandoned backups can be removed
-        // from the UI view for the server.
-        $schedule->command('p:maintenance:prune-backups', [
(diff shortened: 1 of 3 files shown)


DaneEveritt:
Never having it prune runs a very high risk of backups just constantly showing up as the spinner icon if there is ever something that bugs out and causes the status to not be reported. This impacts user ability to create additional backups until someone goes in and soft-deletes said backup.

Basically, I am strongly opposed to allowing it to be disabled rather than just making it run every 12 hours or something by default to avoid 99% of issues.

matthewpi:
> Never having it prune runs a very high risk of backups just constantly showing up as the spinner icon if there is ever something that bugs out and causes the status to not be reported. This impacts user ability to create additional backups until someone goes in and soft-deletes said backup.
> 
> Basically, I am strongly opposed to allowing it to be disabled rather than just making it run every 12 hours or something by default to avoid 99% of issues.

I can remove that functionality but someone could easily just set the time to `999999` or something ridiculous to get the same effect anyways.

DaneEveritt:
Yes. Always true. Lets go with 6 hours as the default time, should be fine for 99% of users.

PR #2883, opened 2020-12-17
Add Transfer Logs

matthewpi:
This adds the functionality to view transfer logs similarly to install logs.

Instead of suspending the server, I track the `transfer` relation of the server to determine if the server is currently being transferred to independently lock endpoints or functionality that would interfere with a server being transferred (such as #2677 allowing access to the file manager).

I am pretty confident that these changes are stable but further testing and code review would be appreciated.


app/Http/Middleware/Api/Client/Server/AuthenticateServerAccess.php
```diff
@@ -65,17 +63,26 @@ public function handle(Request $request, Closure $next)
...
                 'This server is currently suspended and the functionality requested is unavailable.'
             );
         }
 
-        if (! $server->isInstalled()) {
-            // Throw an exception for all server routes; however if the user is an admin and requesting the
-            // server details, don't throw the exception for them.
-            if (! $user->root_admin || ($user->root_admin && ! $request->routeIs($this->except))) {
-                throw new ConflictHttpException('Server has not completed the installation process.');
+        // Still allow users to get information about there server if it is installing or being transferred.
```
matthewpi: "about there" -> "about their"

app/Http/Middleware/Api/Client/Server/AuthenticateServerAccess.php
```diff
@@ -65,17 +63,26 @@ public function handle(Request $request, Closure $next)
...
+                // Throw an exception for all server routes; however if the user is an admin and requesting the
+                // server details, don't throw the exception for them.
+                if (! $user->root_admin || ($user->root_admin && ! $request->routeIs($this->except))) {
+                    throw new ConflictHttpException('Server has not completed the installation process.');
+                }
+            }
+
+            if ($server->transfer !== null) {
+                if (! $user->root_admin || ($user->root_admin && ! $request->routeIs($this->except))) {
+                    throw new ConflictHttpException('Server is currently being transferred.');
```
DaneEveritt: Can make a `ServerTransferingException` that implements the `HttpExceptionInterface` like some others do so it is a defined exception with a preset message and response code. 👍🏼

app/Http/Middleware/Api/Client/Server/AuthenticateServerAccess.php
```diff
@@ -65,17 +63,26 @@ public function handle(Request $request, Closure $next)
...
+        if (! $request->routeIs('api:client:server.view')) {
+            if (! $server->isInstalled()) {
+                // Throw an exception for all server routes; however if the user is an admin and requesting the
+                // server details, don't throw the exception for them.
+                if (! $user->root_admin || ($user->root_admin && ! $request->routeIs($this->except))) {
+                    throw new ConflictHttpException('Server has not completed the installation process.');
+                }
+            }
+
+            if ($server->transfer !== null) {
```
DaneEveritt: Prefer `is_null`

PR #2887, opened 2020-12-18
Favicons wasent point to right folder

Pomdre:
Fixed that manifest wasent point to the folder favicons. Same with browserconfig.

### DIFF
--- a/public/favicons/browserconfig.xml
+++ b/public/favicons/browserconfig.xml
@@ -1,2 +1,2 @@
 <?xml version="1.0" encoding="utf-8"?>
-<browserconfig><msapplication><tile><square70x70logo src="/ms-icon-70x70.png"/><square150x150logo src="/ms-icon-150x150.png"/><square310x310logo src="/ms-icon-310x310.png"/><TileColor>#ffffff</TileColor></tile></msapplication></browserconfig>
+<browserconfig><msapplication><tile><square70x70logo src="/favicons/ms-icon-70x70.png"/><square150x150logo src="/favicons/ms-icon-150x150.png"/><square310x310logo src="/favicons/ms-icon-310x310.png"/><TileColor>#ffffff</TileColor></tile></msapplication></browserconfig>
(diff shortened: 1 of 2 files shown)


trenutoo:
Why do you change the density of the 36x36 image from low-density to medium-density?

Also, your manifest and config changes break it. Their paths are correct since they're located inside the favicon folder.

Pomdre:
Is density still a thing? Does not say anything about it see url
[Mozilla](https://developer.mozilla.org/en-US/docs/Web/Manifest/icons)

trenutoo:
> Is density still a thing? Does not say anything about it see url
> [Mozilla](https://developer.mozilla.org/en-US/docs/Web/Manifest/icons)

Yes, they're still a thing, if you're not sure, then why make any changes?

Also, your paths are broken and invalid. The manifest and config files are already inside the favicons folder, you are now requesting icons from a non-existing path.

Pomdre:
image
When you do not have /favicons, it does not point to anything. Try navigating to ./favicon.ico it wil result in 404 But whan you navigate to /favicons/favicon.icon you wil get a result

trenutoo:
> image
> When you do not have /favicons, it does not point to anything. Try navigating to ./favicon.ico it wil result in 404 But whan you navigate to /favicons/favicon.icon you wil get a result

I think you're confused about how files are fetched on a web server, your PR will break the favicons. The files are located inside public/favicon hence why you can only access them via domain/favicons. 

The manifest and config are located inside the favicon folder, the paths for them are inside that folder. Your PR change is looking for a favicon folder inside the favicon folder which does not exist.

Pomdre:
Makes sense.  But why do I get a error message when it points to ./?

trenutoo:
> Makes sense. But why do I get a error message when it points to ./?

They're hosted at `public/favicon` and the website fetches them from there to display to users. If you want to directly access them yourself through your browser, then you must specify the full hosted path which is `/favicon/<file>`.

Your change would be basically looking for `/favicon/favicon/<file>`

Pomdre:
Okay, sorry for waisting your time. Thanks!

PR #2917, opened 2020-12-27
Separate permission for downloading files

UUBlueFire:
I require a separate permission for downloading files from the panel. I have implemented and tested this on my own panel.

### DIFF
--- a/app/Models/Permission.php
+++ b/app/Models/Permission.php
@@ -49,6 +49,7 @@ class Permission extends Model
     const ACTION_ALLOCATION_DELETE = 'allocation.delete';
 
     const ACTION_FILE_READ = 'file.read';
+    const ACTION_FILE_DOWNLOAD = 'file.download';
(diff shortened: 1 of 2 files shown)


DaneEveritt:
How is this any different than the read content one?

UUBlueFire:
I would like it to be so that people can view the contents of a file but not download the file

DaneEveritt:
There is literally no difference is what I'm saying. If they want to download the file it is trivial for them to do so if they can already view the contents, and absolutely nothing would stop them with SFTP access. In fact, this permission only makes it more confusing for things like SFTP which is why the permissions are laid out how they are.

Appreciate the idea, but it doesn't really work well with how everything else is setup permissions wise.

diamkil:
@DaneEveritt Yeah but people add users to configure plugins but don't want them downloading the paid plugins, maybe the download perm should be on the SFTP permission instead of read-contents

PR #2920, opened 2020-12-28
Add CopyOnClick to sftp information

notAreYouScared:
Allows the end-user to click the sftp address and username to copy it to the clipboard.


notAreYouScared:
Githubbing is hard.

PR #2952, opened 2021-01-01
fix file manager transition animations

schrej:
There seem to be CSSTransitions all over the place, some of which are either not needed, or should be applied somewhere else (at least in my opinion).

Removing the transition from the PageContentBlock and using the pathname and search of the location as the TransitionRouter key does not seem to break any animations, but doing both fixes the file manager reloading entirely on each click.

Maybe I'm also missing something though.


schrej:
Wait, this is ugly and not the actual problem.

PR #2953, opened 2021-01-01
Keep copyright notice up to date

DarthShmev:
Automatically uses the current year for the copyright notice. This will avoid the need for manual updates in the future.


DaneEveritt:
I can handle this on my own time, thanks though. It really isn't that important in the end. 👍🏼

PR #2955, opened 2021-01-01
Add image selection for minecraft java eggs for java 8 & 11

TekExplorer:
Set minecraft eggs to have both java versions as options by default.
is there some way to set a default or an alias for images so that users know what they are?

### DIFF
--- a/database/seeds/eggs/minecraft/egg-sponge--sponge-vanilla.json
+++ b/database/seeds/eggs/minecraft/egg-sponge--sponge-vanilla.json
@@ -8,7 +8,7 @@
     "author": "support@pterodactyl.io",
     "description": "SpongeVanilla is the SpongeAPI implementation for Vanilla Minecraft.",
     "features": ["eula"],
-    "images": "quay.io\/pterodactyl\/core:java-glibc",
+    "images": ["quay.io\/pterodactyl\/core:java-glibc"],
     "startup": "java -Xms128M -Xmx{{SERVER_MEMORY}}M -jar {{SERVER_JARFILE}}",
     "config": {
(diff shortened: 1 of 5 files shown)


TekExplorer:
actually, perhaps the deprecated `image` could be used to set the default?

matthewpi:
> actually, perhaps the deprecated `image` could be used to set the default?

No point as these changes will only affect newer versions of the panel that have support for multiple images.

TekExplorer:
> > actually, perhaps the deprecated `image` could be used to set the default?
> 
> No point as these changes will only affect newer versions of the panel that have support for multiple images.

what i mean is, while `images` can be used to specify the list, `image` could be used to set the default image *for* the multi-image supported versions, which plays well with keeping backwards compatibility i think as older versions would just use that "default"

DaneEveritt:
The default image is just the first one, no need for complexity.

database/seeds/eggs/minecraft/egg-sponge--sponge-vanilla.json
```diff
@@ -8,7 +8,7 @@
     "author": "support@pterodactyl.io",
     "description": "SpongeVanilla is the SpongeAPI implementation for Vanilla Minecraft.",
     "features": ["eula"],
-    "image": "quay.io\/pterodactyl\/core:java-glibc",
+    "images": "quay.io\/pterodactyl\/core:java-glibc",
```
DaneEveritt: This needs to be an array.

DaneEveritt (changes requested):
see comment

database/seeds/eggs/minecraft/egg-sponge--sponge-vanilla.json
TekExplorer: good catch, didnt notice

PR #2957, opened 2021-01-02
URL encode mail address in password reset link

### LINKED ISSUE #2896: Impossible to reset password when email contains "+"
* Panel or Daemon: Panel
* Version of Panel/Daemon: 1.1.3 
* Server's OS: Ubuntu 20.04
* Your Computer's OS & Browser: Windows 10, Chrome

**Describe the bug**
Users with a + in their email cannot reset their password. They can request to reset their password and will receive a password link to reset it. However, clicking on that password link will show them a page that resets the password for an account that has the "+" replaced with a space. For instance, `purpur+testing@gmail.com` would be taken to a password reset link for email `purpur testing@gmail.com` since the "+" in the URL bar is interpreted as a space. Upon clicking reset password, they'll be told that the email is an invalid email address since the email address should have a + rather than a space. Example:
image

**To Reproduce**
1. Use an email that you have access to. Use a provider that supports plus addressing such as Gmail or Outlook. Create a panel account with an email that has +[something] suffixed to the username. For example, if your email is `name@gmail.com`, create an account with the email `name+testing@gmail.com`.
2. Click on forgot password on the login screen and try to reset your password
3. Click on the link given to you in your email
4. You'll be unable to reset your password because `name testing@gmail.com` doesn't exist; `name+testing@gmail.com` does

**Expected behavior**
You should be able to reset your password and the email should show up as `name+testing@gmail.com` in the password reset screen.

**Potential solutions**
I think a possible solution could just be to make the panel replace all spaces it receives with a "+" on the password reset email field since spaces are not possible in emails.

schrej:
fixes #2896

this took waaaay too long :smile:

### DIFF
--- a/app/Notifications/SendPasswordReset.php
+++ b/app/Notifications/SendPasswordReset.php
@@ -57,7 +57,7 @@ class SendPasswordReset extends Notification implements ShouldQueue
         return (new MailMessage)
             ->subject('Reset Password')
             ->line('You are receiving this email because we received a password reset request for your account.')
-            ->action('Reset Password', url('/auth/password/reset/' . $this->token . '?email=' . $notifiable->email))
+            ->action('Reset Password', url('/auth/password/reset/' . $this->token . '?email=' . urlencode($notifiable->email)))
             ->line('If you did not request a password reset, no further action is required.');
     }
 }


sergejostir:
You should probably follow the Laravel's example: https://github.com/laravel/framework/blob/502ed3a5b6d82bfe8e6792d42edb2b0c6ba1b2d4/src/Illuminate/Auth/Notifications/ResetPassword.php#L69-L72

So it should be:
`            ->action('Reset Password', url(route('auth.reset', ['token' => $this->token, 'email' => $notifiable->email], false)))`

DaneEveritt:
We don't use named routes on the frontend, so this is fine.

PR #2977, opened 2021-01-05
Fixed "migrate:refresh" command

zKoz210:


DaneEveritt:
Please include the error this is fixing, I've never seen it happen in test runs or production environments.

PR #2985, opened 2021-01-07
Properly add the version to Docker builds

matthewpi:
Uses the same logic the release workflow does to add the release version to the `config/app.php` file.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,46 +1,36 @@
 name: Publish Docker Image
-
 on:
   push:
     branches:
       - 'develop'
     tags:
       - 'v*'
-
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -1,37 +1,54 @@
...
+
       - uses: crazy-max/ghaction-docker-meta@v1
         id: docker_meta
         with:
           images: ghcr.io/pterodactyl/panel
+
       - uses: docker/setup-qemu-action@v1
+
       - uses: docker/setup-buildx-action@v1
+
```
DaneEveritt: please don't add these gaps

DaneEveritt (changes requested):
see comment

Issue #2995, opened 2021-01-12
Teamspeak Egg wrong Query Startup Command


Paddyboy1991:
Hello,
I found a wrong Startup Command at the Teamspeak Egg..
The standard Query Port from Teamspeak is 10011 and the Container start with this Port, but in the completly Egg is in the startup command the SERVER_PORT defined.. so is my Teamspeak Port for Users (standart 9987) then is the query port the same port too.. I miss a option to define the query port extra..

![Screenshot_1](https://user-images.githubusercontent.com/16993167/104266087-391a1800-548f-11eb-9c68-87843a0f88f4.png)
Filemanager have an Option but Query Port not.
Cheers

Slitzo3:
I think this is a issue for this repo: https://github.com/parkervcp/eggs

As this is the repo for the panel/wings. So i would suggest you to post the issue there and close it here.

gOOvER:
Teamspeak Egg is default shipped with ptero, so the Report is fine here

Paddyboy1991:
> I think this is a issue for this repo: https://github.com/parkervcp/eggs
> 
> As this is the repo for the panel/wings. So i would suggest you to post the issue there and close it here.

"name": "Teamspeak3 Server",
"author": "support@pterodactyl.io",

i think it is right here ;)

PR #2997, opened 2021-01-12
Teamspeak 3 Egg - added queryport and use default query port

gOOvER:
At the Moment, the Teamspeak 3 Egg uses the Serverport as Queryport. I added default Quryport 10011.


gOOvER:
Fix #2995

PR #3004, opened 2021-01-15
Optimize Images

lpuv:
Simple PR, just thought it would be appreciated.


DaneEveritt:
I don't see a reason to do this, the changes are basically nothing and there is no dramatic difference in anything. I also generally dislike PRs that are just some bot action that occurred.

PR #3021, opened 2021-01-21
Fix 'other servers' toggle being persisted across non-admin accounts

matthewpi:
The toggle only available to admins on the landing page of the panel is persisted using local storage but if you logout of an admin account, then login with a regular user account, the state is still persisted which can hide your own servers from view.


DaneEveritt:
Thanks for finding this, actually just going to key this more uniquely in the UI so we don't constantly have to fight it.

PR #3022, opened 2021-01-21
Update egg-forge-minecraft.json

jmiln:
This changes it so the FORGE_JAR gets set, so it can have it's name changed to SERVER_JARFILE at the end


DaneEveritt:
Please fix file conflicts.

PR #3027, opened 2021-01-24
Allow copy on click for SFTP information

notAreYouScared:
Allows the end-user to click the sftp address and username to copy it to the clipboard.

Duplicate of https://github.com/pterodactyl/panel/pull/2920 ... just without the un-needed 5 commits


notAreYouScared:
So over this, the tests past but they're dumb

https://github.com/pterodactyl/panel/runs/1755907034 < passing test from last pr

PR #3101, opened 2021-02-18
Set CSGO Steam Account Token to nullable

Okupateco:
Set the Steam Account Token to nullable so it's not required, makes it work out of the box for the WHMCS module.


trenutoo:
It's set to required, because without the GSLT token server will fail to start. You should pass it through your WHMCS using configurable options or set it to nullable if you prefer so.

Making this nullable globally would most likely cause a lot more people to complain about their server not working. Required variables for the server to function are always set to required

PR #3110, opened 2021-02-22
Added SERVER_NAME environment variable

aussieserverhosts:
Added SERVER_NAME environment variable to stop laravel framework server name defaulting to localhost, causing mail relays such as Gmail to stop silently dropping emails due to sender name being localhost.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -25,7 +25,11 @@ MAIL_USERNAME=
 MAIL_PASSWORD=
 MAIL_ENCRYPTION=tls
 MAIL_FROM=no-reply@example.com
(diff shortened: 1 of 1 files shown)


DaneEveritt:
Shouldn't this be blank by default or just commented out? Haven't really heard of it being an issue in the past. I do know that setting that example.com as the default will probably break things.

aussieserverhosts:
Either or. If left undefined, it will default to 'localhost' which will cause email providers like Google to reject emails despite having a whitelisted IP and correct headers

DaneEveritt:
Lets leave it blank and include a comment above it (use `#`) indicating what it does and maybe a link back to this PR? :)

aussieserverhosts:
Updated. Feel free to change it to something neater!

PR #3112, opened 2021-02-22
Change Rust RCON Password input rule

trenutoo:
RCON will fail to connect if the password contains special characters such as á,ö,ä, /%¤&. This input rule resolves that issue by only accepting alphanumeric letters, dots, underscores, and dashes.

### DIFF
--- a/database/Seeders/eggs/rust/egg-rust.json
+++ b/database/Seeders/eggs/rust/egg-rust.json
@@ -121,7 +121,7 @@
             "default_value": "CHANGEME",
             "user_viewable": true,
             "user_editable": true,
-            "rules": "required|regex:\/^[a-zA-Z0-9_.-]*$\/|max:64"
+            "rules": "required|regex:/^[\w.-]*$/|max:64"
         },
         {
             "name": "Save Interval",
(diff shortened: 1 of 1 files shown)


DaneEveritt:
`\w-.` is much shorter for this regex. :)

There is also the `alpha_num` rule but that might exclude `-` and `.` not sure.

trenutoo:
> `\w-.` is much shorter for this regex. :)
> 
> There is also the `alpha_num` rule but that might exclude `-` and `.` not sure.

alpha_num and alpha_dash all accept special characters such as ä ö á. `\w-.` doesn't tag underscores, dashes, or dots and might be more difficult to understand for users who don't know regex at all?

Personally, I can't come up with any shorter regex for those values, please let me know if you have any suggestions and I can adjust it accordingly.

DaneEveritt:
`^[\w.-]*$` is fine, and honestly I prefer a more terse regex, it is way easier to understand that `\w` means alpha numeric with underscores. Also, realistically, if you don't understand regex neither format is going to make sense.

https://regex101.com/r/Ujgzol/1

trenutoo:
> `^[\w.-]*$` is fine, and honestly I prefer a more terse regex, it is way easier to understand that `\w` means alpha numeric with underscores. Also, realistically, if you don't understand regex neither format is going to make sense.
> 
> https://regex101.com/r/Ujgzol/1

Yeah that works, thank you

PR #3114, opened 2021-02-22
Fix server.jar renaming

trenutoo:
This resolves the issue of the .jar file never being renamed if you use the FORGE_VERSION variable to use a specific version. The forge jar value overwrite was missing, thus the renaming of the file from forge-${FORGE_VERSION}.jar into server.jar always failed

This exact same file is currently present on Parker's egg repo.


parkervcp (approved):
LGTM

parkervcp:
@Software-Noob Please make sure this makes it to the egg repo too please.

PR #3115, opened 2021-02-22
Changing daemonSFTP to unsigned

FC-FailCoder:
Just a little spelling change can make the issues go away :)


DaneEveritt:
Sorry, this is not a proper fix, it needs to be a migration to convert existing records to unsigned, not just new installs.

PR #3160, opened 2021-03-08
Textarea can't be resized horizontally

ghostdevv:
I noticed on pages such as the node add page where you have the description textarea it's able to be resized horizontally but it doesn't make sense to have that since it would then be bigger than it's container, see here:

[](https://i.imgur.com/OsaMtML.gif)

With this change you can now only resize it vertically:

[](https://i.imgur.com/t104IaC.gif)


DaneEveritt:
Thanks, but we're redesigning the entire admin area into react and we can fix this then since it isn't really impacting much right now. :)

Issue #3161, opened 2021-03-08
Self upgrade doesn't upgrade if you manually download the files. 


FoksVHox:
**Background (please complete the following information):**
* Panel or Daemon: Panel
* Version of Panel/Daemon: 1.3.0
* Server's OS: Ubuntu 20.04
* Your Computer's OS & Browser: Windows 10 Student & Brave

**Describe the bug**
I use PHP 8.0.3 and I've followed the installation guide. 

When you run `p:upgrade` and choose "no" when it asks you if you'd like to download the files, assuming that you've downloaded the files manually it won't unpack the panel.tar.gz, it'll run the composer, and migrations, but it won't unpack. 

**To Reproduce**
1. Download the files from GitHub manually using `curl -Lo panel.tar.gz https://github.com/pterodactyl/panel/releases/latest/download/panel.tar.gz` 
2. Run `p:upgrade`
3. Choose "No" when it asks you if you'd like to download the files
4. Say yes to the two next questions
5. It'll run, and then you go to the admin area and see that your panel version hasn't changed. (And assumingly it hasn't upgraded)

**Expected behavior**
It'll unpack the files, and upgrade the panel correctly.

matthewpi:
I believe this is expected behavior, if you use choose not to download the files, it will not extract a local archive, it's expecting you to have already done that.  However, the command will still run the rest of the commands that are generally needed for the upgrade process (`down`, `chmod`, `composer install`, `view:clear`, `config:clear`, `migrate`, `chown`, `queue:restart`, and finally `up`).  If you would like this "already downloaded" behavior, that seems like it would be a feature request, but this behavior seems to be working correctly.

FoksVHox:
> I believe this is expected behavior, if you use choose not to download the files, it will not extract a local archive, it's expecting you to have already done that. However, the command will still run the rest of the commands that are generally needed for the upgrade process (`down`, `chmod`, `composer install`, `view:clear`, `config:clear`, `migrate`, `chown`, `queue:restart`, and finally `up`). If you would like this "already downloaded" behavior, that seems like it would be a feature request, but this behavior seems to be working correctly.

Okay, I just thought it would extract the archive :)

PR #3162, opened 2021-03-08
Fix artisan upgrade failing to download

### LINKED ISSUE #3159: Self Update fails because files are not downloaded
* Panel or Daemon: Panel
* Version of Panel/Daemon: 1.3.0
* Server's OS: Debian 9
* Your Computer's OS & Browser: Win 10 & Opera

**Describe the bug**
When running the new self update command curl fails to download the new files:
> $upgrader> curl -L "https://github.com/pterodactyl/panel/releases/latest/download/panel.tar.gz" | tar -xzv
> [...]
> curl: (23) Failed writing body (402 != 1371)

**To Reproduce**
1. Go to your panel's dir
2. Run 'php artisan p:upgrade'
3. Start the process with 3 times 'yes'
5. Installation "finishes" but with old files

**Expected behavior**
Download the newest files. :D

trenutoo:
Tar -f flag is not required in this context as no filename is specified. The output is passed to stdout without it and should invoke the unarchiving properly as the artisan upgrade is executed in the webroot directory. I've tested it myself a couple of times and was able to upgrade and downgrade.

Resolves #3159, #3161

### DIFF
--- a/app/Console/Commands/UpgradeCommand.php
+++ b/app/Console/Commands/UpgradeCommand.php
@@ -79,7 +79,7 @@ class UpgradeCommand extends Command
         if (!$skipDownload) {
             $this->withProgress($bar, function () {
                 $this->line("\$upgrader> curl -L \"{$this->getUrl()}\" | tar -xzv");
-                $process = Process::fromShellCommandline("curl -L \"{$this->getUrl()}\" | tar -xzvf");
+                $process = Process::fromShellCommandline("curl -L \"{$this->getUrl()}\" | tar -xzv");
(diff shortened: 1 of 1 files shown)


FoksVHox:
Resolves #3161 as well

Boy132:
Can confirm that this fixes my problem (#3159 )

gOOvER:
works like charm :)

gOOvER (approved):
This fix work for me an selfupgrade Runs without errors

matthewpi (approved):

matthewpi:
Should be noted that this doesn't resolve #3161

PR #3190, opened 2021-03-15
Fix forge universal jar renaming 

trenutoo:
The only change is to add a wildcard for the renaming command to match both normal forge-version.jar and forge-version-universal.jar files. This resolves the issue of failing to rename versions that would have universal in the name.

`FORGE_JAR=forge-${FORGE_VERSION}*.jar` which is the same change as in [Parker's repository](https://github.com/parkervcp/eggs/pull/1033).


parkervcp (approved):
LGTM

PR #3200, opened 2021-03-19
Add a message for users to apply different versions

Martmists-GH:
Many users may seem confused at why changing a version does not apply properly, so I added a line that tells them to reinstall to apply the version change.


DaneEveritt (approved):

Martmists-GH:
I just realized I forgot the other Minecraft-related eggs, and there's also the other games which I have not tested.

One problem with the other eggs, such as Fabric, is that they have 2 or more version fields that all require a restart. How should these get the notification?

trenutoo:
It is the same process for all Pterodactyl eggs when making any changes to branches or versions. The variable description doesn't really provide any notifications

PR #3220, opened 2021-03-24
Increase Rust server name length

trenutoo:
The current length of 40 is too short


Daxorinator:
+1 Rust Server names often have tags in them so they can be quite long.

PR #3243, opened 2021-04-01
Update egg-ark--survival-evolved.json

wingdings255:
Updated the list of available ark maps to include the new CrystalIsles map

### DIFF
--- a/database/Seeders/eggs/source-engine/egg-ark--survival-evolved.json
+++ b/database/Seeders/eggs/source-engine/egg-ark--survival-evolved.json
@@ -47,7 +47,7 @@
         },
         {
             "name": "Server Map",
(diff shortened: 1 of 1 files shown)


parkervcp:
This is more informational and not really necessary

wingdings255:
> 
> 
> This is more informational and not really necessary

I just noticed it when I wanted to switch my server to that map and realized no one updated it since that map was released. I figured I would make a pull request to update it

parkervcp:
> > This is more informational and not really necessary
> 
> I just noticed it when I wanted to switch my server to that map and realized no one updated it since that map was released. I figured I would make a pull request to update it

As it's a map that ships with ARK it's reasonable. This also changes if someone uploads a custom map.

parkervcp (approved):
LGTM

PR #3244, opened 2021-04-02
More descriptive error for when backup limit is 0

trenutoo:
At this moment, the error message is confusing and leads to a lot of questions on how to configure backups to work with Pterodactyl

### DIFF
--- a/resources/scripts/components/server/backups/BackupContainer.tsx
+++ b/resources/scripts/components/server/backups/BackupContainer.tsx
@@ -60,7 +60,7 @@ const BackupContainer = () => {
             </Pagination>
             {backupLimit === 0 &&
             <p css={tw`text-center text-sm text-neutral-300`}>
-                Backups cannot be created for this server when the backup limit is set to 0.
+                Backups cannot be created for this server because the backup limit is set to 0.
             </p>
             }
             <Can action={'backup.create'}>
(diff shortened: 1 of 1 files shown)


FoksVHox:
Love the fact that you basically just changed a string, and it broke the test suite

https://github.com/pterodactyl/panel/pull/3244/checks?check_run_id=2255132997#step:11:351

DaneEveritt:
The test suite being broken has nothing to do with this PR, it is broken on the main branches from other stuff.

resources/scripts/components/server/backups/BackupContainer.tsx
```diff
@@ -60,7 +60,7 @@ const BackupContainer = () => {
             </Pagination>
             {backupLimit === 0 &&
             <p css={tw`text-center text-sm text-neutral-300`}>
-                Backups cannot be created for this server.
+                Backups cannot be created for this server when the backup limit is set to 0.
```
TASelwyn: Wouldn't it make more sense to do
"because the backup limit is set to 0."

broimsuperman: @TASelwyn how so? imo it's alot easier for people to understand.

matthewpi: I also prefer the use of "because" rather than "when" in this context as it is more clear as the limit being 0 is the *cause* of not being able to create backups.

TASelwyn: > @TASelwyn how so? imo it's alot easier for people to understand.

Based on the context, "because" makes more grammarical sense and more clearly defines what is the cause.

PR #3246, opened 2021-04-03
Change to get database host from location

mundotv789123:
Getting location of database host from associated node.
I think this can help to avoid problems where the database is created in a different location making it difficult to connect the server to the database.

### DIFF
--- a/app/Services/Databases/DeployServerDatabaseService.php
+++ b/app/Services/Databases/DeployServerDatabaseService.php
@@ -6,6 +6,7 @@ use Webmozart\Assert\Assert;
 use Pterodactyl\Models\Server;
 use Pterodactyl\Models\Database;
 use Pterodactyl\Models\DatabaseHost;
+use Pterodactyl\Models\Node;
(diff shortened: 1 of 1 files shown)


matthewpi:
This seems very inefficient and the wrong way of approaching this problem.  The better option would be to actually add a `database_host_id` column to the location and select it that way rather than going through all of the nodes in a location to find a database host.  However, this problem has already been resolved on react-admin when I reversed the current relation so you can link a node to a database host, rather than a database host to a node, allowing multiple nodes to be linked to a single database host.

DaneEveritt:
Well aside from the original logic being hilariously inefficient at scale (lol, thanks me), this approach doesn't seem right to me either. I'm honestly not entirely sure what exactly you're solving for here (which is why a clear description on PRs is _vital_) since I can't keep track of every moving part in this project.

But either way, this is a pretty drastic change to how the logic works and we've never really gotten complaints about things like this in the past. For that reason I'm going to close this PR since it isn't associated with any known issues that have gotten significant attention over the last months.

Issue #3268, opened 2021-04-16
Add server description to server info route


NoodlesDev987:
**Is your feature request related to a problem? Please describe.**
I don't like that server description is not visible to users also when they are on server page (`/server/id` route)

**Describe the solution you'd like**
I think that server description should be visible also on `/server/id` route.

**Describe alternatives you've considered**
If you think that it will not look good there, it can be in Settings, in DEBUG INFORMATION field.

**Additional context**
Now the description is only visible on the main page (`/` route):
![image](https://user-images.githubusercontent.com/81709846/114984680-45e54800-9e92-11eb-8ce6-29e913092209.png)

I have some idea where the description should be on `/server/id` route:
![image](https://user-images.githubusercontent.com/81709846/114985144-c015cc80-9e92-11eb-8ac6-abb9b733a7ab.png)

alexgn100:
Hey NoodlesDev987,

would this be your idea?

![msedge_66acfbMwyQ](https://user-images.githubusercontent.com/32128537/114993344-03286d80-9e9c-11eb-88a2-0e35234a25ee.png)

NoodlesDev987:
@MG-100 Yes, something like that.

I think that also this should be added  to `/admin/servers/view/internal-id` route below the server name field
![image](https://user-images.githubusercontent.com/81709846/115004839-88fde600-9ea7-11eb-8a9e-721535b7275d.png)

alexgn100:
> @MG-100 Yes, something like that.
> 
> I think that also this should be added to `/admin/servers/view/internal-id` route below the server name field
> ![image](https://user-images.githubusercontent.com/81709846/115004839-88fde600-9ea7-11eb-8a9e-721535b7275d.png)

Mean you this?

![msedge_0C7CkWh66h](https://user-images.githubusercontent.com/32128537/115007249-12161c80-9eaa-11eb-8382-3daa211b0102.png)

and when nothing is set:

![msedge_Tql5RmJI0v](https://user-images.githubusercontent.com/32128537/115007284-193d2a80-9eaa-11eb-8937-f3e624681051.png)

NoodlesDev987:
Yes I mean that

alexgn100:
Okay.
The Pull Request is open we must waiting for the Pterodactyl Team to approved these changes.

DaneEveritt:
I don't see any reason to show a server's description in the sidebar like that. Also, if the description is any significant length the entire sidebar is going to be flooded with text.

[...]

PR #3269, opened 2021-04-16
Add Description Field

alexgn100:
Add a Description Field like #3268


DaneEveritt:
I don't see any reason to show a server's description in the sidebar like that. Also, if the description is any significant length the entire sidebar is going to be flooded with text.

NoodlesDev987:
Now clients can't see the description, please add it atleast to `/server/id`
If the description is longer, write there ... to provent from being flooded with text

NoodlesDev987 (commented):
Looks good

PR #3279, opened 2021-04-20
Add group input to upgrade command

Boy132:
This allows the user to also enter a group that is used for _chown_. Usefull if your user isn't the same as your group.


Rhydium:
You practically didn't change anything as the command being runned is `chown -R {$user}:{$user}`, so when you choose `www-data` the command being executed would already be `chown -R www-data:www-data`.

Boy132:
> You practically didn't change anything as the command being runned is `chown -R {$user}:{$user}`, so when you choose `www-data` the command being executed would already be `chown -R www-data:www-data`.

I changed the command to be `chown -R {$user}:{$group}`. This allows the group to be different from the user which is the whole point of the PR.

DaneEveritt:
Well if you're going to do that it should accept a `--group` flag as well, because otherwise you cannot actually set a different user and group unless it is already configured that way on the system.

Also, any variable that is more than a word should be `camelCase` not `snake_case`. :)

Boy132:
> Well if you're going to do that it should accept a `--group` flag as well, because otherwise you cannot actually set a different user and group unless it is already configured that way on the system.
> 
> Also, any variable that is more than a word should be `camelCase` not `snake_case`. :)

Made some changes. :) Should be good now.

Kaesewuerfel (approved):

PR #3280, opened 2021-04-20
Drop explicit transaction from store_node_tokens_as_encrypted_value

### LINKED ISSUE #3229: Migration store_node_tokens_as_encrypted_value fails: savepoint does not exist
* Panel or Daemon: panel
* Version of Panel/Daemon: 1.3.1 (`ghcr.io/pterodactyl/panel:v1.3.1`)
* Server's OS: Linux
* Your Computer's OS & Browser: n/a

**Describe the bug**

As noted in #1959, the `2020_04_10_141024_store_node_tokens_as_encrypted_value` migration can sometimes fail with an error `SQLSTATE[42000]: Syntax error or access violation: 1305 SAVEPOINT trans2 does not exist`. However, that bug was closed without any resolution.

I've encountered the `SAVEPOINT trans2 does not exist` issue again, and believe this is caused by attempting to use an explicit transaction while inside the implicit transaction of a migration that does DDL, because DDL statements cause implicit commit.

To illustrate the issue, I can reproduce a similar problem manually. Using an empty database on MariaDB 10.3:

```sql
> CREATE TABLE table1 (id int);
> BEGIN;
> ALTER TABLE table1 ADD COLUMN (value int);
> SAVEPOINT trans2;
> INSERT INTO table1(id, value) VALUES (1, 1);
> RELEASE SAVEPOINT trans2;
ERROR 1305 (42000): SAVEPOINT trans2 does not exist
```

This follows what the migration in question does, where it is wrapped in a transaction implicitly, does some DDL, then creates a savepoint that silently fails to be created because [executing DDL caused an implicit commit](https://mariadb.com/kb/en/sql-statements-that-cause-an-implicit-commit/). Attempting to release that savepoint then causes the visible error. Once it's in this state, attempting to roll back the migration also fails because it was recorded as complete but never finished, so the rollback fails to drop the `nodes_daemonsecret_unique` index because it was never created.

The issue could be fixed in a few ways:
 * Remove the explicit `DB::Transaction` from the migration. Minimal change but still has issues with being impossible to roll back in case of error while running, due to implicit commit.
 * Split into two or three migrations: one for the initial DDL, one to run the `UPDATE`s, and optionally one to do the index updates (which could be part of the update one, but then still has potentially problematic implicit commits).

As for why this isn't a common error that people see, I suspect it may be down to the database engine or settings often being such that this error isn't reported- possibly related to `STRICT_TRANS_TABLES` mode or lack of support for transactions in the current engine (MyISAM..).

tari:
Migrations are executed in transactions anyway, and creating a savepoint can cause
spurious failures on databases that don't support transactional DDL (like
MySQL and MariaDB) when it attempts to commit a savepoint that was silently
not created because there wasn't an active transaction after some DDL was
executed.

While a better solution might involve splitting this migration into several so each
one is only DDL or only data manipulation, I don't think that can be done very
easily while maintaining compatibility with existing deployments.

Fixes #3229.

### DIFF
--- a/database/migrations/2020_04_10_141024_store_node_tokens_as_encrypted_value.php
+++ b/database/migrations/2020_04_10_141024_store_node_tokens_as_encrypted_value.php
@@ -33,19 +33,17 @@ class StoreNodeTokensAsEncryptedValue extends Migration
             $table->text('daemon_token')->change();
         });
 
-        DB::transaction(function () {
-            /** @var \Illuminate\Contracts\Encryption\Encrypter $encrypter */
-            $encrypter = Container::getInstance()->make(Encrypter::class);
+        /** @var \Illuminate\Contracts\Encryption\Encrypter $encrypter */
+        $encrypter = Container::getInstance()->make(Encrypter::class);
 
-            foreach (DB::select('SELECT id, daemon_token FROM nodes') as $datum) {
-                DB::update('UPDATE nodes SET uuid = ?, daemon_token_id = ?, daemon_token = ? WHERE id = ?', [
-                    Uuid::uuid4()->toString(),
-                    substr($datum->daemon_token, 0, 16),
-                    $encrypter->encrypt(substr($datum->daemon_token, 16)),
-                    $datum->id,
-                ]);
-            }
-        });
+        foreach (DB::select('SELECT id, daemon_token FROM nodes') as $datum) {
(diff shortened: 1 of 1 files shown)


DaneEveritt:
This might  not matter since it probably affects other migrations too, but do you know if this will cause the dry-run for migrations to accidentally execute this logic? I've run into that in the past but I'm not sure how I avoided it.

tari:
`migrate --pretend` seems to avoid running any actual queries against the database, so I don't think this change will affect that.

PR #3285, opened 2021-04-21
Update forge download URL to match the new maven host

trenutoo:
Forge download URL has migrated from files.minecraftforge.net to maven.minecraftforge.net


parkervcp (approved):
LGTM

PR #3292, opened 2021-04-24
Make it possible to manually run disable schedule

### LINKED ISSUE #3291: Allow schedules tasks to be manually executed despite being disabled
I'm frustrated that I can't manually run scheduled tasks when the status is disabled. I find myself using often to run a set of task like this:
image
And would love to be able to run this without attaching it to a "cron" 

Today I'm doing the following steps:
- enable the schedule
- click the run now button
- disable the schedule.

A quick workaround would be to allow users to run disabled schedule jobs, The ideal would probably be than the schedule is renamed "Jobs" or something like this and that the schedule is an optional parameter of the job.

JulienTant:
fixes #3291

I wanted to find a way to execute Jobs that consists in multiple tasks. Create a disabled schedule and give the possibility to execute it is the easiest way to get there.

### DIFF
--- a/app/Jobs/Schedule/RunTaskJob.php
+++ b/app/Jobs/Schedule/RunTaskJob.php
@@ -30,16 +30,16 @@ class RunTaskJob extends Job implements ShouldQueue
     /**
      * @var bool
      */
-    public $now;
+    public $manualRun;
 
     /**
      * RunTaskJob constructor.
      */
-    public function __construct(Task $task, $now = false)
+    public function __construct(Task $task, $manualRun = false)
     {
         $this->queue = config('pterodactyl.queues.standard');
         $this->task = $task;
-        $this->now = $now;
+        $this->manualRun = $manualRun;
     }
 
     /**
@@ -52,8 +52,8 @@ class RunTaskJob extends Job implements ShouldQueue
         InitiateBackupService $backupService,
         DaemonPowerRepository $powerRepository
     ) {
-        // Do not process a task that is not set to active, unless it's been trigger by the run now API.
-        if ($this->task->schedule->is_active && !$this->now) {
+        // Do not process a task that is not set to active, unless it's been manually triggered.
+        if (!$this->task->schedule->is_active && !$this->manualRun) {
             $this->markTaskNotQueued();
             $this->markScheduleComplete();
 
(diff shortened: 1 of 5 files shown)


app/Jobs/Schedule/RunTaskJob.php
```diff
@@ -46,8 +52,8 @@ public function handle(
         InitiateBackupService $backupService,
         DaemonPowerRepository $powerRepository
     ) {
-        // Do not process a task that is not set to active.
-        if (!$this->task->schedule->is_active) {
+        // Do not process a task that is not set to active, unless it's been trigger by the run now API.
+        if ($this->task->schedule->is_active && !$this->now) {
```
DaneEveritt: I do not follow the reason for changing the logic right here. Shouldn't this be asking if the schedule is not active and `$this->now` is not set?

Otherwise it looks like this is just going to explode when a task runs because it is set to active but `$this->now` is false for those tasks?

app/Jobs/Schedule/RunTaskJob.php
```diff
@@ -27,13 +27,19 @@ class RunTaskJob extends Job implements ShouldQueue
...
+
     /**
      * RunTaskJob constructor.
      */
-    public function __construct(Task $task)
+    public function __construct(Task $task, $now = false)
     {
         $this->queue = config('pterodactyl.queues.standard');
         $this->task = $task;
+        $this->now = $now;
```
DaneEveritt: This is a pretty confusing variable name from trying to parse the logic here. This can probably be renamed to something like `$runNow` or `$manualRun` might be even better since that seems to be what it actually is.

DaneEveritt (changes requested):

DaneEveritt:
The tests were also broken by this change, you'll need to account for how to prevent a schedule from running if it is not currently marked as active and somehow gets into the pipeline. Or document why that is a scenario that doesn't need to be guarded against in the code and remove the test.

JulienTant:
Hi Dane, thanks for the feedback :) I mostly wanted to make sure that feature was ok with you guys. I'll work on making this PR better now! thanks.

JulienTant:
About the test removal from commit b10f618:
The test was making sure that manually running a disabled job was not possible through the API. The point of this PR is to make that possible.

When the jobs are ran automatically (using the worker/command), the job don't receive the optional `$now` parameter.
https://github.com/pterodactyl/panel/blob/c449ca5155b135329237fe2e03cf94af24b6683b/app/Console/Commands/Schedule/ProcessRunnableCommand.php#L66 
So the job will run with the parameter `manualRun` at `false`, and the check on the `->is_active` property of the model will be done.

app/Jobs/Schedule/RunTaskJob.php
JulienTant: there was an accidental '!' removal :P This is fixed

DaneEveritt (approved):
Looks great, thanks for making this happen!

Issue #2025, opened 2020-05-13
Add PostgreSQL support


GalacticLion7:
Please add PostgreSQL support in the panel for consistency with other projects.

PostgreSQL could be natively implemented into the panel, or using the [mysql_fdw](https://github.com/EnterpriseDB/mysql_fdw) wrapper to convert MySQL queries into PostgreSQL queries.

matthewpi:
Laravel already supports Postgres, but some migrations and queries that are in the panel don't.  I'd love for Postgres to be supported, the only issue I see is the potential issues with changing older migrations, but I believe that should be mostly fine.

DaneEveritt:
Unless someone opens a PR to add this support, I will not be implementing it. I'm closing this issue in that case, however feel free to reference it in the PR.

Weilbyte:
Hi @DaneEveritt, I am currently working on supporting Postgres, can this issue be reopened and assigned to me? Thanks!

PR will be opened soon, just need to do some testing on my own. But so far, every integration test passes.

ririko5834:
Will be PostgreSQL support added also to database hosts and that users can pick databse type when creating database?

PR #3305, opened 2021-04-27
PostgreSQL Support

Weilbyte:
## PostgreSQL Support
The goal of this PR is to add Postgres support to Pterodactyl panel. Related issue: #2025.

#### Migrations
At first I had attempted to modify the MySQL migrations to work for Postgres but that proved to be beyond my abilities. Migrations are now separated by driver (`database/migrations/[mysql,pgsql]`), and the Postgres migration is a SQL file of a converted database (dirty but it seems to work). 

#### Postgres specific queries
Quite a lot of places in the code use MySQL-exclusive queries (for example, if they involve `INET_ATON()`), this has been fixed by adding a Postgres version of that query and using them depending on the driver loaded, like so:

```php
if (DB::getDriverName() === 'pgsql') {
    // Postgres query
} else {
    // MySQL query
}
```

~~It works, but I think the statement is too long/repetitive and should maybe be replaced by something that would look like `if (isPostgres()) { ...`. I have also thought about using a ternary operator in the queries themselves to completely remove the `if` but I am unsure how clean that would look. For now this works.~~  
  (`config(..)` has since been replaced with `DB::getDriverName()`).

#### Tests
Currently all the integration tests pass.   
     
Some tests had to endure some changes (such as when testing by querying ids that should not exist: `nil` was replaced by `-1` because Postgres does not allow working on a serial field with `nil`).    
     
A copy of `phpunit.xml` was created (`phpunit.pgsql.xml`) specifically for Postgres.
    
Additionally, for the Postgres connections under `config/database.php`, the `Time::getMySQLTimezoneOffset` function for the `timezone` field has been removed and now just uses `APP_TIMEZONE`. ~~Probably not the best but for now it works, will work on a similar function but for Postgres later on~~. 
    
#### GitHub Test workflow
The test workflow now tests Postgres as well.    

Additionally, I also fixed the if expressions so that only the necessary integration tests are ran (previously, for example, the mariadb runs ran integration tests for both mysql, and mariadb because of an issue in the expressions). Unsure whether I should include this as its own PR or not. 

### Ready for review.


sergejostir:
You can get the driver name using https://laravel.com/api/8.x/Illuminate/Database/Connection.html#method_getDriverName

app/Http/Controllers/Admin/Nodes/NodeViewController.php
```diff
@@ -118,13 +119,23 @@ public function allocations(Request $request, Node $node)
...
-                ->orderByRaw('INET_ATON(ip) ASC')
-                ->get(['ip']),
-        ]);
+        if (DB::getDriverName() === 'pgsql') {
+            return $this->view->make('admin.nodes.view.allocation', [
+                'node' => $node,
+                'allocations' => Allocation::query()->where('node_id', $node->id)
+                    ->groupBy('ip')
+                    ->orderByRaw('ip ASC')
+                    ->get(['ip']),
```
VottusCode: This can be simplified by making a simple ternary: 
```php
->orderByRaw(DB::getDriverName() === 'pgsql' ? 'ip ASC' : 'INET_ATON(ip) ASC')
```
If you do that, you can remove the if statement and you won't have to repeat the same code twice.

app/Models/Filters/MultiFieldServerFilter.php
```diff
@@ -58,13 +59,24 @@ function (Builder $builder) use ($value) {
...
-                    ->orWhereRaw('LOWER(servers.name) LIKE ?', ["%$value%"]);
-            });
+        if (DB::getDriverName() === 'pgsql') {
+            $query
+                ->where(function (Builder $builder) use ($value) {
+                    $builder->where('servers.uuid', $value)
+                        ->orWhere('servers.uuid', 'LIKE', "$value%")
+                        ->orWhere('servers.uuidShort', $value)
+                        ->orWhere('servers.external_id', $value)
+                        ->orWhereRaw('servers.name ILIKE ?', ["%$value%"]);
```
VottusCode: Again, a ternary can be used to simplify this code and follow DRY.

app/Repositories/Eloquent/AllocationRepository.php
```diff
@@ -42,17 +43,29 @@ public function getUnassignedAllocationIds(int $node): array
      */
     protected function getDiscardableDedicatedAllocations(array $nodes = []): array
     {
-        $query = Allocation::query()->selectRaw('CONCAT_WS("-", node_id, ip) as result');
+        if (DB::getDriverName() === 'pgsql') {
```
VottusCode, line 46: Either I'm blind, or the query is exactly the same.

app/Repositories/Eloquent/AllocationRepository.php
```diff
@@ -42,17 +43,29 @@ public function getUnassignedAllocationIds(int $node): array
...
             $query->whereIn('node_id', $nodes);
         }
 
-        return $query->whereNotNull('server_id')
-            ->groupByRaw('CONCAT(node_id, ip)')
-            ->get()
-            ->pluck('result')
-            ->toArray();
+        if (DB::getDriverName() === 'pgsql') {
+            return $query->whereNotNull('server_id')
```
VottusCode: Ternary.

app/Repositories/Eloquent/NodeRepository.php
```diff
@@ -100,15 +115,27 @@ public function loadLocationAndServerCount(Node $node, bool $refresh = false): N
...
-                ->orderBy('port', 'asc')
-                ->with('server:id,name')
-                ->paginate(50)
-        );
+        if (DB::getDriverName() === 'pgsql') {
+            $node->setRelation(
+                'allocations',
+                $node->allocations()
+                    ->orderByRaw('server_id IS NOT NULL DESC, server_id IS NULL')
+                    ->orderByRaw('ip ASC')
```
VottusCode: Again, ternary to simplify the code.

VottusCode (changes requested):
Some minor changes that could simplify some of the queries a bit, but otherwise LGTM.

app/Repositories/Eloquent/AllocationRepository.php
Weilbyte: The quotations on the `CONCAT_WS` separator are different.

app/Http/Controllers/Admin/Nodes/NodeViewController.php
Weilbyte: Thanks, will change to ternary where possible.

app/Repositories/Eloquent/AllocationRepository.php
VottusCode: My bad.

Weilbyte:
PR almost ready ~, will be thinking of adding a `Time::getMySQLTimezoneOffset` alternative for Postgres.~
(Apparently no need because Postgres handles named time zones better than MySQL)

leon0399:
this would be so nice! Since I use Laravel Forge for my own personal projects, I wanted to install Panel to my Forge server, and then faced an issues with database :disappointed:

broimsuperman:
@DaneEveritt have you considered adding this?

parkervcp:
> DaneEveritt have you considered adding this?

If it's open it's considered. It's also a PR and will get reviewed when it can be.

app/Http/Controllers/Admin/Nodes/NodeViewController.php
```diff
@@ -122,7 +123,7 @@ public function allocations(Request $request, Node $node)
             'node' => $node,
             'allocations' => Allocation::query()->where('node_id', $node->id)
                 ->groupBy('ip')
-                ->orderByRaw('INET_ATON(ip) ASC')
+                ->orderByRaw(DB::getDriverName() === 'pgsql' ? 'ip ASC' : 'INET_ATON(ip) ASC')
```
FoksVHox, line 126: As Laravel supports PostgreSQL, out-of-the-box this shouldn't be necessary, as Eloquent and the query builder, knows how to form it.

app/Providers/AppServiceProvider.php
```diff
@@ -24,6 +26,9 @@ public function boot()
     {
         Schema::defaultStringLength(191);
 
+        $driver = DB::getDriverName();
+        $this->loadMigrationsFrom(database_path("migrations/$driver"));
+
```
FoksVHox, line 31: As Laravel supports PostgreSQL, out-of-the-box this shouldn't be necessary, as Eloquent and the query builder know how to form it.

database/migrations/pgsql/initial.sql
```diff
@@ -0,0 +1,632 @@
...
+
+-- public.subusers foreign keys
+
+ALTER TABLE public.subusers ADD CONSTRAINT subusers_server_id_foreign FOREIGN KEY (server_id) REFERENCES public.servers(id) ON DELETE CASCADE ON UPDATE RESTRICT;
+ALTER TABLE public.subusers ADD CONSTRAINT subusers_user_id_foreign FOREIGN KEY (user_id) REFERENCES public.users(id) ON DELETE CASCADE ON UPDATE RESTRICT;
+
+
+-- public.tasks foreign keys
+
+ALTER TABLE public.tasks ADD CONSTRAINT tasks_schedule_id_foreign FOREIGN KEY (schedule_id) REFERENCES public.schedules(id) ON DELETE CASCADE ON UPDATE RESTRICT;
```
FoksVHox, line 632: As Laravel supports PostgreSQL, out-of-the-box this shouldn't be necessary, as Eloquent and the query builder, knows how to form it. The migrations should work just fine with Laravel & PostgreSQL.

tests/Browser/BrowserTestCase.php
```diff
@@ -53,7 +53,7 @@ protected function setUp(): void
     {
         // Don't accidentally run the migrations aganist the non-testing database. Ask me
         // how many times I've accidentally dropped my database...
-        if (env('DB_CONNECTION') !== 'testing') {
+        if (str_starts_with(env('DB_CONNECTION'), 'testing')) {
```
FoksVHox, line 56: Unneeded change.

database/migrations/pgsql/2021_04_23_025321_build_initial_database.php
```diff
@@ -0,0 +1,56 @@
...
+				backups,
+				databases,
+				schedules,
+				server_transfers,
+				server_variables,
+				servers,
+				subusers
+		CASCADE');
+	}
+}
```
FoksVHox, line 56: As Laravel supports PostgreSQL, out-of-the-box this shouldn't be necessary, as Eloquent and the query builder, knows how to form it. The migrations should work just fine with Laravel & PostgreSQL.

FoksVHox (commented):
As Laravel supports PostgreSQL, out-of-the-box this shouldn't be necessary, as Eloquent and the query builder, knows how to form it. The migrations should work just fine with Laravel & PostgreSQL, so there isn't a need to have `migrations/<driver>/`.

app/Http/Controllers/Admin/Nodes/NodeViewController.php
sergejostir: Laravel supporting Postgre doesn't mean that every function of every database engine is implemented in Eloquent.

Weilbyte: Eloquent and the query builder do not know how to form it, as sergiz has pointed out.

tests/Browser/BrowserTestCase.php
Weilbyte: MySQL test connection is called `testing` and Postgres testing connection is called `testing-pgsql`, so I am pretty sure it is a necessary change so this bit of the code works for both the MySQL and PGSQL testing connections, and not just the MySQL one.

Weilbyte:
@FoksVHox that is not the case, through all my tests and attempts, the MySQL migrations have not worked for Postgres because they include either data types that Postgres does not have (which Eloquent could possibly convert) and also custom SQL queries (which sometimes don't seem to work for Postgres because they use MySQL-exclusive syntax/etc).

If it could just work on its own as you say it can, then that'd be great but it doesn't.

FoksVHox:
@Weilbyte I can see it being an issue with the raw statements that Pterodactyl includes. However, Eloquent's migrations should work just fine with PostgreSQL, as stated in the [Laravel documentation](https://laravel.com/docs/8.x/database#introduction). In most cases, it should just be a case of changing the values in the `.env` file.

DaneEveritt:
Due to the extreme size of this PR, the difficulty in reviewing it with my limited time, and that it does not significantly change anything with the Panel right now I am going to be closing it out.

I appreciate the effort you put into it, and I am sorry that it isn't going to be merged per the new PR policy: https://github.com/pterodactyl/panel/blob/develop/CONTRIBUTING.md

Issue #2619, opened 2020-10-27
CPU usage readability


VibeGAMESNL:
**Background (please complete the following information):**
* Panel or Daemon: Panel
* Version of Panel/Daemon: 1.0.3
* Server's OS: Ubuntu 20.03
* Your Computer's OS & Browser: Chrome

**Describe the bug**
The CPU usage can be higher then 100%. I understand that 1 core = 100% and 2 cores are 200% but customers don't understand this.

I understand that this isn't really a bug but more like a change request.

![image](https://user-images.githubusercontent.com/66668585/97324821-916b5180-1872-11eb-9d8e-b66ef5872474.png)

**To Reproduce**
Create a load that will use more then 1 core.

**Expected behavior**
CPU usage should be devided by the number of cores I gues? So it will be better readable for a customer.

parkervcp:
100% == 1 core is a linux standard. It's not really something we control. This was like this in 0.7 as well. This isn't a new thing to 1.0.

VibeGAMESNL:
> 100% == 1 core is a linux standard. It's not really something we control. This was like this in 0.7 as well. This isn't a new thing to 1.0.

Yes it's a linux standard but not everybody knows this. A maximum of 100% is better to understand for most people.
I understand if you don't want to change this. Just a suggestion.

DaneEveritt:
I won't be changing this, we've previously capped things at 100% and it just caused even more confusion to people, and was not intuitive in the long run.

schrej:
How about adding a `/ x00%` so it's clear what the limit is?

Edit regarding the marking as spam: If you like an idea, like the comment. Anything that's basically a `+1` (which includes "awesome idea") without providing extra content, e.g. an good argument why it's a good solution, is considered spam on GitHub.

VibeGAMESNL:
> How about adding a `/ x00%` so it's clear what the limit is?

Awesome idea? @DaneEveritt 👍:D

PR #3307, opened 2021-04-28
Show CPU limit next to CPU usage

trenutoo:
Shows the server's cpu limit next to the CPU usage, together with limiting CPU usage to 2 decimals on the dashboard to prevent 105.53564% CPU usage. 

This should help with confusion regarding "why I have 160% CPU usage", ex #2619

image

image

image

### DIFF
--- a/resources/scripts/components/server/ServerDetailsBlock.tsx
+++ b/resources/scripts/components/server/ServerDetailsBlock.tsx
@@ -74,7 +74,7 @@ const ServerDetailsBlock = () => {
 
     const diskLimit = limits.disk ? megabytesToHuman(limits.disk) : 'Unlimited';
     const memoryLimit = limits.memory ? megabytesToHuman(limits.memory) : 'Unlimited';
-    const cpuLimit = limits.cpu !== 0 ? limits.cpu + '%' : 'Unlimited';
+    const cpuLimit = limits.cpu ? limits.cpu + '%' : 'Unlimited';
 
     return (
         <TitledGreyBox css={tw`break-words`} title={name} icon={faServer}>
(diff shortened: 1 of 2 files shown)


resources/scripts/components/server/ServerDetailsBlock.tsx
```diff
@@ -74,6 +74,7 @@ const ServerDetailsBlock = () => {
 
     const diskLimit = limits.disk ? megabytesToHuman(limits.disk) : 'Unlimited';
     const memoryLimit = limits.memory ? megabytesToHuman(limits.memory) : 'Unlimited';
+    const cpuLimit = limits.cpu !== 0 ? limits.cpu + '%' : 'Unlimited';
```
schrej: Maybe do this one like the others above? This would error if `limits.cpu` is undefined.

trenutoo: That was some kind of nosleep brainfart. Thanks for pointing it out.

Kaesewuerfel (dismissed):

DaneEveritt:
@Kaesewuerfel I appreciate the desire to be involved in the process, but please refrain from reviewing any of the PRs. We do not use community reviews for any purpose when evaluating these PRs and they only cause notification spam for all of the maintainers. Thanks.

Boy132:
@Software-Noob Might be also useful to add the limit to the ServerRow.tsx so it's visible from the server list view. (Also for the sake of completeness)

trenutoo:
> @Software-Noob Might be also useful to add the limit to the ServerRow.tsx so it's visible from the server list view. (Also for the sake of completeness)

I will commit that next week when I'll be back from my trip. .

PR #3309, opened 2021-04-28
Update SECURITY.md

broimsuperman:
was looking around and thought it should be an @ lol


TrixterTheTux:
Pretty sure this is done on purpose to avoid scraping bots from sending spam

Issue #3277, opened 2021-04-19
php artisan node creation/setup & php artisan allocation creation/setup


JackW6809:
**Is your feature request related to a problem? Please describe:**
I'm always frustrated when I have to go into the panel and create the node and allocations manually.

**Describe the solution you'd like:**
Like for example ```p:location:make``` works great. So why not make a command for node creation and allocation creation? Also while we are there, why not allow the Wings configuration file to be automatically filled in according to what is in the panel rather than having to generate the token, etc?

**Describe alternatives you've considered:**
I have been using the API for a while, however, sometimes it is lengthy. Especially when I have to put in lots of allocations. I think having the command available will make installs quicker and easier. Also more foolproof.

**Additional context:**
If this feature is added, and node creation is all been able to be done via a command. And allocations are able to be specified when creating the node. Then it automatically configuring Wings on the same machine, then it would allow for an entirely headless installation.

Boy132:
You can auto-deploy nodes with one command:
![image](https://user-images.githubusercontent.com/8203120/115255435-8d880000-a12e-11eb-8e54-b328e7af2bea.png)

JackW6809:
Yes, however that is not the point I am attempting to make. I am trying to make it a headless installation. The Auto-Deploy command just configures Wings after the Node is created in the panel. Whereas I would be looking to create the node, deploy wings and set allocations using either a command or several commands.

FoksVHox:
I'm most likely going to try to make this tomorrow

matthewpi:
Is there a reason you don't use the Application API to achieve this?

FoksVHox:
It could be faster to create/deploy new nodes?

PR #3319, opened 2021-05-03
Ability to create nodes with artisan

Fabian-Ser:
This is a public release which (partially) solves issue #3277, these changes allow a node to be created via artisan (cli).


app/Console/Commands/Node/MakeNodeCommand.php
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * @var string
+     */
+    protected $description = 'Creates a new node on the system via the CLI.';
+
+    /**
+     * Create a new command instance.
+     */
+    public function __construct(NodeCreationService $creationService)
```
DaneEveritt: Don't inject into the constructor, this is executed whenever any artisan command is run, not just when this command runs. You want the injection in the `handle()` command instead which is only invoked on run.

app/Console/Commands/Node/MakeNodeCommand.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Handle the command execution process.
+     *
+     * @throws \Pterodactyl\Exceptions\Model\DataValidationException
+     */
+    public function handle()
+    {
+        $name = $this->option('name') ?? $this->ask(trans('command/messages.node.ask_name'));
```
DaneEveritt: I'd recommend doing something like `$data = [];` and then `$data['name'] = $this->option() ?? $this->ask();`

This avoids the compact call, and a bunch of variables that don't meet the code style requirements of the repo.

app/Console/Commands/Node/MakeNodeCommand.php
```diff
@@ -0,0 +1,102 @@
...
+        $description = $this->option('description') ?? $this->ask(trans('command/messages.node.ask_description'));
```
DaneEveritt: You can remove all of the translation stuff for this, that was a bad choice made awhile back but it is pretty pointless for the CLI.

app/Console/Commands/Node/MakeNodeCommand.php
```diff
@@ -0,0 +1,102 @@
...
+        $location_id = $this->option('locationId') ?? $this->ask(trans('command/messages.node.ask_location_id'));
+        $fqdn = $this->option('fqdn') ?? $this->ask(trans('command/messages.node.ask_fqdn'));
+        if (!filter_var(gethostbyname($fqdn), FILTER_VALIDATE_IP)) {
+            $this->error(trans('admin/node.validation.fqdn_not_resolvable'));
+            return;
+        }
+        $public = $this->option('public') ?? $this->ask(trans('command/messages.node.ask_public'));
+        $scheme = $this->option('scheme') ?? $this->ask(trans('command/messages.node.ask_scheme'));
```
DaneEveritt: This can use the pre-defined selectors rather than just being a generic input. Should auto-suggest `http(s)`

app/Console/Commands/Node/MakeNodeCommand.php
```diff
@@ -0,0 +1,102 @@
...
+        if ($scheme !== 'https' && $scheme !== 'http') {
+            $this->error(trans('command/messages.node.no_valid_scheme'));
+            return;
+        }
+        if (filter_var($fqdn, FILTER_VALIDATE_IP) && $scheme === 'https') {
+            $this->error(trans('admin/node.validation.fqdn_required_for_ssl'));
+            return;
+        }
+        $behind_proxy = $this->option('proxy') ?? $this->ask(trans('command/messages.node.ask_behind_proxy'));
```
DaneEveritt: Yes/No rather than ask, same for just about any question similar.

DaneEveritt (changes requested):

Fabian-Ser:
Thanks for the review, I'm just stuck with the language files, wouldn't this become an feature in the future (so that pterodactyl becomes multi-language again just like 0.7)? The rest will be fixed asap.

DaneEveritt:
There is no reason to make the CLI translatable, just a huge waste of time for everyone for something used very infrequently. Hard-coded language strings are completely fine here.

Fabian-Ser:
The code is fixed with the help of @JackW6809

I had to take a little break from this project.

JackW6809:
@Fabian-Ser 

They will not accept it now until V2 anyways. As I kept my code before pulling my PR's out and they have changed the backend variables making this artisan command useless (it errors using the current code when using V2 as the database table for node's has different column names) until we can get our hands on V2 which won't be out for a while yet.

trenutoo:
You don't need the changes in message file as it's not being used.

Your options should be the same as on the Panel. Behind proxy is not for Cloudflare, it's for any proxy that will manage the traffic

Fabian-Ser:
Edited, thanks for the advice.

PR #2416, opened 2020-09-25
Hide Create backup if limit = 0

notAreYouScared:
Don't show the 'Create Backup' option when the limit is 0
Closes #2395

PR #3323, opened 2021-05-03
Don't allow backups to be made via schedules if limit = 0

### LINKED ISSUE #2395: Backups can still be made via schedules
* Panel or Daemon: Panel
* Version of Panel/Daemon: 1.0.0-rc.6
* Server's OS: Ubuntu
* Your Computer's OS & Browser: Mac, Chrome

**Describe the bug**
Creating a schedule on a server that can not have backups allows you to still have backups

**To Reproduce**
1. Go to a server that does not allow backups
2. Click on schedule
3. Click on new task
4. Click on action, then new backup
5. You can still have it as part of the schedule

**Expected behavior**
It does not appear or it will show a error message

Screen Shot 2020-09-22 at 16 42 28
Screen Shot 2020-09-22 at 16 42 40

notAreYouScared:
Closes https://github.com/pterodactyl/panel/issues/2395

Original PR https://github.com/pterodactyl/panel/pull/2416

~~I have no idea how to protect the endpoint to prevent the task from being made via the API :(~~

API is protected 👍 

Also, fixes the title when directly navigating to the details of a specific schedule. ( panel.com/server/<uuid>/schedules/# )


Kaesewuerfel (dismissed):

nikrozman21:
This looks like a useful feature to me, but you are probably missing a column in the schedules migration.

notAreYouScared:
> This looks like a useful feature to me, but you are probably missing a column in the schedules migration.

Huh? I'm not adding or editing any migrations? What's missing?

nikrozman21:
> > This looks like a useful feature to me, but you are probably missing a column in the schedules migration.
> 
> Huh? I'm not adding or editing any migrations? What's missing?

Sorry, accidentally posted it on the wrong thread.
Comment hidden.

app/Http/Controllers/Api/Client/Servers/ScheduleTaskController.php
```diff
@@ -49,6 +50,10 @@ public function store(StoreTaskRequest $request, Server $server, Schedule $sched
             throw new ServiceLimitExceededException("Schedules may not have more than {$limit} tasks associated with them. Creating this task would put this schedule over the limit.");
         }
 
+        if ($server->backup_limit === 0 && $request->action === 'backup'){
+            throw new HttpForbiddenException("A backup task cannot be created when the backup limit for a server is set to {$server->backup_limit}.");
```
matthewpi: This should return a bad request response, forbidden is used for bad tokens.  Message should be changed to something like `This server is not able to create any backups` or `A backup task cannot be created because the server has a backup limit of 0` (if you leave the message the same, you can replace `{$server->backup_limit}` with `0`).

app/Http/Controllers/Api/Client/Servers/ScheduleTaskController.php
```diff
@@ -49,6 +50,10 @@ public function store(StoreTaskRequest $request, Server $server, Schedule $sched
             throw new ServiceLimitExceededException("Schedules may not have more than {$limit} tasks associated with them. Creating this task would put this schedule over the limit.");
         }
 
+        if ($server->backup_limit === 0 && $request->action === 'backup'){
```
matthewpi: `){` -> `) {`

app/Http/Controllers/Api/Client/Servers/ScheduleTaskController.php
```diff
@@ -81,6 +87,10 @@ public function update(StoreTaskRequest $request, Server $server, Schedule $sche
             throw new NotFoundHttpException();
         }
 
+        if ($server->backup_limit === 0 && $request->action === 'backup'){
```
matthewpi: `){` -> `) {`

app/Http/Controllers/Api/Client/Servers/ScheduleTaskController.php
```diff
@@ -81,6 +87,10 @@ public function update(StoreTaskRequest $request, Server $server, Schedule $sche
...
+            throw new HttpForbiddenException("A backup task cannot be created when the backup limit for a server is set to {$server->backup_limit}.");
```
matthewpi: This should return a bad request response, forbidden is used for bad tokens.  Message should be changed to something like `This server is not able to create any backups` or `A backup task cannot be created because the server has a backup limit of 0` (if you leave the message the same, you can replace `{$server->backup_limit}` with `0`).

resources/scripts/components/server/schedules/TaskDetailsModal.tsx
```diff
@@ -78,21 +79,27 @@ const TaskDetailsModal = ({ schedule, task }: Props) => {
...
     const submit = (values: Values, { setSubmitting }: FormikHelpers<Values>) => {
         clearFlashes('schedule:task');
-        createOrUpdateScheduleTask(uuid, schedule.id, task?.id, values)
-            .then(task => {
-                let tasks = schedule.tasks.map(t => t.id === task.id ? task : t);
-                if (!schedule.tasks.find(t => t.id === task.id)) {
-                    tasks = [ ...tasks, task ];
-                }
+        if (backupLimit === 0 && values.action === 'backup') {
+            console.error('This server is not allowed to create backups.');
```
matthewpi: You don't need the `console.error` if you are showing a flash to the user.

resources/scripts/components/server/schedules/TaskDetailsModal.tsx
```diff
@@ -78,21 +79,27 @@ const TaskDetailsModal = ({ schedule, task }: Props) => {
...
+            setSubmitting(false);
+            addError({ message: 'This server is not allowed to create backups.', key: 'schedule:task' });
```
matthewpi: This message might be better being `This server is not able to create backups` rather than `not allowed`.

matthewpi (changes requested):

app/Http/Controllers/Api/Client/Servers/ScheduleTaskController.php
notAreYouScared: image
image

resources/scripts/components/server/schedules/TaskDetailsModal.tsx
notAreYouScared: I was just following what was above it... Should it be removed in both places?

app/Http/Controllers/Api/Client/Servers/ScheduleTaskController.php
notAreYouScared: Changed to just 0 as the message would only be shown if the limit === 0. 
as for the wording, it's from Dane...

resources/scripts/components/server/schedules/TaskDetailsModal.tsx
notAreYouScared: Removed.

resources/scripts/components/server/schedules/TaskDetailsModal.tsx
notAreYouScared: Changed

notAreYouScared:
Whelp I broke the schedule creation test, as it fails on making a backup... Guess that means I blocked it correctly :P

I would add 
```PHP

    /**
     * Test that backups can not be tasked when the backup limit is 0
     */
    public function testBackupsCanNotBeTaskedIfLimit0()
    {
        [$user, $server] = $this->generateTestAccount();

        /** @var \Pterodactyl\Models\Schedule $schedule */
        $schedule = Schedule::factory()->create(['server_id' => $server->id]);

        $this->actingAs($user)->postJson($this->link($schedule, '/tasks'), [
            'action' => 'backup',
            'time_offset' => 0,
        ])
            ->assertStatus(Response::HTTP_FORBIDDEN)
            ->assertJsonPath('errors.0.detail', 'A backup task cannot be created when the server\'s backup limit is set to 0.');

        $this->actingAs($user)->postJson($this->link($schedule, '/tasks'), [
            'action' => 'backup',
            'payload' => "file.txt\nfile2.log",
            'time_offset' => 0,
        ])
            ->assertStatus(Response::HTTP_FORBIDDEN)
            ->assertJsonPath('errors.0.detail', 'A backup task cannot be created when the server\'s backup limit is set to 0.');
    }
 ```    
So it would have a test that would pass if backup_limit = 0
I don't know how to set the variable so the backup_limit would be >1

DaneEveritt (approved):

PR #3335, opened 2021-05-07
Add nogui to the Forge server startup command

DusterTheFirst:
Exactly the title. The forge servers installed with the egg fail to start since they attempt to load gui libraries which are missing, since the server is running headless. By adding nogui to the end of the script, it disables the loading of these libraries, allowing forge servers to start.

### DIFF
--- a/database/Seeders/eggs/minecraft/egg-forge-minecraft.json
+++ b/database/Seeders/eggs/minecraft/egg-forge-minecraft.json
@@ -16,7 +16,7 @@
         "quay.io\/pterodactyl\/core:java-11"
     ],
     "file_denylist": [],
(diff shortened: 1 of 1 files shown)


parkervcp:
I feel like this wasn't required before.

trenutoo:
I can't replicate this, the server starts without the -nogui requirement just like all the years before. Are you referring to X11 error when trying to use client mods?

snaildos:
Well, It works without it, but for the best "understanding" it probably the best to add it. :)

DusterTheFirst:
You might be right, it could be a client-side mod that I had accidentally put in the server's mod folder. It is interesting that this error did not crop up when I ran this mod on my local PC, but I guess it must be because my local PC has the gui libraries so it just silently shuts off. `nogui` does seem to force any gui based mod to not even attempt to load themselves, so it might be worth adding.

PR #3339, opened 2021-05-07
Allow sort when requesting server-list (API)

Nice-Hosting:
Adds the option to sort the server-list instead of the default sort-by internalId.
This is usefull when having different games. Ex. sorting by name or description can improve the overview.


DaneEveritt:
I'm not sure I see any reason to sort by a field other than `id` or `name` I guess? Sorting by description sounds like a fantastic way to overburden MySQL with a massive unindexed text field.

DaneEveritt:
Closing, branch deleted and per https://github.com/pterodactyl/panel/blob/develop/CONTRIBUTING.md

Issue #2625, opened 2020-10-30
Allow to find servers by short UUID with the API


Hadimhz:
**Is your feature request related to a problem? Please describe.**
Getting a server by its visible ID requires many requests.
if I want to get a specific server using the ID visible to the user (example: efcecec0) I will have to list all servers and then filter the list. but that requires a lot of requests depending on the number of pages you have.

**Describe the solution you'd like**
can you add a filter to the servers API that will allow us to get a server using the serverID (the one visible to the user example: efcecec0) I've been told this is already a thing for the client API why not introduce it to the Admin one

**Describe alternatives you've considered**
add a filter that we can use when listing all servers. (similar to the one for listing all users)

matthewpi:
Is there a reason you don't just filter by the `uuid` column which always starts with `uuidShort`?
https://github.com/pterodactyl/panel/pull/3340#issuecomment-835837117

trenutoo:
Wasn't this solved by https://github.com/pterodactyl/panel/pull/3340 ?

PR #3340, opened 2021-05-07
Allow to find servers by short UUID (Application API)

markross2:
Allow uuidShort(Identifier) to be used as a filter when listing all servers on the application API
Related to issue #2625


VeeBack (approved):

matthewpi:
Is there a reason you don't just filter by the `uuid` column which always starts with `uuidShort`?

markross2:
> Is there a reason you don't just filter by the `uuid` column which always starts with `uuidShort`?

Nop, not really.
The only reason I made the change was because of the issue I mentioned. I don't know their use case but it is an enhancement anyways (at least for those who need to be highly specific for some reason).

PR #3365, opened 2021-05-17
Create Database Host's with Artisan

JackW6809:
Requires https://github.com/pterodactyl/panel/pull/3396
Created a Database Host php artisan command.


DaneEveritt:
I'm not sure I understand the goal of all the artisan commands? I don't want to make the entire Panel into a CLI tool, that is far from the goal and there are plenty of things that I expect will only be done in the UI.

Could you explain your idea here a little more?

JackW6809:
Yeah, sure.  The main goal of all the Artisan commands is for people who have had experience with the panel already and want to just get the installs done and out of the way. Having these additional commands just allows it to all be configured in one place. Additionally, for those of us who have our own install scripts for our install's it allows us to just run our script and it be done with. I tested it earlier and it has made my life a lot easier. I was able to have Wings fully responsive and online within 2 minutes and 6 seconds on one of my machines with these additional Artisan commands.

I do completely understand your concern however the commands do not have to be publicly shown on the documentation, as long as they are there they can be utilised by people as they so please. Also, the creation service files for many of these already exist in your file structure already. I always suggest for people to learn the install process first using the documentation so they actually have an understanding of how the panel works, and hopefully debug the issues themselves so they know for further reference.

All I ask is that they are at least considered.

Kind regards,
Jack.

PR #2554, opened 2020-10-18
Duplicate save/invite button at top

notAreYouScared:
Adds the Save / Invite User button at the top also

Closes https://github.com/pterodactyl/panel/issues/2383

image
image

Issue #2383, opened 2020-09-21
[Subusers] Add select all button back, as-well as "done" at top.


fadedmax:
**Is your feature request related to a problem? Please describe.**
As i have to migrate all my subusers over too the new system it would help to have a done button back.

**Describe the solution you'd like**
I just want users to have "control" permissions, and i dont really like scrolling all the way down, this is ui/ux
it would also help to have a select all button back.

**Additional context**
<img width="690" alt="Screen Shot 2020-09-21 at 00 17 23" src="https://user-images.githubusercontent.com/32599075/93737215-d3ccaf00-fb9f-11ea-8be8-d2cdb054d2e6.png">
<img width="694" alt="Screen Shot 2020-09-21 at 00 17 31" src="https://user-images.githubusercontent.com/32599075/93737222-d8916300-fb9f-11ea-937d-b3d3e9a687ae.png">

**ui/ux** and **1.0**

fadedmax:
this is ui and ux mainly, good first issue.

I tried to polish this as best i can, but its 2am and i need sleep.

notAreYouScared:
I see no need to have a select all button for all the permissions when each has a check box for each section.

fadedmax:
@AreYouRlyScared I do, if a server owner wants to quickly add a bunch of perms, its just much simpler to press 3 times then 15 and scrolling a ton

PR #3366, opened 2021-05-18
Minor UI fixes

il516:
The first fix is the Ram Usage % being right against the line, like below
**Before**
image

**After**
image

There were two buttons to invite a user to have access to a server, I removed the top one, considering that the owner of the server will have to scroll to the bottom to check the desired checkboxes.

The final fix is having the database button be on the right to be the same as the other buttons when using mobile.

**Before**
image

**After**
image


notAreYouScared:
The duplicate button was added by request of others so they didn't have to scroll down if those permissions weren't going to be assigned. 

Issue https://github.com/pterodactyl/panel/issues/2383
Pr https://github.com/pterodactyl/panel/pull/2554

il516:
Oh, okay. I added that back.

PR #3379, opened 2021-05-27
#BlameDane (fix backup pruning)

matthewpi:
Previously running the prune backups command would prune any backups that are currently running (unless they were already completed), because the `created_at` where in the query was not functioning properly.


DaneEveritt:
This is why I don't trust that whereDate selector...

PR #3400, opened 2021-06-06
Added mysql over ssl connection support

TheBlueKingLP:
Added option to secure database connection with SSL


bastianleicht (approved):

config/database.php
```diff
@@ -45,7 +45,14 @@
             'collation' => 'utf8mb4_unicode_ci',
             'prefix' => env('DB_PREFIX', ''),
             'strict' => env('DB_STRICT_MODE', false),
+            'sslmode' => env('DB_SSLMODE', 'prefer'),
             'timezone' => env('DB_TIMEZONE', Time::getMySQLTimezoneOffset(env('APP_TIMEZONE', 'UTC'))),
+            'options'   => array(
+                PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT => env('DB_VERIFY_SERVER_CERT', false),
```
DaneEveritt, line 51: This should be defaulted true and people can set it to false if they explicitly need it disabled.

config/database.php
```diff
@@ -45,7 +45,14 @@
...
+                PDO::MYSQL_ATTR_SSL_KEY => env('DB_SSL_KEY', ''),
+                PDO::MYSQL_ATTR_SSL_CERT => env('DB_SSL_CERT', ''),
+                PDO::MYSQL_ATTR_SSL_CA => env('DB_SSL_CA', ''),
```
DaneEveritt, line 54: These environment names should all be named following the constant name, e.g. `MYSQL_ATTR_SSL_CA`, similar to [Laravel's default setup](https://github.com/laravel/laravel/blob/8.x/config/database.php#L61-L63).

config/database.php
```diff
@@ -45,7 +45,14 @@
             'collation' => 'utf8mb4_unicode_ci',
             'prefix' => env('DB_PREFIX', ''),
             'strict' => env('DB_STRICT_MODE', false),
+            'sslmode' => env('DB_SSLMODE', 'prefer'),
             'timezone' => env('DB_TIMEZONE', Time::getMySQLTimezoneOffset(env('APP_TIMEZONE', 'UTC'))),
+            'options'   => array(
```
DaneEveritt, line 50: This codebase does not use `array()`, you should be using `[]` and there should only be a single space between the key and the arrow.

.env.example
```diff
@@ -14,6 +14,11 @@ DB_PORT=3306
 DB_DATABASE=panel
 DB_USERNAME=pterodactyl
 DB_PASSWORD=
+DB_SSLMODE=
```
DaneEveritt, line 17: These don't need to exist in the example file.

DaneEveritt (changes requested):

matthewpi:
Replaced by https://github.com/pterodactyl/panel/pull/3464

PR #3419, opened 2021-06-17
Permission for referencing other tables (foreign keys)

Leystryku:
Add permission to reference other tables so one can use foreign key constraints i.e. FOREIGN KEY(x) REFERENCES tbl(y)


DaneEveritt (approved):

niekcandaele:
Bump :)

Is there anything I can do to help this get merged? I see the tests fail because of a duplicate username but the test is about schedules. Perhaps it just needs a rerun?

PR #3421, opened 2021-06-18
Updated formatting for the docker compose file to fix a error.

mhmatthewhugley:
This fixes the error message "There is an error in the yaml syntax: YAMLSemanticError: Map keys must be unique; "<<" is repeated" which is received if trying to deploy the docker compose using portainer it will not even allow you to click the button because of the formatting error. This is a page showing a example of what I did "http://yaml.org/type/merge.html" Instead of 2 merge requests docker/portainer only allows 1 merge request in this situation so they had to be combined into 1 line instead of 2.

### DIFF
--- a/docker-compose.example.yml
+++ b/docker-compose.example.yml
@@ -61,8 +61,7 @@ services:
       - "/srv/pterodactyl/certs/:/etc/letsencrypt/"
       - "/srv/pterodactyl/logs/:/app/storage/logs"
     environment:
-      <<: *panel-environment
-      <<: *mail-environment
+      <<: [*panel-environment, *mail-environment]
       DB_PASSWORD: *db-password
       APP_ENV: "production"
       APP_ENVIRONMENT_ONLY: "false"


parkervcp:
The way we have it should work. Each line is an item in the array. It's the standard yaml format.

mhmatthewhugley:
> 
> 
> The way we have it should work. Each line is an item in the array. It's the standard yaml format.

For me the way it is does not work at all. I am using OMV installed on debian then docker and portainer. This is what it says my docker version is "5:20.10.7~3-0~debian-buster" and the portainer version that I tried it with and the one that gave me the error is "2.5.1" If nothing else I think there should atleast be a note saying if you get the error that I received then you can fix it by changing those 2 lines into 1.

parkervcp:
Does docker-compose have this error? Portainer is not a supported platform right now.

mhmatthewhugley:
> 
> 
> Does docker-compose have this error? Portainer is not a supported platform right now.

I dont know how portainer sends the compose information to docker. I assume it just runs whatever command you would run in the command prompt but instead you are just clicking the button in the gui but I am not 100% sure. So I cant tell you if it had the error. Anyway If the fix I suggest work in whatever testing environment you guys test them on. Then it is probably in a format that is more widely accepted and might cause less issues even if someone uses it somewhere other then what you guys actively support.

DaneEveritt (approved):

TekExplorer:
This is merged to the wrong branch guys... This should go to develop.
@DaneEveritt

PR #3431, opened 2021-06-22
ServerCreationService: send 'start_on_completion' option to wings

matthewpi:
Untested but should work :)


VottusCode:
"Untested but should work :)" - matthew's last words

just joking, tested it, works for me.

PR #3459, opened 2021-07-09
yarn upgrade

matthewpi:


matthewpi:
This PR has been merged into the `v2` branch awhile ago and will stay that way.  I don't really see a reason to add this to `v1` anymore.

PR #3465, opened 2021-07-13
Remove userInteraction from 'config_startup'

notAreYouScared:
Removes UserInteraction from existing eggs in the panel, as its no longer supported.

Co-created with @lancepioch


trenutoo:
I've tested this and it works as it should.

```
{
    "done": ")! For help, type ",
    "userInteraction": [
        "Go to eula.txt for more info."
    ]
}
```
turns into

```
{
    "done": ")! For help, type "
}
```

PR #3469, opened 2021-07-15
Fix backup dropdown permissions

### LINKED ISSUE #3375: Backup context menu not shown without download permission
The context menu on the backup page won't show when the user does not have permission to download a backup.

As a user I would like to allow another user to restore a backup without allowing the other user to download a backup. It should also display the three dots for the context menu (and therefore the context menu) when the user has either the permission to download, restore or delete a backup.

None.

**Additional context**
In the [BackupRow.tsx#L84](https://github.com/pterodactyl/panel/blob/develop/resources/scripts/components/server/backups/BackupRow.tsx#L84) file, it is only checked if the user has the download backup permission before it decides to render the three dots for the context menu. It does not check for the restore or delete permission.

Linux123123:
Just check for any of backup download, restore, delete permissions instead of only download.

Fixes #3375

### DIFF
--- a/resources/scripts/components/server/backups/BackupRow.tsx
+++ b/resources/scripts/components/server/backups/BackupRow.tsx
@@ -81,7 +81,7 @@ export default ({ backup, className }: Props) => {
                 </p>
                 <p css={tw`text-2xs text-neutral-500 uppercase mt-1`}>Created</p>
             </div>
-            <Can action={'backup.download'}>
+            <Can matchAny action={[ 'backup.download', 'backup.restore', 'backup.delete' ]}>
                 <div css={tw`mt-4 md:mt-0 ml-6`} style={{ marginRight: '-0.5rem' }}>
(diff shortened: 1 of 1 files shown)


resources/scripts/components/server/backups/BackupRow.tsx
```diff
@@ -81,7 +81,7 @@ export default ({ backup, className }: Props) => {
                 </p>
                 <p css={tw`text-2xs text-neutral-500 uppercase mt-1`}>Created</p>
             </div>
-            <Can action={'backup.download'}>
+            <Can matchAny action={[ 'backup.download', 'backup.restore', 'backup.delete' ]}>
```
matthewpi: Could you move the `matchAny` to after `action`?

matthewpi (changes requested):

matthewpi (approved):

PR #3486, opened 2021-07-22
Change paper startup command using Aikar's flag

magrigry:
Change paper startup command using Aikar's flag which will make the server more stable, is recommanded by many community and will avoid the container to freeze running out of memory.


matthewpi:
What you decided is the exact opposite effect of using Aikar's flags, it will increase the chances of a server running out of memory, not decrease it.  It might be recommended by people, but it shouldn't be blindly applied by default without the user needing it or doing their own testing.

PR #3488, opened 2021-07-23
style(templates): Default title tag for issue templates

trenutoo:
Adds default titles tags for bug and feature request templates to easily differentiate them when scrolling through issues.

[Bug]: Title here

[Feature Request]: Title here

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -1,5 +1,6 @@
 name: Bug Report
 description: Something isn't working quite right in the software.
+title: "[Bug]: "
 body:
 - type: markdown
   attributes:
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -1,6 +1,6 @@
 name: Bug Report
 description: Something isn't working quite right in the software.
-title: "[Bug]: "
+labels: [bug, not confirmed]
 body:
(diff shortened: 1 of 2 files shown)


FoksVHox:
Couldn't this also be achieved with labels?

trenutoo:
> Couldn't this also be achieved with labels?

It could have labels or both. I prefer titles as they provide the feedback right away for the initial submission, without opening or hovering over the issue when using third-party clients. Labels are something that members will assign based on their own discretion

FoksVHox:
> > Couldn't this also be achieved with labels?
> 
> 
> 
> It could have labels or both. I prefer titles as they provide the feedback right away for the initial submission, without opening or hovering over the issue when using third-party clients. Labels are something that members will assign based on their own discretion

Okay, personally I prefer labels over tags, but I'm not the one to make that call 😃

matthewpi:
I'd also prefer tags.  The way this should be setup is to label them as `bug` or `feature request` like currently exists, except `bug` should also have `not confirmed` by default.

DaneEveritt:
Can we just leave off bug entirely? I only apply that to specific issues that are raised, I think "not confirmed" is sufficient for bug reports, and once I have time to look at it I can apply the correct labels to it.

trenutoo:
> Can we just leave off bug entirely? I only apply that to specific issues that are raised, I think "not confirmed" is sufficient for bug reports, and once I have time to look at it I can apply the correct labels to it.

Remove titles, and only use not confirmed and feature request labels?

PR #3493, opened 2021-07-24
fix modals z-index overlapping

### LINKED ISSUE #3428: [UI - Dashboard] Console z-index overlaps with search modal

* Panel or Wings: Panel, frontend
* Version of Panel/Wings: Both `1.4.2` and `canary` (develop branch) are affected
* Panel Logs: -
* Wings Logs: -
* Server's OS: -
* Your Computer's OS & Browser: -

**Describe the bug**
Incorrect z-index causes the terminal loader to interfere with the modal

Screenshot 2021-06-21 at 11-05-00 Proxy Console

**To Reproduce**
Steps to reproduce this behavior:

1. Go to the server
2. Open the search modal
3. See error

**Expected behavior**
The z-index of the modal should be higher than the loading overlay of the console, not causing the overlay to interfere with the modal.

araskihu:
Fixes: #3428 

Tested on most pages, works fine. Other solution would have been to lower the SpinnerOverlay zIndex, but as I am not that familiar with this codebase I went with this solution.

### DIFF
--- a/resources/scripts/components/elements/Modal.tsx
+++ b/resources/scripts/components/elements/Modal.tsx
@@ -21,7 +21,8 @@ export interface ModalProps extends RequiredModalProps {
 }
 
 export const ModalMask = styled.div`
-    ${tw`fixed z-50 overflow-auto flex w-full inset-0`};
+    ${tw`fixed overflow-auto flex w-full inset-0`};
+    z-index: 10000;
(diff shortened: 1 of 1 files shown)


trenutoo:
The correct solution would be to lower the spinner overlays z-index instead to 40-45. Right now, it's 9999, and this is pushing another forced z-index to 10000

PR #3526, opened 2021-08-04
ui(server): fix used backup count

matthewpi:
Fixes the used backup count on the backups tab by using the same query used to check if the server is going to exceed it's backup limit.  The used backup count will also properly mutate when a backup is created or removed by a user.

### DIFF
--- a/app/Services/Backups/InitiateBackupService.php
+++ b/app/Services/Backups/InitiateBackupService.php
@@ -9,6 +9,7 @@ use Pterodactyl\Models\Backup;
 use Pterodactyl\Models\Server;
 use Illuminate\Database\ConnectionInterface;
 use Pterodactyl\Extensions\Backups\BackupManager;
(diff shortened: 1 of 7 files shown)


app/Services/Backups/InitiateBackupService.php
```diff
@@ -174,4 +172,12 @@ public function handle(Server $server, string $name = null, bool $override = fal
             return $backup;
         });
     }
+
+    public function getNonFailedBackups(Server $server): HasMany
```
DaneEveritt: This should be in a repository if anything, not on a service. It is also called in one place, I'm fine if you just do it right there in the controller.

DaneEveritt (changes requested):

app/Services/Backups/InitiateBackupService.php
DaneEveritt: Or even a scope on the model.

DaneEveritt (approved):

Issue #2079, opened 2020-05-29
Servers should be able to default to oom-killer enabled


iamkubi:
Some virtualizations (Parallels and OpenVZ based on users I've talked to) will fail to start a container if oom-killer is disabled.  When creating a new server this defaults to `off`, but for systems where `on` is the only valid option it would be useful to make the default value a system-wide setting.

iamkubi:
The guy I just talked to had this on KVM, so I don't know what these people have in common.

DaneEveritt:
If someone wants to PR something for this, they can, but there is a reason I don't support OpenVZ systems.

iamkubi:
Wouldn't this be a panel thing?  I was imagining having a node setting that changes the default value of the radio button on the create server page.  I wouldn't think anything needs to be done in the daemon, unless the Go daemon doesn't currently support the oom-killer setting.

DaneEveritt:
It is available per-server in the panel, I don't see any reason to make it per-node which only adds complexity to try and support an already unsupported system.

DaneEveritt:
Actually, on that note, I'm just going to close this out since it is easy enough to toggle it on each server.

SmartHosting:
sorry but I have the same question even tho it's easy enough not with whmcs my boy

57-Wolve:
@iamkubi I was able to fix our issue today by changing the oom_disable from true to false on server creation in [app/Services/Servers/ServerCreationService.php](https://github.com/pterodactyl/panel/blob/1260965dfdc162e9a8507d148467e2ec3f2f3ae2/app/Services/Servers/ServerCreationService.php#L228) line #228. This allowed us to create docker containers again as it was by default being disabled on container creation in the panel.

```
`INFO: [Aug 13 22:35:56.318] beginning installation process for server server=0cc279e4-1bdd-405b-84ec-09f20db0be83
INFO: [Aug 13 22:36:00.803] creating install container for server process install_script=/tmp/pterodactyl/0cc279e4-1bdd-405b-84ec-09f20db0be83/install.sh server=0cc279e4-1bdd-405b-84ec-09f20db0be83
INFO: [Aug 13 22:36:01.350] running installation script for server in container container_id=71ac1596a5513121c70020c7e5ab14c0d63e6a78433cd0b4a17143f55fcd9870 server=0cc279e4-1bdd-405b-84ec-09f20db0be83
[...]

PR #3548, opened 2021-08-18
Introduce OOM Killer to Server Creation

trenutoo:
Introduces OOM killer option on server creation. This also solves issues for specific OpenVZ setups where OOM breaks the server #2079

Description for the server owner variable that email must be used to prevent common confusion when a username is typed and no results are displayed

Changes CPU pinning explanation to threads similar to CPU limit

image

### DIFF
--- a/resources/views/admin/servers/new.blade.php
+++ b/resources/views/admin/servers/new.blade.php
@@ -216,8 +216,8 @@
                     </div>
                     <div class="form-group col-xs-12">
                         <div class="checkbox checkbox-primary no-margin-bottom">
-                            <input type="checkbox" id="pOomKillerEnabled" name="oom_disabled" value="0" {{ \Pterodactyl\Helpers\Utilities::checked('oom_disabled', 0) }} />
-                            <label for="pOomKillerEnabled" class="strong">Enable OOM Killer</label>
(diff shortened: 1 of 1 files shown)


notAreYouScared:
I see no reason for this change with the admin area getting updated to react.

trenutoo:
> I see no reason for this change with the admin area getting updated to react.

Introducing the OOM killer on the creation page has no real effect on V2 progress, but it resolves the issue that many are having with their VPS.

resources/views/admin/servers/new.blade.php
```diff
@@ -210,6 +214,14 @@
 
                         <p class="text-muted small"><strong>Advanced</strong>: The IO performance of this server relative to other <em>running</em> containers on the system. Value should be between <code>10</code> and <code>1000</code>. Please see <a href="https://docs.docker.com/engine/reference/run/#/block-io-bandwidth-blkio-constraint" target="_blank">this documentation</a> for more information about it.</p>
                     </div>
+                    <div class="form-group col-xs-12">
+                        <div class="checkbox checkbox-primary no-margin-bottom">
+                            <input type="checkbox" id="pOomKillerEnabled" name="oom_disabled" value="0" {{ \Pterodactyl\Helpers\Utilities::checked('oom_disabled', 0) }} />
```
matthewpi: `pOomKillerEnabled` should be changed to `pOomDisabled` to follow the `id` and be consistent with the rest of the code.

matthewpi (approved):

PR #3550, opened 2021-08-18
Filter allocations by assigned in application api.

LoaiDev:
I have ran into this issue when I was creating a website that was using the panel's application api. When creating a new server the data must include a default allocation for the server to use, but most of the time i just need any random non-assigned allocation which is easy at the start but once you have over 50 or 100 servers you either have to make the per_page bigger or search through pages for unassigned one. Now can just set assigned_to to null to get those ones.


matthewpi:
Thanks for the pull request.  Sadly, because of the changes being made in v2 to the application api, I'd like to avoid making any changes on v1 that may end up being reverted or broken by the changes made in v2.

PR #3552, opened 2021-08-18
fix: Forge egg stdin

trenutoo:
Like the current Paper egg, this resolves the issue affecting Forge egg after migration to alpine-based Yolks images. Without this, the servers do not accept any input, including stop power action.


trenutoo:
Superseded by #3676

PR #3600, opened 2021-09-05
Custom SSL Cert Path for Nodes in Node Settings

andrei0465:
php artisan migrate --path=/database/migrations/2021_09_05_152343_add_sslcert_sslkey_column.php


ItzKazuki:
Screenshot_13
I've tried it and it works... but maybe this should be fixed first :)

ItzKazuki:
Screenshot_15
maybe in the future like this :)

matthewpi:
While this would be nice, there is no reason for us to keep adding individual Wings config options as fields in the database, at some point we may expose editing the Wings config from the admin panel, but until then we aren't going to add any more fields.

PR #3616, opened 2021-09-13
Fix SSL config docker

waseem-h:
By default nginx in alpine has the following include statement: `include /etc/nginx/http.d/*.conf`. Normal HTTP configuration was working because it was being copied over to the image at build time. I've removed that. Now the entrypoint decides which config to use and that gets picked up by nginx properly. Tested with the latest version of docker-compose and docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,6 @@ RUN rm /usr/local/etc/php-fpm.conf \
     && sed -i s/ssl_session_cache/#ssl_session_cache/g /etc/nginx/nginx.conf \
     && mkdir -p /var/run/php /var/run/nginx
 
-COPY .github/docker/default.conf /etc/nginx/http.d/default.conf
 COPY .github/docker/www.conf /usr/local/etc/php-fpm.conf
 COPY .github/docker/supervisord.conf /etc/supervisord.conf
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -30,7 +30,6 @@ RUN rm /usr/local/etc/php-fpm.conf \
     && sed -i s/ssl_session_cache/#ssl_session_cache/g /etc/nginx/nginx.conf \
     && mkdir -p /var/run/php /var/run/nginx
 
-COPY .github/docker/default.conf /etc/nginx/http.d/default.conf
```
parkervcp: I'd rather leave this just in case.

waseem-h: Leaving it in will cause the if check to not work in entrypoint.sh which determines if the config was already in place or not

waseem-h: This one to be specific `if [ -f /etc/nginx/conf.d/default.conf ]; then`
@parkervcp

waseem-h: Another way of solving it would be to change the name of the custom conf to be `panel.conf` and have a check on that and remove the default conf when its configured

waseem-h:
@parkervcp It should keep the original file now

parkervcp (approved):

waseem-h:
@parkervcp can you please get this merged?

PR #3675, opened 2021-10-09
Added Audit logs

Tdash2:
Added a simple audit log view to the admin panel.


DaneEveritt:
Thanks, but we're not accepting changes to the software at this time as we work to complete a massive rewrite of the admin area.

PR #3677, opened 2021-10-10
Change ^C to ^^C in egg menus

JohnB17:
For some reason, ^^C actually sends a SIGINT to the server, and ^C does not, so I changed the egg UI to say to add ^^C


DaneEveritt:
This should be reported as a bug and fixed in Wings, not hacked around like this.

PR #3688, opened 2021-10-17
"Manage User" for consistency

joshcannotcode:
just noticed, nothing major but i've changed for consistency


FoksVHox:
I'm sure this is already fixed in the rewrite of the admin areas (Aka v2), so you're probably just best off with closing this pr

joshcannotcode:
> I'm sure this is already fixed in the rewrite of the admin areas (Aka v2), so you're probably just best off with closing this pr

Ah, cheers, will do

PR #3700, opened 2021-10-22
Don't force enable-query on Minecraft servers

Spirit55555:
The query protocol is not required for a Minecraft server to work, so there is no need to force it on.

Make sure to set the query.port to the server port, so it works, if the server owner enables query.
This was not the case for all Minecraft eggs.

Minecraft (Java) runs on TCP and query on UDP, so they can both use the same port.

### DIFF
--- a/database/Seeders/eggs/minecraft/egg-forge-minecraft.json
+++ b/database/Seeders/eggs/minecraft/egg-forge-minecraft.json
@@ -4,7 +4,7 @@
         "version": "PTDL_v1",
         "update_url": null
     },
-    "exported_at": "2021-10-10T07:10:13-04:00",
+    "exported_at": "2021-10-22T19:29:26+02:00",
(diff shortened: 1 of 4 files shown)


trenutoo (approved):
The query port is already parsed to the server port. Querying is most likely enabled by default to prevent complaints about servers not responding to trackers?

Although, in that case, it's better to have it be an optional variable than forcefully enabled. 

LGTM 👍

Spirit55555:
> The query port is already parsed to the server port. Querying is most likely enabled by default to prevent complaints about servers not responding to trackers?

The query port was not set on paper and vanilla, so this pull request fixes that inconsistency.

Most trackers I have seen, kindly tells the server owner how to enable query, so that should not be a problem.

DaneEveritt (approved):

PR #3706, opened 2021-10-24
change display format of the container uptime

trenutoo:
Display `day, hour, min` if days is more than 0, otherwise default to existing `hour, min, sec`. Removes pads to make it cleaner in this new format. There are probably better ways to do this especially with the frags.

image

image

### DIFF
--- a/resources/scripts/components/server/UptimeDuration.tsx
+++ b/resources/scripts/components/server/UptimeDuration.tsx
@@ -1,14 +1,23 @@
 import React from 'react';
 
 export default ({ uptime }: { uptime: number }) => {
-    const hours = Math.floor(Math.floor(uptime) / 60 / 60);
+    const days = Math.floor(uptime / (24 * 60 * 60));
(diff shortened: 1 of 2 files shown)


resources/scripts/components/server/UptimeDuration.tsx
```diff
@@ -1,14 +1,23 @@
...
 export default ({ uptime }: { uptime: number }) => {
-    const hours = Math.floor(Math.floor(uptime) / 60 / 60);
+    const days = Math.floor(uptime / (24 * 60 * 60));
+    const hours = Math.floor(Math.floor(uptime) / 60 / 60 % 24);
     const remainder = Math.floor(uptime - (hours * 60 * 60));
-    const minutes = Math.floor(remainder / 60);
+    const minutes = Math.floor(remainder / 60 % 60);
     const seconds = remainder % 60;
 
     return (
```
DaneEveritt:
```
if (days > 0) {
    return <>{days}d ...</>
}

return <>{hours}...</>
```
```

DaneEveritt (approved):

PR #3707, opened 2021-10-24
Fix Nginx conf path in docker-compose.example.yml

TheExeQ:
Volume path seems to be wrong. `etc/nginx/conf.d/` conf.d doesn't exist in this directory.


trenutoo:
It would be useful to provide information on the PR. From looking at the files, I assume this is something that was missed with https://github.com/pterodactyl/panel/commit/8b236c6907f84a46b885f6d71e6969ce9c97e19c when the paths were changed?

**ED:** According to your edit that's it. I've tested this and LGTM.

PR #3714, opened 2021-10-28
Change "Game Port" to "Primary Port"

Zarklord:
This changes makes the terminology match what's said in the Network tab
And removes the confusing requirement that the Primary Port be the game port even if that port is irrelevant to the server hoster when supplying connection information to users.


notAreYouScared:
No point in making changes to the admin area, as it will just be over written when v2 comes out

PR #3746, opened 2021-11-15
GSL Token Modal Feature

Boy132:
This PR adds a GSL Token Modal Feature that will be shown if the console contains messages like `gsl token expired` or `account not found`. When entering a new token (or leaving the field blank) the server will be restarted automatically to apply the changes.
I added this feature to the default Garry's Mod and CS:GO Egg because I know the GSL Token is used for these games but there might be other source games where this could be added.

image

I know this is a game specific feature but I think it would be helpful. The current problem is that servers with invalid gsl tokens are stuck at "Starting" and especially with Garry's Mod the message in the console can be easily overlooked.


trenutoo:
This doesn't pass ESLint. I'd recommend installing and running this through ESLint to see what's wrong (tabs vs. spaces and inconsistent spacing)

Boy132:
> This doesn't pass ESLint. I'd recommend installing and running this through ESLint to see what's wrong (tabs vs. spaces and inconsistent spacing)

All ESLint errors should now be resolved.

DaneEveritt (approved):

fabm3n:
There is a warning in your feature:
image

PR #3753, opened 2021-11-18
Add app_url for mail sender

NotYusta:


trenutoo:
It would be added during the [setup process](https://github.com/pterodactyl/panel/blob/develop/app/Console/Commands/Environment/AppSettingsCommand.php#L107). Should it be missing a critical step of the installation has been skipped.

PR #3763, opened 2021-11-22
Added Server Suspended message

Tdash2:
Added a message the show that your server is suspended on the console page. 

image


trenutoo:
This view would only be displayed to the admins making this message unnecessary. Normal users would see a suspended error message on all server pages, while admins currently don't see it on the console page.

image

https://github.com/pterodactyl/panel/blob/develop/resources/scripts/routers/ServerRouter.tsx#L47

Tdash2:
I have never see a suspend server from a normal users side. I thought it stilled showed the normal console page.  Is there a reason that the page is not displayed to admin?

trenutoo:
> I have never see a suspend server from a normal users side. I thought it stilled showed the normal console page. 

You'd have to actually log in as a normal user and not admin to see it. I've done that in the screenshot above earlier to verify it. Displaying message to contact administrator for admins doesn't really make sense.
> Is there a reason that the page is not displayed to admin?

Probably an oversight to display the exception on the console, but there's a pending pull request allowing admins to access the servers when they're suspended. https://github.com/pterodactyl/panel/pull/2677

PR #3766, opened 2021-11-23
Clarify password creation method

mrxbox98:


DaneEveritt:
This sentence was already grammatically (and logically) correct, but this PR makes it incorrect.

PR #3796, opened 2021-12-09
Add the DB_PORT environmental variable to the example Docker Compose config

Wuzado:
#3762 / commit 50881f2 introduces the requirement for a DB_PORT environmental variable, which is not present in the current example config. If an user chooses to update their Docker image to a version after the commit 50881f2, the startup will fail with the panel `Waiting for database connection...`.

I'm not too confident in my shell abilities and there are a couple ways to solve that issue, so I'll leave that to the maintainers of Pterodactyl, but I'd recommend adjusting the entrypoint file so in case the DB_PORT environmental variable is missing, it either falls back to the port 3306, or crashes with an according message.

The default port for MariaDB and the previous setting in the entrypoint file is 3306, which is the value of the new variable.


trenutoo (approved):
LGTM.  Would also have to set default DB_PORT in the entrypoint for backward compatibility for users that have older compose files as discussed on discord.

```
if [[ -z $DB_PORT ]]; then
  echo -e "DB_PORT not specified, defaulting to 3306"
  DB_PORT=3306
fi
```

Wuzado:
Entrypoint fix submitted as the PR #3808.

PR #3802, opened 2021-12-12
Add current network traffic and network chart

Prophet731:
Add total network traffic to stats block since startup

### DIFF
--- a/resources/scripts/helpers.ts
+++ b/resources/scripts/helpers.ts
@@ -67,3 +67,49 @@ export function hashToPath (hash: string): string {
 export function formatIp (ip: string): string {
     return /([a-f0-9:]+:+)+[a-f0-9]+/.test(ip) ? `[${ip}]` : ip;
 }
+
+/**
+ * Network helpers
+ */
+// Store old values to compare to new values for network speed.
+export let oldBytesReceived = 0;
+export let oldBytesSent = 0;
+export let oldBytesReceivedChart = 0;
+export let oldBytesSentChart = 0;
+
+export function setOldNetworkValues (bytesReceived: number, bytesSent: number, chart: boolean | null): void {
+    if (chart === true && chart !== null) {
+        oldBytesReceivedChart = bytesReceived;
+        oldBytesSentChart = bytesSent;
+    } else {
+        oldBytesReceived = bytesReceived;
+        oldBytesSent = bytesSent;
+    }
+}
+
+// Convert bytes to bits per second (bps) by comparing to old values.
+export function bytesToBps (bytesReceived: number, bytesSent: number, chart: boolean | null): number {
+    const bpsReceived = (bytesReceived - (chart === true ? oldBytesReceivedChart : oldBytesReceived)) * 8;
+    const bpsSent = (bytesSent - (chart === true ? oldBytesSentChart : oldBytesSent)) * 8;
+
+    setOldNetworkValues(bytesReceived, bytesSent, chart);
+
+    return bpsReceived + bpsSent;
+}
+
+// Convert bits per second (bps) to human-readable format.
+export function bpsToHuman (bps: number): string {
+    if (bps < 1000) {
+        return `${bps} bps`;
+    }
+
+    if (bps < 1000000) {
+        return `${(bps / 1000).toFixed(2)} kbps`;
+    }
+
+    if (bps < 1000000000) {
+        return `${(bps / 1000000).toFixed(2)} mbps`;
+    }
+
+    return `${(bps / 1000000000).toFixed(2)} gbps`;
+}
--- a/resources/scripts/components/server/ServerDetailsBlock.tsx
+++ b/resources/scripts/components/server/ServerDetailsBlock.tsx
@@ -2,7 +2,7 @@ import React, { useEffect, useState } from 'react';
 import tw, { TwStyle } from 'twin.macro';
 import { faCircle, faEthernet, faHdd, faMemory, faMicrochip, faServer, faNetworkWired } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { bytesToHuman, megabytesToHuman, formatIp } from '@/helpers';
+import { bytesToHuman, megabytesToHuman, formatIp, bytesToBps, bpsToHuman } from '@/helpers';
 import TitledGreyBox from '@/components/elements/TitledGreyBox';
 import { ServerContext } from '@/state/server';
 import CopyOnClick from '@/components/elements/CopyOnClick';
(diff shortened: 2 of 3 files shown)


notAreYouScared:
I feel the better approach would be to calculate the difference between the previous number and the new number to get the actual usage per second. Then if the user hovers over it, then we should show total since last boot

Prophet731:
image

Updated so now it looks like this. I don't know how to add the tooltip to it though. If someone wants to give a quick how-to then I can amend the PR.

notAreYouScared:
I wanted to do this a long time ago, just never knew how, should we keep the rx and tx separate still?

Prophet731:
If we had a way to store that data then we can have the overall total traffic the server has ever sent but as far as I know, it doesn't support that (yet). So maybe for now just having the tooltip for total traffic since the container has been started would be the easist way.

ItzKazuki:
maybe updated in the form of statistics such as cpu and ram usage

notAreYouScared:
> If we had a way to store that data then we can have the overall total traffic the server has ever sent but as far as I know, it doesn't support that (yet). So maybe for now just having the tooltip for total traffic since the container has been started would be the easist way.

We could add columns to the MySQL table for `servers` like a `total_tx` `total_rx`
image

Before we send the server shutdown command we can take the latest values we have and add them to the current value stored.

Not sure of another way to do it.

Prophet731:
image

Added the graph for the networking as well.

Prophet731:
> > If we had a way to store that data then we can have the overall total traffic the server has ever sent but as far as I know, it doesn't support that (yet). So maybe for now just having the tooltip for total traffic since the container has been started would be the easist way.
> 
> We could add columns to the MySQL table for `servers` like a `total_tx` `total_rx` image
> 
> Before we send the server shutdown command we can take the latest values we have and add them to the current value stored.
> 
> Not sure of another way to do it.

That should be in its own separate table so that way it has better historical data and be able to do more advanced queries. That's a separate issue though.

resources/scripts/helpers.ts
```diff
@@ -67,3 +67,49 @@ export function hashToPath (hash: string): string {
 export function formatIp (ip: string): string {
     return /([a-f0-9:]+:+)+[a-f0-9]+/.test(ip) ? `[${ip}]` : ip;
 }
+
+/**
+ * Network helpers
+ */
+// Store old values to compare to new values for network speed.
+export let oldBytesReceived = 0;
```
DaneEveritt, line 75: These should not be used in this manner, if you were to call these functions in two places at the same time you would be introducing race conditions and get mangled data depending on the order they're called.

The old rates should be tracked in the component itself, and most likely this function doesn't need to be a helper, just a function in the component.

DaneEveritt (changes requested):

Prophet731:
I was trying to reduce duplicate code since it would be used in two different locations. I don't know react that well (vue.js for me).

I created the two variables for the chart and the standard text, otherwise ran into where the chart would always be 0.

What way would be good in your view?

resources/scripts/helpers.ts
Prophet731: I was trying to reduce duplicate code since it would be used in two different locations. I don't know react that well (vue.js for me).

I created the two variables for the chart and the standard text, otherwise ran into where the chart would always be 0.

What way would be good in your view?

trenutoo:
> If we had a way to store that data then we can have the overall total traffic the server has ever sent but as far as I know, it doesn't support that (yet). So maybe for now just having the tooltip for total traffic since the container has been started would be the easist way.

Calculating usage per second and/or statistics for the container lifetime would align with how all the other statistics are displayed since they're fetched from Docker stats.

In my opinion, trying to make Pterodactyl be a monitoring tool and manage all that data in a database is overkill. Should someone want this, they can use SNMP or Docker's own endpoints to collect it using dedicated monitoring tools like Grafana or LibreNMS which many hosts are doing right now.

notAreYouScared:
> > If we had a way to store that data then we can have the overall total traffic the server has ever sent but as far as I know, it doesn't support that (yet). So maybe for now just having the tooltip for total traffic since the container has been started would be the easist way.
> 
> Calculating usage per second and/or statistics for the container lifetime would align with how all the other statistics are displayed since they're fetched from Docker stats.
> 
> In my opinion, trying to make Pterodactyl be a monitoring tool and manage all that data in a database is overkill. Should someone want this, they can use SNMP or Docker's own endpoints to collect it using dedicated monitoring tools like Grafana or LibreNMS which many hosts are doing right now.

So scrap the storing data part. We should just show real time usage in per second and not store anything

Prophet731:
> > If we had a way to store that data then we can have the overall total traffic the server has ever sent but as far as I know, it doesn't support that (yet). So maybe for now just having the tooltip for total traffic since the container has been started would be the easist way.
> 
> Calculating usage per second and/or statistics for the container lifetime would align with how all the other statistics are displayed since they're fetched from Docker stats.
> 
> In my opinion, trying to make Pterodactyl be a monitoring tool and manage all that data in a database is overkill. Should someone want this, they can use SNMP or Docker's own endpoints to collect it using dedicated monitoring tools like Grafana or LibreNMS which many hosts are doing right now.

Wings can just post to an API endpoint in Pterodactyl with the current data usage it sees since it already has access to the docker process. Something like 5-minute checks doesn't have to be every minute. In Ptero, it just takes a few prams to store in a table with a couple of columns (container_id, data_ingress, data_egress, timestamp). 

Pterodactyl can be a monitoring tool ONLY for what it does through Wings, doesn't need to be monitoring Apache, NGINX, Redis, etc..., that's where Prometheus and the likes come in.

A query to just pull total data by (day,month, year) isn't hard todo.

```sql
SELECT `container_id`, SUM(`data_ingress`) AS 'total_ingress', SUM(`data_egress`) AS 'total_egress'
FROM `container_stats`
WHERE `container_id` = 'abc123'
GROUP BY DAY(timestamp), MONTH(timestamp), YEAR(timestamp)
```

This though would be an entirely new feature request as it'll involve Wings as well.

DaneEveritt:
Please do not add logic to store resource usage. We don't do this at all for existing things, and network data is not where I want to start. This should simply be real-time data.

crispyaus:
Update on this?

Prophet731:
@DaneEveritt Do you have any updates to my comment here? 

https://github.com/pterodactyl/panel/pull/3802/files/ca27909bb51fa1fe04f4b4af01759e10b21b251d#r767311026

resources/scripts/helpers.ts
DaneEveritt: Handle it in the component view itself in that case, or use an object passed back into the function needing it.

DaneEveritt:
Closed in 62b178ed0297549be9eb009c968d5d47d4fe8ffb

PR #3807, opened 2021-12-13
Feature/task order

Jckf:
Tasks currently execute from the lowest sequence_id (not necessarily sequence_id 1) up to the first break in sequence IDs. Ie. given tasks with sequence_ids 10, 11, 13, 14; execution will start at 10, and stop at 11. 13 and 14 will not execute.

Tasks are also not sorted by sequence_id when viewed in the panel, which becomes confusing if a schedule has been edited instead of recreated when a new task needed to appear somewhere other than at the end.

This PR makes it so that task will execute in order, even if the sequence contains breaks, and makes it so that they are displayed in order of execution when viewing a schedule.

None of this is a problem if the panel is used as-is without messing with its database, but you can perhaps view the PR as a step in the direction of making tasks re-orderable.

Disclaimer: I haven't had the time to test this yet 🙈, but I'm opening the PR already to get feedback in case I'm going about this all wrong or have made any mistakes that aren't obvious to me.


mason-rogers:
https://github.com/pterodactyl/panel/blob/develop/app/Http/Controllers/Api/Client/Servers/ScheduleTaskController.php#L124-L126
This query should already handle gaps happening in the sequence_id. It re-counts the tasks with a higher sequence than the one which just got deleted.

Jckf:
> https://github.com/pterodactyl/panel/blob/develop/app/Http/Controllers/Api/Client/Servers/ScheduleTaskController.php#L124-L126 This query should already handle gaps happening in the sequence_id. It re-counts the tasks with a higher sequence than the one which just got deleted.

Sure, and that'll fix things if you always delete a task after messing with the sequence IDs, but that won't necessarily always be the case, and I see no reason to explicitly not support sequences with gaps in them.

DaneEveritt:
Why are you messing with the sequence IDs manually outside the scope of the application? It is designed specifically so that there is never a gap in the sequence.

Jckf:
> Why are you messing with the sequence IDs manually outside the scope of the application? It is designed specifically so that there is never a gap in the sequence.

Because the application _has_ no way of reordering tasks.

The specific case that made me do this was having to insert an extra command in the middle of a schedule that already had 10-15 tasks. This is currently impossible without manually changing the sequence IDs.

And as previously stated, I don't see a reason to _explicitly not support_ sequences with gaps, when it can be added so easily. That's only inviting problems. Besides, if reordering is added in the future, then this will have made that _a tiny bit_ easier.

EDIT: If you feel that this change is meaningless without also introducing reordering, then I'll need some help to make that happen.

DaneEveritt:
I don't view it as meaningless, just didn't have any content as to how this scenario was happening with the current setup.

I'll be able to review this more in the coming weeks, currently have a lot going on outside of this project and don't have much access to my machine right now. :)

Jckf:
Do you need anything else from me here?

mason-rogers:
your PR is still marked as a draft by you so it was probably left on hold / unreviewed for that reason

PR #3813, opened 2021-12-16
File manager fix

### LINKED ISSUE #3003: Double file manager entry when you make a nested folder
* Panel or Daemon: Panel
* Version of Panel/Daemon: 1.2.1 
* Server's OS: Ubuntu
* Your Computer's OS & Browser: Windows 10 &Brave

**Describe the bug**
https://streamable.com/fd9i3v
If you make a nested folder, you can see it makes 2 components for it on the file manager till you refresh where its then fixed

**To Reproduce**
1. Go to 'file manager' on any server
2. Create 2 directories both beginning with the same directory (ie. "nice")
3. Profit

**Expected behavior**
Expected to not make the 2nd component after making the 2nd directory, but instead just leave it as is (how it'd be after a refresh)

mrxbox98:
This fixes https://github.com/pterodactyl/panel/issues/3003. It is probably possible to do it in a more efficient way.

### DIFF
--- a/resources/scripts/components/server/files/FileManagerContainer.tsx
+++ b/resources/scripts/components/server/files/FileManagerContainer.tsx
@@ -22,8 +22,8 @@ import { FileActionCheckbox } from '@/components/server/files/SelectFileCheckbox
 import { hashToPath } from '@/helpers';
 
 const sortFiles = (files: FileObject[]): FileObject[] => {
-    return files.sort((a, b) => a.name.localeCompare(b.name))
-        .sort((a, b) => a.isFile === b.isFile ? 0 : (a.isFile ? 1 : -1));
+    return [...new Set(files.sort((a, b) => a.name.localeCompare(b.name))
+        .sort((a, b) => a.isFile === b.isFile ? 0 : (a.isFile ? 1 : -1)))];
 };
 
 export default () => {
--- a/resources/scripts/components/server/files/FileManagerContainer.tsx
+++ b/resources/scripts/components/server/files/FileManagerContainer.tsx
@@ -22,8 +22,21 @@ import { FileActionCheckbox } from '@/components/server/files/SelectFileCheckbox
 import { hashToPath } from '@/helpers';
 
 const sortFiles = (files: FileObject[]): FileObject[] => {
-    return [...new Set(files.sort((a, b) => a.name.localeCompare(b.name))
-        .sort((a, b) => a.isFile === b.isFile ? 0 : (a.isFile ? 1 : -1)))];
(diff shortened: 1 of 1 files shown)


resources/scripts/components/server/files/FileManagerContainer.tsx
```diff
@@ -22,8 +22,21 @@ import { FileActionCheckbox } from '@/components/server/files/SelectFileCheckbox
...
-        .sort((a, b) => a.isFile === b.isFile ? 0 : (a.isFile ? 1 : -1));
+    let f: FileObject[] = files.sort((a, b) => a.name.localeCompare(b.name)).sort((a, b) => a.isFile === b.isFile ? 0 : (a.isFile ? 1 : -1));
+
+    for(let i=1;i<f.length;i++)
+    {
+        if(f[i].isFile)
+        {
+            continue;
+        }
+        if(f[i].name==f[i-1].name)
```
trenutoo: `===` instead of `==` and I'd recommend installing eslint, because the code style doesn't pass Pterodactyls eslint config,

I think `return sortedFiles.filter((file, index) => index === 0 || file.name !== sortedFiles[index - 1].name);` would be cleaner? I've made a quick test and seems to work:

```
const sortFiles = (files: FileObject[]): FileObject[] => {
    const sortedFiles = files.sort((a, b) => a.name.localeCompare(b.name)).sort((a, b) => a.isFile === b.isFile ? 0 : (a.isFile ? 1 : -1));
    return sortedFiles.filter((file, index) => index === 0 || file.name !== sortedFiles[index - 1].name);
};
```

resources/scripts/components/server/files/FileManagerContainer.tsx
```diff
@@ -22,8 +22,21 @@ import { FileActionCheckbox } from '@/components/server/files/SelectFileCheckbox
 import { hashToPath } from '@/helpers';
 
 const sortFiles = (files: FileObject[]): FileObject[] => {
-    return files.sort((a, b) => a.name.localeCompare(b.name))
-        .sort((a, b) => a.isFile === b.isFile ? 0 : (a.isFile ? 1 : -1));
+    let f: FileObject[] = files.sort((a, b) => a.name.localeCompare(b.name)).sort((a, b) => a.isFile === b.isFile ? 0 : (a.isFile ? 1 : -1));
```
trenutoo: This can be const as it's never re-assigned

mrxbox98: Fixed in latest commit

resources/scripts/components/server/files/FileManagerContainer.tsx
mrxbox98: Fixed in latest commit

trenutoo:
I'm not a project maintainer, so my review does nothing in this repo. However, LGTM and I've tested the change.

matthewpi (approved):

PR #3823, opened 2021-12-19
Allow customization of the update checker's cache time

Deveroonie:


Jelcoo:
Maybe rename to `PTERODACTYL_UPDATE_CACHE_TIME`

Deveroonie:
done i think 😅😅😅

DaneEveritt:
Please provide context to this PR, I don't see a specific reason to need this additional configuration option.

Deveroonie:
Just in case someone wants to check, say, every hour instead of every 10 mins

trenutoo:
The default time of 60 minutes should be more than enough. It's not a critical check and someone spamming cdn every minute would be a waste. 

Your change also has a breaking typo. Only one arrow, not two. =>

DaneEveritt:
Thanks, but I don't think this is a needed change.

PR #3826, opened 2021-12-20
Add user registration functionality

camwhi-te:
This PR introduces a new module called User Registration, which allows users to create accounts for themselves from the login page. This module can be disabled by editing the new config file at the root directory of the panel, `settings.json`. 

Please note; changing any settings in this file will require for the panel to rebuilt to sync the changes.


camwhi-te:
THIS WASN'T MEANT TO BE HERE

Craftzman7:
lol

PR #3838, opened 2022-01-01
Added Audit logs to the Client Side of the Panel

Tdash2:
As simple as the title says.

### DIFF
--- a/resources/scripts/components/server/logs/logsContainer.tsx
+++ b/resources/scripts/components/server/logs/logsContainer.tsx
@@ -1,0 +1,81 @@
+import React, { useEffect } from 'react';
+import { ServerContext } from '@/state/server';
+import useFlash from '@/plugins/useFlash';
+import ServerContentBlock from '@/components/elements/ServerContentBlock';
+import tw from 'twin.macro';
+import useSWR from 'swr';
+import FlashMessageRender from '@/components/FlashMessageRender';
+import Spinner from '@/components/elements/Spinner';
+import GreyRowBox from '@/components/elements/GreyRowBox';
+import getLogsRequests from '@/api/server/logs/getlogs';
+
+export interface LogsResponse {
+    requests: any[],
+}
+
+export default () => {
+    const uuid = ServerContext.useStoreState(state => state.server.data!.uuid);
+    const { clearFlashes, clearAndAddHttpError } = useFlash();
+    const { data, error } = useSWR<LogsResponse>([ uuid, '/logs' ], key => getLogsRequests(key));
+
+    useEffect(() => {
+        if (!error) {
+            clearFlashes('server:logs');
+        } else {
+            clearAndAddHttpError({ key: 'server:logs', error });
+        }
+    }, [ error ]);
+
+    return (
+        <ServerContentBlock title={'Audit Logs'} css={tw`flex flex-wrap`}>
+            <div css={tw`w-full`}>
+                <FlashMessageRender byKey={'server:logs'} css={tw`mb-4`} />
+            </div>
+            {!data ?
+                (
+                    <div css={tw`w-full`}>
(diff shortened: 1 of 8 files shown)


Jelcoo:
Yes basically this will break some stuff and it will not work as intended

Tdash2:
What would you like to see changed to make this work as intended

Tdash2:
I just went through and made a few changes, I think it should fix the issues that you bought up.

Jelcoo:
Okay first, don't use raw database imports, use models instead. Secondly, in app/Http/Controllers/Api/Client/Servers/ClogsController.php there are a lot of unused imports. The third thing is the spaghetti code in resources/scripts/components/server/logs/logsContainer.tsx. As last, I have a question, where does the C in ClogsController stand for?

Tdash2:
I will fix what you brought up:
> Okay first, don't use raw database imports, use models instead. Secondly, in app/Http/Controllers/Api/Client/Servers/ClogsController.php there are a lot of unused imports. The third thing is the spaghetti code in resources/scripts/components/server/logs/logsContainer.tsx.
 
> As last, I have a question, where does the C in ClogsController stand for?

a few of the addons out already have a logsController so I added the "c" in front of it to stop issues from occurring. I could changed them to clientLogsController.

Jelcoo:
Maybe use AuditLogsController

FoksVHox:
> I will fix what you brought up:
> 
> > Okay first, don't use raw database imports, use models instead. Secondly, in app/Http/Controllers/Api/Client/Servers/ClogsController.php there are a lot of unused imports. The third thing is the spaghetti code in resources/scripts/components/server/logs/logsContainer.tsx.
> 
> > As last, I have a question, where does the C in ClogsController stand for?
> 
> a few of the addons out already have a logsController so I added the "c" in front of it to stop issues from occurring. I could changed them to clientLogsController.

I don't think that would be the best solution. As of the current state, Pterodactyl doesn't offer official support for add-ons, and therefore Pterodactyl should not take these addons into account. IMO it would be best to remove the "C" completely.

trenutoo:
Without commenting on the code, since you're refactoring it, naming it logs or server logs is misleading since these would be audit logs. It should follow the naming convention and stick to audit-logs for the routes, description, and controller.

resources/scripts/components/server/logs/logsContainer.tsx
```diff
@@ -0,0 +1,117 @@
...
+                                (data.requests.map((item, key) => (
+                                    <GreyRowBox $hoverable={false} css={tw`mb-2`} key={key}>
+                                        <GreyRowBox $hoverable={false} key={key}>
+                                            <div css={tw`flex items-center w-full md:w-auto`}>          
+                                                <div css={tw`flex-initial ml-16 text-center`}>
+                                                    <p css={tw`text-sm`}>{item.user_id}</p>
+                                                    <p css={tw`mt-1 text-2xs text-neutral-300 uppercase select-none`}>User</p>
+                                                </div>
+                                                <div css={tw`flex-initial ml-4 text-center`}>
+                                                    {item.action === 'server:filesystem.download' ?
```
schrej, line 58: Please use a dictionary to map an action to a text representation. This is unreadable.

schrej (changes requested):

resources/scripts/components/server/logs/logsContainer.tsx
Tdash2: > The third thing is the spaghetti code in resources/scripts/components/server/logs/logsContainer.tsx.

I am assuming this is what Jelcoo was talking about.

FoksVHox:
This should really be superseeded by #3856

Tdash2:
> This should really be superseeded by #3856

Yes It will. I will just close this request. Thanks for all the input

PR #3845, opened 2022-01-04
Add PID Modal

### LINKED ISSUE #3844: Egg feature modal for pid limit

Create a new egg feature modal to detect `pthread_create failed` and `java.lang.outofmemoryerror: unable to create new native thread` in the console and trigger modal informing the user to contact the administrator regarding max pid limit or memory is reached.

image

Differentiate between root_admin and normal user to provide an appropriate message. root_admin should receive information to increase the pid limit, such as ```The server has reached the maximum process or memory limit. Please increase the `container_pid_limit` in the Wings `config.yml` file by default located at `/etc/pterodactyl`. Restart Wings and the affected server to apply these new limits.```

Regular users should have a generic message regarding the server attempting to use more resources than allocated and contact the administrator.

_No response_

notAreYouScared:
Adds user (left) and admin (right) error messages.

image

ps. I like the triangle :)

Closes #3844

### DIFF
--- a/resources/scripts/components/server/features/PIDLimitModalFeature.tsx
+++ b/resources/scripts/components/server/features/PIDLimitModalFeature.tsx
@@ -71,7 +71,7 @@ const PIDLimitModalFeature = () => {
                         <FontAwesomeIcon css={tw`pr-4`} icon={faExclamationTriangle} color={'orange'} size={'4x'}/>
                         <h2 css={tw`text-2xl mb-4 text-neutral-100`}>Possible resource limit reached...</h2>
                     </div>
-                    <p css={tw`mt-4`}>This server is attempting to use more resources then allocated. Please contact the administrator and give them the error below.</p>
+                    <p css={tw`mt-4`}>This server is attempting to use more resources than allocated. Please contact the administrator and give them the error below.</p>
                     <p css={tw`mt-4`}><code css={tw`font-mono bg-neutral-900`}>pthread_create failed, Possibly out of memory or process/resource limits reached</code></p>
                     <div css={tw`mt-8 sm:flex items-center justify-end`}>
                         <Button onClick={() => setVisible(false)} css={tw`w-full sm:w-auto border-transparent`}>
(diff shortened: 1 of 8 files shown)


resources/scripts/components/server/features/PIDLimitModalFeature.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <Modal visible={visible} onDismissed={() => setVisible(false)} closeOnBackground={false} showSpinnerOverlay={loading}>
+            <FlashMessageRender key={'feature:pidLimit'} css={tw`mb-4`} />
+            {isAdmin ?
+                <>
+                    <div css={tw`mt-4 sm:flex items-center`}>
+                        <FontAwesomeIcon css={tw`pr-4`} icon={faExclamationTriangle} color={'orange'} size={'4x'}/>
+                        <h2 css={tw`text-2xl mb-4 text-neutral-100 `}>Memory or process limit reached...</h2>
+                    </div>
+                    <p css={tw`mt-4`}>The server has reached the maximum process or memory limit.</p>
+                    <p css={tw`mt-4`}>Please increase the <code css={tw`font-mono bg-neutral-900`}>container_pid_limit</code> in the Wings <code css={tw`font-mono bg-neutral-900`}>config.yml</code> file, located at <code css={tw`font-mono bg-neutral-900`}>/etc/pterodactyl</code>, by default.</p>
```
matthewpi: This message should be worded to let them know the config option exists and that increasing the value may help resolve the issue, rather than telling them to change it.

notAreYouScared: It's not in question if it `may help resolve the issue`, It will resolve the issue, as that's the solution... Or do we just tell them to reduce the plugins / mods / whatever it maybe causing it to use more then 512 processes

matthewpi: > It's not in question if it `may help resolve the issue`, It will resolve the issue, as that's the solution... Or do we just tell them to reduce the plugins / mods / whatever it maybe causing it to use more then 512 processes

That's assuming there isn't an underlying problem causing that amount of processes to be created, while you are right it will resolve the problem, it may not be something you will want to change.

Also, didn't we change the default to `1024`?

trenutoo: > > It's not in question if it `may help resolve the issue`, It will resolve the issue, as that's the solution... Or do we just tell them to reduce the plugins / mods / whatever it maybe causing it to use more then 512 processes
> 
> That's assuming there isn't an underlying problem causing that amount of processes to be created, while you are right it will resolve the problem, it may not be something you will want to change.
> 
> Also, didn't we change the default to `1024`?

Default was 256 and later bumped to 512.

matthewpi (approved):

PR #3854, opened 2022-01-06
Add new audit log types

Jelcoo:
I added new logging types to cover almost all server actions.
I changed the way the old logging worked by migrating the backup uuid to backup name so the user UI can be constructed with more ease, without having to execute hundreds or possibly thousands of databases query's. With this comes a nice, completely functional, database migration file.

### DIFF
--- a/app/Http/Controllers/Api/Client/Servers/ScheduleTaskController.php
+++ b/app/Http/Controllers/Api/Client/Servers/ScheduleTaskController.php
@@ -59,7 +59,7 @@ class ScheduleTaskController extends ClientApiController
         $lastTask = $schedule->tasks()->orderByDesc('sequence_id')->first();
 
         /** @var \Pterodactyl\Models\Task $task */
-        $task = $server->audit(AuditLog::SERVER__SCHEDULE_TASK_CREATE, function (AuditLog $audit, Server $server) use ($request, $schedule) {
+        $task = $server->audit(AuditLog::SERVER__SCHEDULE_TASK_CREATE, function (AuditLog $audit, Server $server) use ($request, $schedule, $lastTask) {
             $task = $this->repository->create([
                 'schedule_id' => $schedule->id,
                 'sequence_id' => ($lastTask->sequence_id ?? 0) + 1,
--- a/app/Http/Controllers/Api/Client/Servers/ScheduleTaskController.php
+++ b/app/Http/Controllers/Api/Client/Servers/ScheduleTaskController.php
@@ -6,6 +6,7 @@ use Pterodactyl\Models\Task;
 use Illuminate\Http\Response;
 use Pterodactyl\Models\Server;
 use Pterodactyl\Models\Schedule;
+use Pterodactyl\Models\AuditLog;
 use Illuminate\Http\JsonResponse;
 use Pterodactyl\Models\Permission;
 use Pterodactyl\Repositories\Eloquent\TaskRepository;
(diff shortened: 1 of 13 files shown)


app/Models/AuditLog.php
```diff
@@ -24,23 +24,48 @@ class AuditLog extends Model
...
 
+    public const SERVER__BACKUP_START = 'server:backup.start';
+    public const SERVER__BACKUP_FAIL = 'server:backup.fail';
+    public const SERVER__BACKUP_COMPLETE = 'server:backup.complete';
+    public const SERVER__BACKUP_DELETE = 'server:backup.delete';
+    public const SERVER__BACKUP_DOWNLOAD = 'server:backup.download';
+    public const SERVER__BACKUP_LOCK = 'server:backup.lock';
+    public const SERVER__BACKUP_UNLOCK = 'server:backup.unlock';
+    public const SERVER__BACKUP_RESTORE_START = 'server:backup.restore.start';
+    public const SERVER__BACKUP_RESTORE_COMPLET = 'server:backup.restore.complete';
```
hwalker928: Missing E

app/Http/Controllers/Api/Client/Servers/ScheduleTaskController.php
```diff
@@ -58,14 +59,18 @@ public function store(StoreTaskRequest $request, Server $server, Schedule $sched
...
         /** @var \Pterodactyl\Models\Task $task */
-        $task = $this->repository->create([
-            'schedule_id' => $schedule->id,
-            'sequence_id' => ($lastTask->sequence_id ?? 0) + 1,
-            'action' => $request->input('action'),
-            'payload' => $request->input('payload') ?? '',
-            'time_offset' => $request->input('time_offset'),
-            'continue_on_failure' => (bool) $request->input('continue_on_failure'),
-        ]);
+        $task = $server->audit(AuditLog::SERVER__SCHEDULE_TASK_CREATE, function (AuditLog $audit, Server $server) use ($request, $schedule) {
```
itsgatto: $lastTask also needs to be passed to the closure, otherwise all tasks will have a sequence_id of 1.

app/Models/AuditLog.php
```diff
@@ -24,23 +24,48 @@ class AuditLog extends Model
...
+    public const SERVER__SCHEDULE_RUN = 'server:schedule.run';
+    public const SERVER__SCHEDULE_TASK_CREATE = 'server:schedule.task.create';
+    public const SERVER__SCHEDULE_TASK_UPDATE = 'server:schedule.task.update';
+    public const SERVER__SCHEDULE_TASK_DELETE = 'server:schedule.task.delete';
+
+    public const SERVER__SETTINGS_NAME = 'server:settings.name.update';
+    public const SERVER__SETTINGS_REINSTALL = 'server:settings.reinstall';
+    public const SERVER__SETTINGS_IMAGE = 'server:settings.image.update';
+
+    public const SERVER__SUBUSER_CREATED = 'server:subuser.create';
```
itsgatto: This should be SERVER__SUBUSER_CREATE so it matches the others.

app/Models/Backup.php
```diff
@@ -105,8 +105,8 @@ public function server()
      */
     public function audits()
     {
-        return $this->hasMany(AuditLog::class, 'metadata->backup_uuid', 'uuid')
+        return $this->hasMany(AuditLog::class, 'metadata->backup_name', 'name')
             ->where('action', 'LIKE', 'server:backup.%');
-        // ->where('metadata->backup_uuid', $this->uuid);
+        // ->where('metadata->backup_name', $this->name);
     }
```
itsgatto: I think you shouldn't be pulling the audits relationship based on the backup name since it's not unique. There's a chance you may pull audit logs for other backups (potentially on other servers) with the same name. 

A better approach could be to save both the backup uuid and name in the metadata, then use the uuid for anything internal and the name for the UI.

itsgatto (commented):
I'm leaving some comments about the issues I mentioned on Discord the other day. :)

DAThosting:
Maybe adding also user actions like login/logout or last login.

DaneEveritt:
Thank you for taking the time to do this, but I'm still on the fence about my audit logging implementation, and it needs more testing from me before I'd feel good about moving forwards with it.

As a result, and also per the contribution changes I am going to close out this PR.

PR #3864, opened 2022-01-08
Add view console as a seperate permission in subusers

### LINKED ISSUE #3466: Add view console as a seperate permission in subusers.
Well, I have a moderator team. I want them to have the ability to start the server if it crashes while I'm asleep. However, if I give them the start power action, they will have permissions to view the console. That contains sensitive info, although they can't type commands, they can still collect IP addresses and it's just not a good idea to show mods this

I'd like to have a new permission called "View Console" that can give people permission to view it or not. Giving people the start/stop/kill permissions will only give them access to the buttons, but to now view the console itself, unless they have the view console permission

Only possible alternative I have considered is trying to use discord bot along with pterodactyl API to make a simple .start command for the mods to do this. But it'd be neat and better to do it on the panel itself!

**Additional context**

Jelcoo:
Closes #3466

### DIFF
--- a/app/Models/Permission.php
+++ b/app/Models/Permission.php
@@ -121,8 +121,8 @@ class Permission extends Model
         'control' => [
             'description' => 'Permissions that control a user\'s ability to control the power state of a server, or send commands.',
             'keys' => [
-                'sendconsole' => 'Allows a user to send commands to the server instance via the console.',
-                'readconsole' => 'Allows a user to read the console.',
+                'send-console' => 'Allows a user to send commands to the server instance via the console.',
+                'read-console' => 'Allows a user to read the console.',
                 'start' => 'Allows a user to start the server if it is stopped.',
                 'stop' => 'Allows a user to stop a server if it is running.',
                 'restart' => 'Allows a user to perform a server restart. This allows them to start the server if it is offline, but not put the server in a completely stopped state.',
--- a/app/Models/Permission.php
+++ b/app/Models/Permission.php
@@ -16,7 +16,8 @@ class Permission extends Model
      * Constants defining different permissions available.
      */
     public const ACTION_WEBSOCKET_CONNECT = 'websocket.connect';
-    public const ACTION_CONTROL_CONSOLE = 'control.console';
+    public const ACTION_CONTROL_CONSOLE_SEND = 'control.sendconsole';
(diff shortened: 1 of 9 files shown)


app/Models/Permission.php
```diff
@@ -16,7 +16,8 @@ class Permission extends Model
      * Constants defining different permissions available.
      */
     public const ACTION_WEBSOCKET_CONNECT = 'websocket.connect';
-    public const ACTION_CONTROL_CONSOLE = 'control.console';
+    public const ACTION_CONTROL_CONSOLE_SEND = 'control.sendconsole';
+    public const ACTION_CONTROL_CONSOLE_READ = 'control.sendconsole';
```
Boy132, line 20: That should be `control.readconsole`. :)

app/Models/Permission.php
```diff
@@ -16,7 +16,8 @@ class Permission extends Model
      * Constants defining different permissions available.
      */
     public const ACTION_WEBSOCKET_CONNECT = 'websocket.connect';
-    public const ACTION_CONTROL_CONSOLE = 'control.console';
+    public const ACTION_CONTROL_CONSOLE_SEND = 'control.sendconsole';
```
matthewpi: Use dashes when multiple words are being used rather than putting them together.  `control.send-console` for example.

matthewpi (changes requested):

app/Models/Permission.php
```diff
@@ -120,7 +121,8 @@ class Permission extends Model
         'control' => [
             'description' => 'Permissions that control a user\'s ability to control the power state of a server, or send commands.',
             'keys' => [
-                'console' => 'Allows a user to send commands to the server instance via the console.',
+                'sendconsole' => 'Allows a user to send commands to the server instance via the console.',
+                'readconsole' => 'Allows a user to read the console.',
```
matthewpi: You missed adding the `-` in these

matthewpi (changes requested):

matthewpi (changes requested):
Marking this as requesting changes so it doesn't get merged.  Adding/changing these permissions also requires changing them on Wings.

DaneEveritt:
Closing as this is both a breaking permissions change and requires updates to Wings that are not present.

PR #3870, opened 2022-01-10
Fixed error in Build Configuration

mrxbox98:
The server editing menu incorrectly stated that 100% means 1 physical core when it really means 1 virtual core (thread).


Boy132:
That will most likely be fixed with the admin rewrite in v2.
Also afaik PR's for the admin area aren't currently accepted because of said rewrite.

PR #3879, opened 2022-01-12
Fix request types

Jelcoo:
Added missing backup request types to match all the other controllers


DaneEveritt:
Thanks, but this was intentional to avoid flooding that directory with tons of request classes that do nothing of value. There is a decent amount of inconsistency, but that is because I slowly update code as time allows, rather than refactoring entire chunks of work at the same time. 👍🏼

PR #3902, opened 2022-01-21
Delete backups when server deleted

### LINKED ISSUE #2564: Backups Removed on Server Deletion
When a server is deleted/terminated, the backups remain. This is when the S3 filesystem is configured.

It would be nice to have all backups removed when a server is deleted. For those of us using S3 or off-site filesystems, this will save on storage costs.

n/a

**Additional context**
I have only configured my install using S3 backup filesystems. I am not sure if this issue is present using local storage.

Jelcoo:
Deletes all backups of a server when the server is deleted.

Closes #2564

### DIFF
--- a/app/Services/Servers/ServerDeletionService.php
+++ b/app/Services/Servers/ServerDeletionService.php
@@ -76,6 +76,18 @@ class ServerDeletionService
      */
     public function handle(Server $server)
     {
+        try {
+            $this->daemonServerRepository->setServer($server)->delete();
+        } catch (DaemonConnectionException $exception) {
+            // If there is an error not caused a 404 error and this isn't a forced delete,
+            // go ahead and bail out. We specifically ignore a 404 since that can be assumed
+            // to be a safe error, meaning the server doesn't exist at all on Wings so there
+            // is no reason we need to bail out from that.
+            if (!$this->force && $exception->getStatusCode() !== Response::HTTP_NOT_FOUND) {
+                throw $exception;
+            }
+        }
+
         $this->connection->transaction(function () use ($server) {
             foreach ($server->backups as $backup) {
                 // Unlock backup to prevent BackupLockedException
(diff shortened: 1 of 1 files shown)


yesBad:
That isnt how backups work but ok

Jelcoo:
> That isnt how backups work but ok

How should it work then?

yesBad:
Arent backups supposed to stay... Thats kinda how backups work? So you can get them back if u need em?

Jelcoo:
First, that doesn't explain why the issue is still open, and second the backups are for the user, not for the admin. If you want a backup of servers as an admin, you want to invest in external node backups.

matthewpi:
> Arent backups supposed to stay... Thats kinda how backups work? So you can get them back if u need em?

There is an open issue about this that hasn't been closed and it would if we thought that this would be bad behaviour.  Backups are created by end-users, not Panel administrators and should be deleted when a server is.  The backups created by Pterodactyl should not be relied on for disaster recovery or for protecting against accidental deletions.  The references of the backups are deleted when a server is due to them using a foreign key (if I remember correctly) so an admin would have to go manually find and restore a backup, this is made worse by local backups not being stored in a parent server id folder and just loosely thrown into the `backups` directory.

While there might be a better way to handle this in terms of having a global config option, request parameter, etc.  Making this change is fine, however a discussion about how or if it should be implemented should take place on the [issue](https://github.com/pterodactyl/panel/issues/2564), not this PR.

yesBad:
Understandable. Also yeah the backups are stored pretty randomly :p

app/Services/Servers/ServerDeletionService.php
```diff
@@ -68,29 +76,33 @@ public function withForce($bool = true)
      */
     public function handle(Server $server)
     {
-        try {
```
trenutoo: This DaemonConnectionException handler should not be deleted and also is why all the tests fail

DaneEveritt:
@Jelcoo I'm good to accept this PR, but not with this current logic. If a server has say 20+ backups, the deletion process is going to be quite slow. Can we move the backup deletion off into a separate queued job that can run and delete them in the background? That way they get removed, but with less priority and if a backup deletion fails just log the error and move on.

snaildos:
What happens to servers that have already been deleted, backups will just sit their forever?
Is there a way we can trigger some sort of cleanup job?

Jelcoo:
No intention of going through the process of doing this

Issue #3712, opened 2021-10-27
Add support for interactive programs to server console


anondev-sudo:
### Is there an existing feature request for this?

- [X] I have searched the existing issues before opening this feature request.

### Describe the feature you would like to see.

Add way to run command in console when another command is running and allow to use keyboard keys when you click on the console.

### Describe the solution you'd like.

When you run some command in console, you can't stop it, as you can't use CTRL + C in the console.

Allow us to do so, and to run any command while another command is running.
Also you can't use F keys on keyboard in pterodactyl console.

notAreYouScared:
Why would you need to stop a command? The console is basically forwarding the commands to whatever program is running in the startup.

Jelcoo:
There is litterly a `stop` button next to the console

anondev-sudo:
Ik, but you can't use keyboard keys in console. Like normally when you write for example `node index.js` to console and then you want to stop the process, you can press CTRL + C keys on the keyboard. But inside pterodactyl console you can't do this.

Also you can't use arrow keys in pterodactyl console, so you can't make choice prompts using Inquirer.js

Jelcoo:
You can use arrow key up and down to cycle through commands you used before
<img src="https://cdn.jelco.xyz/biVi1/NenoqeXO63.gif/raw">

anondev-sudo:
Ok, but you can't use keyboard keys like F keys in console. And you can't use CTRL + letter in console too

Jelcoo:
Pterodactyl Software is not designed to be interactive, it is designed to be easy to use. If you want to use keybinds, I recommend using a local terminal.

DaneEveritt:
Correct, but @anondev-sudo is making a completely valid point that you cannot run interactive programs using the console right now, and there are a lot of programs you can run in Pterodactyl that would benefit from that.

anondev-sudo:
You can't connect to pterodactyl container using ssh, so users can't do that. If it will be possible then it will solve the issue or make the console more interactive

Marfjeh:
Maybe an idea to use xterm.js as a console?

schrej:
@Marfjeh It's already using xterm.js: https://github.com/pterodactyl/panel/blob/develop/resources/scripts/components/server/Console.tsx

Marfjeh:
> @Marfjeh It's already using xterm.js: https://github.com/pterodactyl/panel/blob/develop/resources/scripts/components/server/Console.tsx

[...]

PR #3904, opened 2022-01-22
Popup Console

### LINKED ISSUE #2584: Button to open console in a separate window (Like in <1.0.x)
A button that opens separate window with server's console.

**Additional context**
It was a really useful feature for server management. I believe it was just overlooked so this is just a reminder.

mrxbox98:
This commit adds a popup console. Currently the console opens in a new tab, but I can make it a new window if needed. The code also might have a lot of unnecessary checks as I was testing it for a while. This has been tested on desktop and works.
This closes https://github.com/pterodactyl/panel/issues/2584

Preview:


Rezanans-wow:
Nice

camwhi-te:
A nice addition would be to make the button change depending on whether the console is in fullscreen mode or not. For example, if the console is fullscreen, the button should change to an X or an arrow which links to the main console page for easy navigation. (Also, side note, this needs to be eslinted)

dtrasca:
any comment on how to use this ? I changed the 3 files (Console, ServerConsole and ServerRouter) and restarted nginx, but Im not good at it as you can see. Any help ? x

mrxbox98:
> any comment on how to use this ? I changed the 3 files (Console, ServerConsole and ServerRouter) and restarted nginx, but Im not good at it as you can see. Any help ? x

I suggest waiting until this is merged, but if you apply the patch file you can run ``yarn build:production`` to use it.

mrxbox98:
Failing check irrelevant to this commit.

camwhi-te:
Having the console pop-out in a new window might be the best approach. It would be cleaner while also making it an isolated window instead of another 'page'.

mrxbox98:
> Having the console pop-out in a new window might be the best approach. It would be cleaner while also making it an isolated window instead of another 'page'.

https://github.com/pterodactyl/panel/pull/3904/commits/f3a402445e3cbb499f3728d8248552ce808ecb41 should do this

resources/scripts/components/server/Console.tsx
```diff
@@ -211,12 +215,12 @@ export default () => {
                 ]}
                 style={{ minHeight: '16rem' }}
             >
-                <TerminalDiv id={'terminal'} ref={ref} />
+                <TerminalDiv style={window.location.href.includes('/console') ? { height: 'calc(100% - 54px)' } : {}} id={'terminal'} ref={ref} />
```
OpenSource03: Please just use inline if-else inside the height property instead of defining it inside the whole style block

resources/scripts/components/server/Console.tsx
```diff
@@ -201,8 +201,12 @@ export default () => {
         };
     }, [ connected, instance ]);
 
+    function openConsole () {
```
OpenSource03: Use const instead of a function to better match the overall code-style of this file.

resources/scripts/routers/ServerRouter.tsx
```diff
@@ -90,50 +90,60 @@ const ServerRouter = ({ match, location }: RouteComponentProps<{ id: string }>)
 
     return (
         <React.Fragment key={'server-router'}>
-            <NavigationBar/>
+            {window.location.href.includes('/console') ?
+                <>
+                </>
+                :
+                <NavigationBar/>
```
OpenSource03: Replace with `!window.location.href.includes('/console') && <NavigationBar/>`

resources/scripts/routers/ServerRouter.tsx
```diff
@@ -90,50 +90,60 @@ const ServerRouter = ({ match, location }: RouteComponentProps<{ id: string }>)
...
-                                {rootAdmin &&
-                                <a href={'/admin/servers/view/' + serverId} rel="noreferrer" target={'_blank'}>
-                                    <FontAwesomeIcon icon={faExternalLinkAlt}/>
-                                </a>
-                                }
-                            </div>
-                        </SubNavigation>
-                    </CSSTransition>
+                    {window.location.href.includes('/console') ?
+                        <></>
```
OpenSource03: Replace with `!window.location.href.includes('/console') && <>...</>`

mrxbox98: Fixed in https://github.com/pterodactyl/panel/pull/3904/commits/b1ed950cff39e1bbe3ed53e697fe34a9b91e7eb3

resources/scripts/routers/ServerRouter.tsx
mrxbox98: Fixed in https://github.com/pterodactyl/panel/pull/3904/commits/b1ed950cff39e1bbe3ed53e697fe34a9b91e7eb3

resources/scripts/components/server/Console.tsx
mrxbox98: Fixed in https://github.com/pterodactyl/panel/pull/3904/commits/b1ed950cff39e1bbe3ed53e697fe34a9b91e7eb3

resources/scripts/components/server/Console.tsx
mrxbox98: Fixed in https://github.com/pterodactyl/panel/pull/3904/commits/b1ed950cff39e1bbe3ed53e697fe34a9b91e7eb3

resources/scripts/components/server/Console.tsx
```diff
@@ -201,8 +201,12 @@ export default () => {
         };
     }, [ connected, instance ]);
 
+    const openConsole = () => {
+        window.open(window.location.href + '/console', 'Server Console', 'height=1000,width=1500');
+    };
+
     return (
-        <div css={tw`text-xs font-mono relative`}>
+        <div css={window.location.href.includes('/console') ? tw`text-xs font-mono relative w-full h-screen overflow-hidden` : tw`text-xs font-mono relative`}>
```
DaneEveritt: Can you use the react router helpers to check the route rather than the window location item?

resources/scripts/components/server/Console.tsx
```diff
@@ -225,6 +229,11 @@ export default () => {
                             onKeyDown={handleCommandKeyDown}
                         />
                     </div>
+                    <button css={tw`m-auto mr-2 inline-block`} hidden={window.location.href.includes('/console')} onClick={openConsole}>
+                        <svg xmlns="http://www.w3.org/2000/svg" css={tw`m-auto w-6 h-6`} fill="none" viewBox="0 0 24 24" stroke="currentColor">
```
DaneEveritt: This should be using the heroicons package, not the raw SVG.

resources/scripts/components/server/Console.tsx
```diff
@@ -201,8 +201,12 @@ export default () => {
         };
     }, [ connected, instance ]);
 
+    const openConsole = () => {
+        window.open(window.location.href + '/console', 'Server Console', 'height=1000,width=1500');
```
DaneEveritt: Can you do a smaller window size, this is massive.

Pick whatever seems best to you:
a.) 640x whatever the other dimension that normally goes with this is.
b.) Whatever the old console size was in 0.7
c.) Some scaling based on the detected resolution of the user. Like, 1/4 of the scale, or 1/3.

resources/scripts/components/server/ServerConsole.tsx
```diff
@@ -21,50 +21,67 @@ const ServerConsole = () => {
...
-                {isInstalling ?
-                    <div css={tw`mt-4 rounded bg-yellow-500 p-3`}>
-                        <ContentContainer>
-                            <p css={tw`text-sm text-yellow-900`}>
-                                This server is currently running its installation process and most actions are
-                                unavailable.
-                            </p>
-                        </ContentContainer>
+        <ServerContentBlock title={'Console'} css={fullscreen ? tw`absolute h-screen w-screen left-0 top-0 px-0 mt-0` : tw`flex flex-wrap`}>
+            {fullscreen ?
```
DaneEveritt: This code is incorrect, you've disabled all of the features such that a popped out console won't correctly alert the user to issues.

It might make sense to create a new component here that has the console and console feature stuff, and just render it twice on this screen, alone if fullscreen, or inside the correct block location otherwise.

DaneEveritt: Or just use some CSS to make the console "full-screen" without hiding other elements, something like an absolutely position element using a portal so that it is overlaying everything, and just make the portal and styling deterministic based on full screen.

Basically you have a lot of very specific console logic spread out over files that really shouldn't contain that.

DaneEveritt (changes requested):

anondev-sudo:
This is really good I think, maybe remove the enter command field too and make that it will display cursor and that every key will be forwarded there. It can solve https://github.com/pterodactyl/panel/issues/3712

For example in xterm.js demo, when you press CTRL + C key it is sent there. https://xtermjs.org/
image
On pterodactyl console you can't do this. I think that type a command field should be removed and rather use normal cursor and forward keys to concols when you have clicked with mouse on the console.

resources/scripts/components/server/Console.tsx
mrxbox98: The project currently doesn't have the HeroIcons package. Should I add it?

resources/scripts/components/server/Console.tsx
KaspianDev: 640x480 should look perfectly

mrxbox98:
I havent worked on this pr in a while, [here](https://github.com/pterodactyl/panel/pull/4132) is better version from VinGal0

mrxbox98:
Accidentally closed this one too not sure what git and github are doing but apparently this is the only way for me to update my PRs

mrxbox98:
There may be some unintended side effects with this PR since I spent an hour trying to fix git

robertdrakedennis:

PR #3906, opened 2022-01-23
Added a new environment variable

alexevladgabriel:
The commit add a new environment variable, what will add ability to create the feature called FastDownload required for Source Games for panel-side. I'm kinda exhausted to paste this variable every time, when I'm installing the FastDL plugin.


trenutoo:
I don't see any use for this in Pterodactyl itself. Pterodactyl panel doesn't ship any Fast DL features. 

Should you have third-party addons relying on code modifications, that's something to be applied when installing those addons.

PR #3916, opened 2022-01-25
feat(docker): automated certificate renewal with certbot

### LINKED ISSUE #3907: Panel in Docker - Certbot auto-renew not working

### Current Behavior

A little while ago, my panel stopped working due to an expired SSL certificate. I had to recreate the docker container to get it working again. After a quick test, I found out that the renew command fails (inside the container) because it can't bind to port 80. I'm guessing this is the reason auto-renewal isn't working.

### Expected Behavior

The docker image should either use the [Nginx integration](https://eff-certbot.readthedocs.io/en/stable/using.html?highlight=renew#nginx) or [webroot](https://eff-certbot.readthedocs.io/en/stable/using.html?highlight=renew#webroot), so certbot can renew certificates while Nginx is running. This way, SSL certificates get automatically renewed by certbot inside the panel container.

### Steps to Reproduce

Run the panel inside a Docker container using the official image. Leave it running until the certificate expires. You now get an error when visiting the panel from your browser.

### Panel Version

1.7.0

### Wings Version

1.5.5

### Error Logs

_No response_

trenutoo:
Executes cron daily at 23:00 to reduce the load on LE when everyone else is doing it at 00. A certificate that expires in less than 30 days will be renewed.

closes #3907

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ RUN apk add --no-cache --update ca-certificates dcron curl git supervisor tar un
 
 RUN rm /usr/local/etc/php-fpm.conf \
     && echo "* * * * * /usr/local/bin/php /app/artisan schedule:run >> /dev/null 2>&1" >> /var/spool/cron/crontabs/root \
-    && echo "0 23 1 * * certbot renew --nginx --quiet" >> /var/spool/cron/crontabs/root \
+    && echo "0 23 * * * certbot renew --nginx --quiet" >> /var/spool/cron/crontabs/root \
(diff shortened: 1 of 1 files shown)


DaneEveritt:
I feel like everything I've ever seen just runs renew each night and the command is smart enough to know if it should re-attempt or not. Otherwise you might have way less than a month remaining before expiration depending on when the cert was created.

trenutoo:
Indeed the renew command should prevent spamming to cause rate-limiting, so I guess it doesn't hurt to do it daily.

My logic with this was the certificate being renewed whenever a container is created, so the cron would have run by then should a container run long enough for this to be an issue.

parkervcp (approved):
LGTM

PR #3982, opened 2022-02-23
Go to configuration page instead of allocations after creating a node

JohnB17:
When a node is created, you get redirected to the allocations page, but the next thing to do is get the configuration from the configuration tab. This pull request changes it so instead of the allocations page it goes to the configuration page.


DaneEveritt:
This is intentional, most people don't understand they need to also add allocations. We've also rewritten the entirety of the admin area for V2 and are only accepting security fixes to it.

PR #2888, opened 2020-12-18
Implementing i18n

fabm3n:
This is currently a WIP Implementation for basic i18n in account and auth components.
There won't be any language switching in this PR.

- [x] Implement (and / or update) the 18n backend and dependencies
- [x] Implement i18n in all account and auth components
- [x] resources/scripts/routers/DashboardRouter.tsx

Files i need help:
- [ ] resources/scripts/components/elements/ScreenBlock.tsx

Things to do:
- [ ] Remove Trans component. From the docs: "It does ONLY interpolation. It does not rerender on language change or load any translations needed."

I will update this post if more progress is done.

This will not close #2518 because this is only a small step for i18n.

PR #3992, opened 2022-03-07
Remove old language feature

trgaming:
Currently I add a translation feature to the Panel.

To begin I deleted the old feature.

In no time I'm going to add the feature first for the Admin-Panel

Later I add the feature for the User UI.


Jelcoo:
This is already going to be a thing
https://github.com/pterodactyl/panel/pull/2888

Jelcoo:
And editing the admin UI doesn't make sense because it is going to get replaced soon

DaneEveritt:
Thank you, but we do not accept PRs for anything other than bug fixes.

PR #3994, opened 2022-03-09
Update: Garry's Mod now has x64 server support

ZEROPOINTBRUH:
Servers can perform a lot smoother than traditional x86 servers, using a lot less CPU usage, and can allocate a lot more memory than the x86. Making it easier to run off of fewer resources than a traditional Garry's mod server.

### DIFF
--- a/database/Seeders/eggs/source-engine/egg-garrys-mod.json
+++ b/database/Seeders/eggs/source-engine/egg-garrys-mod.json
@@ -4,10 +4,10 @@
         "version": "PTDL_v1",
         "update_url": null
     },
-    "exported_at": "2022-01-18T07:04:20-05:00",
-    "name": "Garrys Mod",
-    "author": "support@pterodactyl.io",
-    "description": "Garrys Mod, is a sandbox physics game created by Garry Newman, and developed by his company, Facepunch Studios.",
+    "exported_at": "2022-03-09T13:08:34-07:00",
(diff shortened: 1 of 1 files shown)


parkervcp:
I would say that specifying x64 'could' be bad if someone isn't expecting it. Hosts can always update their start commands on their own if they want to optimize for that.

ZEROPOINTBRUH:
> I would say that specifying x64 'could' be bad if someone isn't expecting it. Hosts can always update their start commands on their own if they want to optimize for that.

What if we made it an option to choose from x86 or x64?

Boy132:
Garry's Mod does _not_ have 64 bit support (yet). 64 bit support is currently only available on the beta/ in-dev branch which means that everyone that would use this updated egg would be forced on the beta branch - with 64 bit support but also many other in-dev features that might cause other problems. So I totally agree with that parker said above.

trenutoo:
This is not how you accomplish this, since you're forcing dev branch on all users and the Docker image entrypoint will overwrite this once it runs a steamcmd update on boot without specifying dev branch. We have existing flags to install dev branches, and [there's an egg](https://github.com/parkervcp/eggs/blob/master/stock_eggs/source-engine/egg-garrys-mod.json) on the egg repo that added this optional functionality using the correct variables, although it wasn't added upstream to this repo.

PR #3998, opened 2022-03-11
fix(paper): fetching older versions

trenutoo:
Limit the results to a single match to prevent results such as "true true true" since the current version check expects only a single match


tmunsch:
Nice! Good work!

PR #4034, opened 2022-04-04
Auto Mounts Feature

alexevladgabriel:
This PR adds mounts to the installation script on Panel.
People will be able to activate the option from mounts page.
This feature enable people with low bandwidth to create a mount where will contain all the binaries required
for server installation. The feature will save time on installation & network bandwidth.
Right now the database has two new boolean values (mount_on_install, auto_mount).
The mount for installation script is deleted after server installation event.
Auto mount boolean column should be used to auto mount on server creation and remain available until the server is deleted.

### DIFF
--- a/app/Listeners/RemoveMountsAfterInstall.php
+++ b/app/Listeners/RemoveMountsAfterInstall.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Pterodactyl\Listeners;
+
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Queue\InteractsWithQueue;
+use Pterodactyl\Models\Mount;
+use Pterodactyl\Models\MountServer;
+
+class RemoveMountsAfterInstall
+{
+    /**
+     * Create the event listener.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        //
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


Rezanans-wow:
Your THIS pull request need changes for wings too? https://github.com/pterodactyl/wings/pull/125
Because you not write nothing about that

alexevladgabriel:
> Your THIS pull request need changes for wings too? [pterodactyl/wings#125](https://github.com/pterodactyl/wings/pull/125) Because you not write nothing about that

Yes, the wings modification you tagged are necessary to work mounting on installation script. I wanted to detail more on panel side about the future.

fragger1991:
@alexevladgabriel i have imported and tested your changes. Can you explain what the new options are supposed to do exactly and when they are triggered or what their status should be?

The directory is properly mounted during installation, otherwise the directory is not mounted. I activated both options. I would have expected that with the "Auto Mount" option, the directory would be and stay permanently mounted.

Rezanans-wow:
Hello sir, auto-mount option and on install isnt work

alexevladgabriel:
@Rezanans-wow I will update soon to latest version. Stay tuned.

alexevladgabriel:
image
image

Update soon :tm:

Issue #4047, opened 2022-04-14
Being able to get config of node from CLI (with feature code)


Hanarion:
### Is there an existing feature request for this?

- [X] I have searched the existing issues before opening this feature request.

### Describe the feature you would like to see.

I would like the possiblity to get the config of a node directly by CLI, so I added the artisan command

### Describe the solution you'd like.

`cat app/Console/Commands/Node/GetConfigNodeCommand.php `
```php
<?php
namespace Pterodactyl\Console\Commands\Node;

use Illuminate\Console\Command;
use Pterodactyl\Contracts\Repository\NodeRepositoryInterface;

class GetConfigNodeCommand extends Command
{

    /**
     * @var string
     */
    protected $description = 'Get a node configuration from the Panel.';

    /**
     * @var \Illuminate\Support\Collection
     */
    protected $nodes;

    /**
     * @var \Pterodactyl\Contracts\Repository\NodeRepositoryInterface
     */
    protected $repository;

    /**
     * @var string
     */
    protected $signature = 'p:node:config {--id= : The id of the node.}';

    /**
     * GetConfigNodeCommand constructor.
     */
    public function __construct(
        NodeRepositoryInterface $repository
    ) {
        parent::__construct();

        $this->repository = $repository;
    }

    /**
     * Respond to the command request.
     *
     * @throws \Pterodactyl\Exceptions\Repository\RecordNotFoundException
     */
    public function handle()
    {
        $this->nodes = $this->nodes ?? $this->repository->all();
        $id = $this->option('id') ?? $this->anticipate(
            "Enter a valid id",
            $this->nodes->pluck('id')->toArray()
        );

        $node = $this->nodes->where('id', $id)->first();
        if (is_null($node)) {
            $this->error("Node not found");
            if ($this->input->isInteractive()) {
                $this->handle();
            }

            return;
        }

        $yaml = $node->getYamlConfiguration();
        $this->line($yaml);
    }
}
```

PR #4048, opened 2022-04-14
Create GetConfigNodeCommand.php

Hanarion:
This is relevant for the issue : https://github.com/pterodactyl/panel/issues/4047

it adds the possiblity to do this type of commands :
`php artisan p:node:config --id=1 -n > /etc/pterodactyl/config.yml`

It allows for exemple auto provisionning of a node, coupled with other artisan commands (location:make and node:make wich was just added) we can fully auto install a pterodactyl node with panel

### DIFF
--- a/app/Console/Commands/Node/GetConfigNodeCommand.php
+++ b/app/Console/Commands/Node/GetConfigNodeCommand.php
@@ -1,0 +1,67 @@
+<?php
+namespace Pterodactyl\Console\Commands\Node;
+
+use Illuminate\Console\Command;
+use Pterodactyl\Contracts\Repository\NodeRepositoryInterface;
+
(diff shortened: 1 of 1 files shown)


mundotv789123:
and can add this command
php artisan p:node:make --name='node' --description='node desc' --location='1'  --visible='1' --host='127.0.0.1' --ssl=false --port='8080' etc...
this is very important for automatic instalation scripts.

Hanarion:
@mundotv789123 actually the node make was already added a few weeks ago : https://github.com/pterodactyl/panel/blob/develop/app/Console/Commands/Node/MakeNodeCommand.php

hwalker928:
#4047 has now been implemented by Dane.

PR #4050, opened 2022-04-15
Literally just automate php artisan when using clean

expxx:
its just so people dont think ur website is broken when using clean

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
         "yarn-deduplicate": "^1.1.1"
     },
     "scripts": {
-        "clean": "cd public/assets && find . \\( -name \"*.js\" -o -name \"*.map\" \\) -type f -delete",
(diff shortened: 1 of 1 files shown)


DaneEveritt:
I don't understand how:

a.) this would actually do anything of value
b.) why it is necessary given that clean is for development environments.

expxx:
Usually, when installing addons, the devs have you run `yarn run production:build`

gOOvER:
> Usually, when installing addons, the devs have you run `yarn run production:build`

Addons are not supported by Pterodactyl  ;)

matthewpi:
All your change does is put the panel in maintenance mode, deletes the existing assets, then puts it out of maintenance mode.  Users will still see a blank screen (assuming the assets aren't cached) as the build process hasn't completed yet.

PR #4062, opened 2022-04-27
Adding make database host artisan command

mundotv789123:


Jelcoo:
This has been suggested like a thousand times and dane didn't see the need for this, so it probably won't be added

Thomas5300:
At the moment it takes the host at password:
`$data['password'] = $this->option('host') ?? $this->secret('Enter the password to the account defined');`
Solution:
`$data['password'] = $this->option('password') ?? $this->secret('Enter the password to the account defined');`

PR #4069, opened 2022-05-07
fix: artisan translations

trenutoo:
We still need this for artisan until someone hard codes them. Removed by mistake in 82818414a3e340bb3aa9ba3d29b489866a7e85eb


DaneEveritt:
Was in a silly goofy mood. 🤪

PR #4103, opened 2022-05-30
New Automatic Installer

Mubeen142:
A simplified and reliable version of the automatic installer. To run the installer type
php artisan pterodactyl:update


DaneEveritt:
Can you please clarify what you changed that makes it more reliable, and why it needed a whole new file rather than just modifying the existing one?

Of concern is that this new one doesn't properly identify the user and group — just because a good chunk of users have the defaults doesn't mean it is always accurate.

Mubeen142:
The current installer, as the site suggests, is in-operable. Personally, the automatic installer has never worked for me. I always ended up using the manual installer. 

What makes it more reliable is that it uses a simple set of commands, exact same ones provided for the manual installation and executes them. It also uses far less code. It shouldn't be hard to get the correct default values such as /var/www/pterodactyl, I can commit that when I am on pc. One thing that might be of concern is that some commands require sudo to be executed, if the user is not logged in as root users, these commands might fail to run. This however, should be relatively simple to fix

DaneEveritt:
The main issue with the current upgrader is that it updates the code that it needs to run itself. I was just going to remove it and replace it with a far simpler bash script that users can run on their own.

Mubeen142:
identifying the user and group with a bash file will be more difficult. I'm not sure how that will work.

I have made the installer more dynamic by using the base_path() function and changed it to use the previous file instead of adding a new one.

DaneEveritt:
I'm closing this out as I do not wish to continue using an upgrader script that is part of the very application being upgraded. In addition you've removed logic and make it less flexible when user/group alignment is not one of the three hard-coded cases.

PR #4138, opened 2022-06-12
Fix grammatical error

argonaut-network:
Changed "... each entities location" to "... each entity's location"; the statement is intended to convey that each entity has a location, so the possessive form should be used.

### DIFF
--- a/database/Seeders/eggs/source-engine/egg-garrys-mod.json
+++ b/database/Seeders/eggs/source-engine/egg-garrys-mod.json
@@ -87,7 +87,7 @@
         },
         {
             "name": "Tickrate",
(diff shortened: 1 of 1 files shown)


trenutoo:
The grammar correction that you've made is correct, but the modification itself is not. You shouldn't modify egg JSON files directly, because in this case, you're escaping something that shouldn't be and thus breaking the whole JSON file.

argonaut-network:
Duly noted. I viewed the JSON after editing in my own panel and I see now that it shouldn't have been escaped. I assumed safe was better than sorry.

trenutoo (approved):
LGTM. Thank you for contributing :+1:

In the future, export the egg and commit that instead. It makes sure there are no user errors

PR #4179, opened 2022-06-27
Fix: Max Ram usage showing wrong while hovering over Memory

Sharkspotone:
The multiplyer for the calculation from MB to GB is wrong. The Panel Calculates 1GB = 1000MB and Not 1GB = 1024MB.
If i set the Max Ram usage to 24.000MB the Panel shows Max 24GB but if i set it to 24.576MB the Panel shows 24,57GB as max.

Ptero Bug
Ptero Bug2

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -9,7 +9,7 @@ return [
     | change this value if you are not maintaining your own internal versions.
     */
 
-    'version' => 'canary',
+    'version' => '1.9.0',
 
     /*
     |--------------------------------------------------------------------------


WavingCatDevs:
Why is this a pull request and the changes you made were unnecessary?

superfalcon18:
Welp,
I think you should have a look on the change logs:
https://github.com/pterodactyl/panel/releases/tag/v1.9.0
It was intentional tho, i hate it too. Its a pretty lame thing to do tbh, a person who is skilled enough to install the panel knows basic computer knowledge, and that 1gb = 1024mb is the least you can know. All the old users will be confused af. Hope they revert it back.

Boy132:
> If i set the Max Ram usage to 24.000MB the Panel shows Max 24GB but if i set it to 24.576MB the Panel shows 24,57GB as max.

Yes, thats correct. 1000 MB = 1 GB. So I don't see the problem?

Also, why is this a PR and not an issue?

PR #4194, opened 2022-06-28
Enable standards mode in React app

ArnaudLier:
Adds `<!DOCTYPE html>` to the template that renders the React app to enable the standards mode (quirks mode makes a warning in the console):
https://developer.mozilla.org/en-US/docs/Web/HTML/Quirks_Mode_and_Standards_Mode
This doesn't seem to change the layout in any way.

### DIFF
--- a/resources/views/templates/wrapper.blade.php
+++ b/resources/views/templates/wrapper.blade.php
@@ -1,3 +1,4 @@
+<!DOCTYPE html>
 <html>
     <head>
         <title>{{ config('app.name', 'Pterodactyl') }}</title>


KaspianDev:
@DaneEveritt this one breaks file container for me (list items are too large)

trenutoo:
> this one breaks file container for me (list items are too large)

This has already been fixed, since the body height is not relative to the viewport in standard mode.

KaspianDev:
> > this one breaks file container for me (list items are too large)
> 
> This has already been fixed, since the body height is not relative to the viewport in standard mode.

Yeah i see, it was fixed right after my comment was made ;p

PR #4195, opened 2022-06-28
Delete backups on server deletion

### LINKED ISSUE #2564: Backups Removed on Server Deletion

Boy132:
Based on #3902, but deletion is handled via a job.

~Current problem is that wings doesn't delete the backups because the server doesn't exist anymore when the job runs. Probably a wings edit is needed? (a wings-sided job or something?) Ideas are welcome, as I'm clueless atm.~
Solved this by making the server deletion itself a job too.

This will close #2564.

### DIFF
--- a/app/Jobs/Backup/DeleteBackupJob.php
+++ b/app/Jobs/Backup/DeleteBackupJob.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Pterodactyl\Jobs\Backup;
+
+use Pterodactyl\Jobs\Job;
+use Pterodactyl\Models\Backup;
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\DispatchesJobs;
+use Pterodactyl\Services\Backups\DeleteBackupService;
+
+class DeleteBackupJob extends Job implements ShouldQueue
+{
+    use DispatchesJobs;
+    use InteractsWithQueue;
+    use SerializesModels;
+
+    /**
+     * @var \Pterodactyl\Models\Backup
+     */
+    public Backup $backup;
+
+    /**
+     * DeleteBackupJob constructor.
+     */
+    public function __construct(Backup $backup)
+    {
+        $this->queue = config('pterodactyl.queues.standard');
+        $this->backup = $backup;
+    }
+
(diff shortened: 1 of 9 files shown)


Jelcoo:
> Current problem is that wings doesn't delete the backups because the server doesn't exist anymore when the job runs.

This was my problem as well, so I closed it and never looked at it again.

xkisu:
One solution I did for this in a cloud-based setup was to store the backups in S3 storage and add a panel-side scheduled task (via an [Artisan command](https://laravel.com/docs/9.x/scheduling#scheduling-artisan-commands) so it could also be ran manually) that scans the S3 object prefixes and checks if the node and servers still exist, and then clears out the ones that don't have a matching server. I also implemented a date check so it would keep backups for a couple of days after assumed server deletion in case the deletion was accidental so it could be recovered.

Not really helpful for people using local file backups - but something similar could possibly be implemented Wings-side to allow the panel to get a list of backup files stored on Wings, check if the corresponding backup still exists and then tell Wings to delete the file. It could even implement a configurable check on the file creation time to allow for optional retention for a few days after deletion to allow for recovery in case of accidental deletion.

Boy132:
So, I now made the server deletion itself its own job. Therefore I added a new status ("deleting") to indicate that the server is queued for deletion. Maybe useful if a server has many backups and the deletion takes a few seconds.
That was the best solution without a wings edit I could think of.

~PS: I'll fix the tests soon.~

DaneEveritt:
I'd rather just fix this with a Wings edit, rather than complicating the state of servers on the Panel.

Boy132:
> I'd rather just fix this with a Wings edit, rather than complicating the state of servers on the Panel.

In that case I'll close this as I have no experience with go and/ or wings' code.

samuxita:
It hasn't been fixed yet.

Technoguyfication:
> I'd rather just fix this with a Wings edit, rather than complicating the state of servers on the Panel.

I don't believe this will work for S3 backups, right? You'd have to create a token on the panel and send it to Wings to delete the backup, at which point you may as well be deleting backups from the panel anyways.

PR #4212, opened 2022-06-29
fix: update upgrade command

### LINKED ISSUE #4251: Panel upgrade won't delete old version files
### Current Behavior

When upgrading the panel from 1.8.2 to 1.9+ the files deleted in 1.9 won't be deleted for the user updating the panel (all console files in the root resources directory and the SetupTwoFactorModal file. Which isn't harmless when having 0 modifications to your panel as you pull the builded version, but when users want to make changes to their panel. E.G building, it will give tons of errors due to those undeleted files.

I know that the priority of Pterodactyl isn't by users modifying their panel but over 75% of the users have at least some kind of modification needing to build for that.

Now all developers like me need to tell their clients to delete all files their selves to build, as this isn't included in the upgrading process.

And in general it is just messy to keep files in the project which aren't in use anymore.

### Expected Behavior

Having all removed files deleted when upgrading from any version (but 1.8 > 1.9 a specially).

### Steps to Reproduce

Upgrade your panel from < 1.8.2 to > 1.9

NOT FRESH INSTALL

### Panel Version

1.9>

### Wings Version

-

_No response_

### Docker Image

_No response_

### Error Logs

_No response_

- [x] I have searched the existing issues before opening this issue.

devnote-dev:
This PR fixes issues present in `v1.8.x` and `v1.9.x` with upgrading the panel frontend files, as well as implements a backup system for the current version that users can rollback to if upgrading fails.

Edit: closes #4251


hwalker928:
What if the user's pterodactyl installation isn't at `/var/www/pterodactyl`?

devnote-dev:
> What if the user's pterodactyl installation isn't at `/var/www/pterodactyl`?

As far as I'm aware the original version didn't make exceptions for this, I don't see why it should now.

FoksVHox:
See previous attempts: #4103, #3934

devnote-dev:
> See previous attempts: #4103, #3934

The first PR you mentioned was more or less an entirely new command that wasn't backwards compatible (i.e. not usable until the next major version), and I'm going to disregard the second one for obvious reasons. Furthermore, unlike those attempts, this PR resolves the existing issues rather than providing alternatives.

I spoke to Dane about this already and I'm aware that this PR may be closed in favor of the bash script being proposed by the core team, but said script is not available yet and I was still given the go-ahead to do this.

Doc94:
> As far as I'm aware the original version didn't make exceptions for this, I don't see why it should now.

i mean the original upgrade use the current directory not? for my case i use a custom directory for the app and the command work with no issues (ignorin when broken a few updates ago) maybe is better ask for the install directory or detect where work?

devnote-dev:
> i mean the original upgrade use the current directory not? for my case i use a custom directory for the app and the command work with no issues (ignorin when broken a few updates ago) maybe is better ask for the install directory or detect where work?

It could probably be implemented as a flag option, but:
1. I don't know if that's something that would be agreed on by the core team
2. that would also mean having to implement extra steps in the rollback system for people not using `/var/www/`

Doc94:
> > i mean the original upgrade use the current directory not? for my case i use a custom directory for the app and the command work with no issues (ignorin when broken a few updates ago) maybe is better ask for the install directory or detect where work?
> 
> It could probably be implemented as a flag option, but:
> 
> 1. I don't know if that's something that would be agreed on by the core team
> 2. that would also mean having to implement extra steps in the rollback system for people not using `/var/www/`

maybe a flag can be good like another things.. or make the .env can set this value for another things?

devnote-dev:
> maybe a flag can be good like another things.. or make the .env can set this value for another things?

I've added this as a flag option now.

devnote-dev:
Closing as this wont be merged into v1.x.

PR #4219, opened 2022-07-01
File upload status (v2 backport)

Boy132:
A backport of v2's file upload status. (https://github.com/pterodactyl/panel/commit/c6dccc568d91d95e6b5fa4dcaad4dfd532b6a5f6)
I only changed two things:
1. Changed the timeout before a completed upload is removed from 5 to 2 seconds. (_UploadButton.tsx#L111_)
2. Added the total and loaded bytes next to the name. (_FileManagerStatus.tsx#L61_)

file_upload


KaspianDev:
After applying this patch drag & drop no longer works.

Boy132:
> After applying this patch drag & drop no longer works.

I can't reproduce that with the given information. ("no longer works" isn't very descriptive)
You maybe got some conflicting third party mod or forget some changes when applying?

KaspianDev:
> > After applying this patch drag & drop no longer works.
> 
> I can't reproduce that. You maybe got some conflicting third party mod or forget some changes when applying?

I run ptero panel from latest commit (as of this message), no mods

Edit: Tried again on 1.9.2, same result

camwhi-te:
I'm running into the same issue with this (although my instance is modified).

Boy132:
What browser are you using? What files are you using? (amount of files, size of files, file types etc.)
Also, can you try on an unmodified panel? Just to be sure.

KaspianDev:
> What browser are you using? What files are you using? (amount of files, size of files, file types etc.) Also, can you try on an unmodified panel? Just to be sure.

Firefox, one 10KB yml file

Works fine on unmodified panel

Jelcoo:
This seems to be a problem on the Firefox browser.
The first browser is Firefox, the second one is Microsoft Edge.
https://user-images.githubusercontent.com/53396500/177564610-b9d6fe76-294f-4b9c-af3f-b8cd5431d34a.mp4

resources/scripts/components/server/files/UploadButton.tsx
```diff
@@ -19,18 +18,39 @@ const InnerContainer = styled.div`
...
     const directory = ServerContext.useStoreState((state) => state.files.directory);
+    const appendFileUpload = ServerContext.useStoreActions((actions) => actions.files.appendFileUpload);
+    const removeFileUpload = ServerContext.useStoreActions((actions) => actions.files.removeFileUpload);
 
     useEventListener(
         'dragenter',
         (e) => {
+            if (!isFileOrDirectory(e)) {
+                return;
+            }
```
Jelcoo, line 54: It seems to be a problem with these 3 lines. When I remove them it works again.

Boy132:
Turns out Firefox also adds a `application/x-moz-file` file to the drag event that caused these issues.

KaspianDev:
I updated the patch, nothing changed.

Firefox 103.0b9 (64-bit)

Edit: uploading files from file manager is fine, but from the "downloads" popup in firefox no, (but that one still works in original pterodactyl, without patch). Maybe there's something else we forgot about?

Edit 2: This might be the fix, https://stackoverflow.com/a/10703964/13768631

KaspianDev:
@DaneEveritt compat issues with that one still aren't fixed.

#4279

PR #4254, opened 2022-07-11
More variables for eggs to use

AlvinVrolijk:
I wanted to use more {{server. variables in my egg configurations


AlvinVrolijk (commented):
My version

james121op:
but why do we need these new variables what is the use case

DaneEveritt:
The only two I'm tempted to allow are name & description, unless there is a justifiable use case for the others.

AlvinVrolijk:
@DaneEveritt It's simple.. I want to know the uuidShort from in the container to use on a API to scale Minecraft servers and have them all have a unique name. The name is so the application knows how to call the server in general (ex: int-dev-34fds4d) and the id good to have if you want to communicate back to the Pterodactyl API. External ID same for development purposes.

DaneEveritt:
closed in 2eda1995b90fd894a40a2b2248f71d5bb89da2ce — I won't be adding other fields at this time.

PR #4262, opened 2022-07-18
fix: include oom_disabled in validated

devnote-dev:
The `oom_disabled` property for the store server request was never validated and passed to the server creation service, meaning that servers created via the API always had the OOM killer disabled. This PR fixes that bug, and should also fix updating `oom_disabled` via the update build endpoint.


matthewpi (approved):

KingTino104:
thank you so much matthew for merging this - i did this in v1.10.1 manually by adding the line of code to the file needed to edit so theres no need for it anymore thank you so much again

PR #4267, opened 2022-07-20
Fixed reCAPTCHA not available in some regions

vlssu:
For example, in China, the content of google.com cannot be used, and recaptcha.net solves this problem.


KaspianDev:
Seems shady, wouldn't it be better to for example just use hcaptcha in addition to recaptcha, or even replace recaptcha with it?

DaneEveritt:
Handled in abac3178020837087e67cfdc078b9bb04b86e2a9

PR #4279, opened 2022-07-24
Fix file uploading not working from download popup

KaspianDev:
This pull request fixes file uploading in firefox, when dragging from the "downloads" pupup after #4219.


DaneEveritt:
1cfa4100adbd51102fc1afdbc581423fd5582fca

PR #4283, opened 2022-07-25
Remove validation of type in fails for upload

### LINKED ISSUE #4280: File uploads being detected as folders
### Current Behavior

When uploading a .db (sqlite database file) the panel returns a `Folder uploads are not supported at this time.` error.

Files with other extensions work fine.

### Expected Behavior

Files with a .db file extension should upload like any other file.

### Steps to Reproduce

Attempt to upload a .db file, the above error should occur. If you rename the file to have a different extension (ie jpg) then the upload will succeed, 

A workaround for now is that after uploading with a different extension you can rename the file within the panel to restore the extension.

### Panel Version

1.10.0

### Wings Version

1.7.0

NA

### Docker Image

NA

### Error Logs

```bash
Wings: https://ptero.co/gigifovihu
Panel: NA -- Panel has not been creating logs since Jan 2021, probably an issue with my docker configuration/directory permissions, will attempt to resolve and append.
```

Doc94:
This close #4280 by removing the validation of file type.

Based in a [note ](https://developer.mozilla.org/en-US/docs/Web/API/File/type) about File.type this property only work for a few files like normal text or images, for the rest of files (like .db, .jar) this return empty then its a valid form to check a file in the upload process.

i dont know fully the system for try to suggest a better form for validation but can suggest this little fix.

### DIFF
--- a/resources/scripts/components/server/files/UploadButton.tsx
+++ b/resources/scripts/components/server/files/UploadButton.tsx
@@ -67,7 +67,7 @@ export default ({ className }: WithClassname) => {
     const onFileSubmission = (files: FileList) => {
         clearAndAddHttpError();
         const list = Array.from(files);
-        if (list.some((file) => !file.type && file.size % 4096 === 0)) {
+        if (list.some((file) => file.size % 4096 === 0)) {
             return addError('Folder uploads are not supported at this time.', 'Error');
         }
 


matthewpi:
This won't fix the problem, all removing that check does is only rely on checking the file size.  There might be a better way to handle detecting folder uploads, but the best bet might be just to remove this logic entirely.

PR #4284, opened 2022-07-26
Add back UptimeDuration for starting containers

AspectTheDev:
Added back UptimeDuration for starting/stopping container. That was removed in v1.9.0


DaneEveritt:
Can you make the string literal `Offline` just be `capitalize(status)` so that if we ever have some weird condition where there is no uptime for a starting/running container we at least display a valid status?

AspectTheDev:
> 
Sure, I've done that.

DaneEveritt:
Might need to do `status === null ? 'Offline' : capitalize(status)`

PR #4295, opened 2022-08-01
Add .gitattributes, exclude tests from archive

iamvar:
Add .gitattributes, exclude tests from archive


matthewpi:
Git archives are supposed to contain these files, I don't see why this would be wanted.

PR #4301, opened 2022-08-04
Increase task action delay limit to 3600 seconds

cymug:
This pull request raises the maximum delay between tasks from 900 seconds to 3600 seconds, and corrects a grammatical error in the message for having a delay that is too long.


DaneEveritt:
Why?

cymug:
The current time limit makes tasks that are longer than an hour too difficult to set up, and requires users to have empty actions with 15 minute delays. Increasing the limit to an hour solves this, and if something needs a longer delay than that it would probably be better off being in its own separate task.

robertdrakedennis:
This branch is targeting old develop branch, the PR is also a bit old, I'd be happy to discuss this in a future release of develop, feel free to open a discussion or @ me in Discord.

PR #4323, opened 2022-08-16
admin(eggs): add `force_outgoing_ip` option

### LINKED ISSUE #3841: Set the outgoing IP correctly when using multiple IPs on one node

By default docker always uses the first interface for outbound traffic. This can lead to issues with self-advertising servers like CS:GO as reported in #459.

Docker has added support for a new label that allows to set the outbound IP of a docker network: https://github.com/moby/libnetwork/pull/2454

In order to support this we would need to create one docker network per IP address and assign the containers to the correct networks.

There is a hacky solution [here](https://github.com/pterodactyl/panel/issues/459#issuecomment-1003668331), but since it's now possible we should add support for the official one.

matthewpi:
Requires https://github.com/pterodactyl/wings/pull/135

Closes #3841

### DIFF
--- a/app/Http/Controllers/Admin/Nests/EggController.php
+++ b/app/Http/Controllers/Admin/Nests/EggController.php
@@ -73,7 +73,7 @@ class EggController extends Controller
      */
     public function store(EggFormRequest $request): RedirectResponse
     {
-        $data = $request->normalize();
+        $data = $request->validated();
         $data['docker_images'] = $this->normalizeDockerImages($data['docker_images'] ?? null);
 
         $egg = $this->creationService->handle($data);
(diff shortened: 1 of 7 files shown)


parkervcp:
I feel like having your traffic use the expected ip is the only way to go here and not making it an option.

matthewpi:
> I feel like having your traffic use the expected ip is the only way to go here and not making it an option.

The problem is you lose internal container networking when this feature is enabled.  The way the feature works is best described by the Wings PR; it creates a dedicated network per-server with the `host_ipv4` option, which is the only way setting the correct outgoing IP seems to work (at least when using Docker).  I previously explored having both a dedicated network for outgoing traffic and one for internal traffic, but it is either not currently possible or requires a very specific configuration in order to work correctly.

parkervcp:
Okay. So it's a double edged sword. On one hand you get the outbound correct. On the other you lose internal networking.

A good callout.

Synkstar:
The only thing about this is I wish pterodactyl had an option to allow the servers to bind on the public ip but without any sort of natting or having to use host networking( since this has security issues ) so you dont get the issue where on source engine servers it can show players ips as 172.19.0.1 for example( this occurs when people join the server right after its restarted or crashed ). Plus not having natting will result in better performance.

I could probably do this with netns-dockergen but using that breaks the network usage graphs. Since I used this to have it go over a tunnel before from the container.

PR #4329, opened 2022-08-19
Support language in admin nav bar

Snipeur060:
Add @lang in process to supportl anguage


DaneEveritt:
As mentioned in `CONTRIBUTING.md` I am not accepting new PRs from non-project staff. Thanks for your contribution, but we're scrapping the existing admin area anyways, and translations are pointless for it at this time.

PR #4353, opened 2022-09-02
Adds hidden passwords in database modal

Paultje52:
The database password and jdbc connection string are hidden (input type password) until the user hovers over the input field with their mouse.
On mobile devices (without a mouse) the user can click on the input field to show it and click somewhere else to hide it.


DaneEveritt:
If we're doing it this way it should just be a standard password field with the `eye` icon that toggles visible and hidden like most every other site handles it. I'm not a fan of this approach in the PR, but thanks for taking a stab at it. Feel free to just open a bug report for it and we can handle it down the road.

PR #4382, opened 2022-09-17
Make maximal S3 part size configurable through env

### LINKED ISSUE #4361: Make S3 part size configurable for backups to prevent 504 errors
### Current Behavior

Currently there is a predefined part size for the multi-part S3 upload defined.

Definition of part size: [app/Http/Controllers/Api/Remote/Backups/BackupRemoteUploadController.php#L18](https://github.com/pterodactyl/panel/blob/ab08afb2da43133cbdcd351ebc3ea67ba518923d/app/Http/Controllers/Api/Remote/Backups/BackupRemoteUploadController.php#L18)

Splitting: [app/Http/Controllers/Api/Remote/Backups/BackupRemoteUploadController.php#L94](https://github.com/pterodactyl/panel/blob/ab08afb2da43133cbdcd351ebc3ea67ba518923d/app/Http/Controllers/Api/Remote/Backups/BackupRemoteUploadController.php#L94)

5GB is a quite large part and can cause issues on some S3 providers due to the time amount necessary to upload a part (depending on network speed).
This sometimes causes timeouts (504 Gateway timeout) due to the time effort necessary to upload a single 5GB part.
A configuration for the part size is missing.

### Expected Behavior

Some environment variable like "S3_MAX_PART_SIZE"  allows to configure some option in bytes (Default: 5GB).
This allows to define a reduced chunk size for S3 providers that run into a timeout with larger parts.

In the end, the solution should work like `BACKUP_PRESIGNED_URL_LIFESPAN`, where there is a default value and an environment variable that allows specifiying this value as needed.

This would allow for example to lower the size of a single part to 1GB instead of 5GB.

### Steps to Reproduce

- Setup the panel
- Setup wings
- Create a server that requires a larger amount of disk space (e. g. 20GB)
  - Something like `dd `can be used  to just dump some random data for testing purposes
- Setup S3 as backup provider
  - you must use a service with a short timeout or mediocre network speed
  - alternativly you have to manually limit the speed of the upload
- Create a backup of the large server
- Depending on current conditions, there is a mediocre chance that the upload of a part times out
- -> Backup failed

### Panel Version

1.10.1

### Wings Version

1.7.0

_No response_

### Docker Image

_No response_

### Error Logs

_No response_

pascal-zarrad:
### Description
This PR allows to configure a maximum part size for the S3 multi-part upload instead of using a fixed one. This can be pretty usefull when an S3 service is used with limited bandwidth (e. g. 100Mbit/s). 

With 100Mbit/s network, a 5GB part can take (depending on network conditions) up to 7 minutes. This can result in a 504 (Gateway timeout) for some S3 providers and let the backup fail due to their slow network.

By decreasing the part size to  e. g. 1GB, it is possible to overcome such issues.

### Fix
This PR adds the new environment variable `BACKUP_MAX_PART_SIZE` that can be configured with the maximum part size (in bytes). When set, the presigned URLs will be generated with the newly defined part size instead. Also, the custom part size will be appended to the response for Wings, so Wings also knows how big a single part should be.

When the new environment variable is not present or an invalid value is set, the previous fixed part size of 5GB will be used as default/fallback value.

### Test Cases
 - Do a backup without `BACKUP_MAX_PART_SIZE` being defined in the environment
   - Backup is created sucessully
   - The count of the parts uploaded by Wings (see Wings log) match the expected count (e. g. 20GB / 5GB = ~4 parts)
   - Download the created backups and ensure the archive and its files is not corrupt
   - Restore the backups to ensure restoring works as expected
 - Do a backup with `BACKUP_MAX_PART_SIZE` being set to 1GB (= 1073741824 bytes)
   - Backup is created sucessully
   - The count of the parts uploaded by Wings (see Wings log) match the expected count (e. g. 20GB / 1GB = ~20 parts)
   - Download the created backups and ensure the archive and its files is not corrupt
   - Restore the backups to ensure restoring works as expected
 - Do a backup with `BACKUP_MAX_PART_SIZE` being set to 5GB (= 5368709120 bytes)
   - Backup is created sucessully
   - The count of the parts uploaded by Wings (see Wings log) match the expected count (e. g. 20GB / 5GB = ~4parts)
   - Download the created backups and ensure the archive and its files is not corrupt
   - Restore the backups to ensure restoring works as expected

### Additional Info
 - If this PR gets accepted and merged, the documentation about backups has to be updated with the new environment variable.
 - Documentation should then also being updated with the `BACKUP_PRESIGNED_URL_LIFESPAN` variable, as this can be useful in conjuction with this change.
 - I did not find any tests for the  `BackupRemoteUploadController` controller, therefore none have been updated in this PR

### Related Issue
Fixes #4361

### DIFF
--- a/app/Http/Controllers/Api/Remote/Backups/BackupRemoteUploadController.php
+++ b/app/Http/Controllers/Api/Remote/Backups/BackupRemoteUploadController.php
@@ -15,7 +15,7 @@ use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 
 class BackupRemoteUploadController extends Controller
 {
-    public const PART_SIZE = 5 * 1024 * 1024 * 1024;
+    public const DEFAULT_MAX_PART_SIZE = 5 * 1024 * 1024 * 1024;
 
     /**
      * @var \Pterodactyl\Repositories\Eloquent\BackupRepository
@@ -89,9 +89,12 @@ class BackupRemoteUploadController extends Controller
         // the other presigned urls.
         $params['UploadId'] = $result->get('UploadId');
 
+        // Retrieve configured part size
+        $maxPartSize = $this->getConfiguredMaxPartSize();
+
         // Create as many UploadPart presigned urls as needed
         $parts = [];
-        for ($i = 0; $i < ($size / self::PART_SIZE); ++$i) {
+        for ($i = 0; $i < ($size / $maxPartSize); ++$i) {
             $parts[] = $client->createPresignedRequest(
                 $client->getCommand('UploadPart', array_merge($params, ['PartNumber' => $i + 1])),
                 $expires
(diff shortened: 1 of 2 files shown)


pascal-zarrad:
Side note:
There is a `\Pterodactyl\Repositories\Eloquent\BackupRepository` injected and stored in field `$repository` in current implementation of ` BackupRemoteUploadController `.

This ` BackupRemoteUploadController::repository` field seems not to be in use anywhere. 

I'd suggest to either 
 - remove the field or
 - replace the `Backup::query()->where('uuid', $backup)->firstOrFail()` call using an repository `getById` function.
I'm more into Symfony and not that into Eloquent and Laravel, so proper solution is up to the team.

Also, this change shouldn't be done in this PR, as it isn't in scope of the current changes.

matthewpi (approved):

matthewpi:
Please rebase against latest changes, or enable the `Allow maintainers to edit` option.

pascal-zarrad:
> Please rebase against latest changes, or enable the `Allow maintainers to edit` option.

Done

PR #4392, opened 2022-09-26
Allow uploading multiple files via web UI

diamkil:
Just allows to upload multiple files via the File Manager's upload button


LevM-A:
I think there should be a things that has SFTP files uploading in Actifity logs, it's just show a list of files in 1 activity log.
image

Because if you will upload multiple files with web than there will be a spam in activity logs with every files.
image

diamkil:
> Because if you will upload multiple files with web than there will be a spam in activity logs with every files. image

Well it worked already via drag and drop either way, just making it easier by allowing it via the button

PR #4395, opened 2022-09-30
Run panel scheduled tasks via cron in docker

antoniomika:
Previously, scheduled tasks created in the panel would not run in docker. There's already a cron mechanism in place, we just specifically need to run the right command.


TrixterTheTux:
They're handled by schedule:run, which in result then invokes the queued schedules (https://github.com/pterodactyl/panel/blob/develop/app/Console/Kernel.php#L29)

antoniomika:
Interesting! I'll see if I can figure out any other reasons for why they aren't triggered. Though, I use a volume mount containing my changes here and everything is working as expected

TrixterTheTux:
One thing I would try is flushing your cache (if possible; note that if you use e.g. Redis and it stores your sessions too they may expire and force login). I've seen in some cases the ->withoutOverlapping() condition get stuck be it due to the process crashing or hanging.

antoniomika:
I'm gonna close this for now and do a bit more debugging. Thanks!

Issue #4183, opened 2022-06-27
Option how much MB should convert into GB


LevM-A:
### Is there an existing feature request for this?

- [X] I have searched the existing issues before opening this feature request.

### Describe the feature you would like to see.

In admin panel, there should be a button that can change how much MB should convert to GB. before 1.9v, 1024mb should be 1gb, but right now it's 1000mb = 1gb. If i want to have 1 gb as 1024mb then I must turn on button in admin area panel.

### Describe the solution you'd like.

mostly people who are using pterodactyl panel are setting for servers as full GB ( such as 3072mb = 3gb, 8192mb = 8gb), But others side most hosting providers are setting 1gb as 1000mb (such as 3000mb = 3gb, 8000mb = 8gb).

TsjipTsjip:
This is wrong.
1 GB is and always will be 1000 MB. (Base 10 prefixes)
1 GiB is and always will be 1024 MiB. (Base 2 prefixes)

This is a common misconception and software adopting the options you are suggesting only add to the confusion. Perhaps you could make  a request to be able to change the unit to MiB if the places you want to use it in is managed by the panel.

Issue #4133, opened 2022-06-09
Wrong units used in console / wrong value conversion


LOENS2:
### Current Behavior

The values for memory, disk and network activity shown in the console are in mebibytes or kibibytes but are labeled with megabytes or kilobytes.

### Expected Behavior

The values should be converted according to the units shown in the console

### Steps to Reproduce

Get the byte value via ```https://pterodactyl.file.properties/api/client/servers/1a7ce997/resources``` and devide it by 1048576 to get the value in mebibytes. This value is shown in the console. If you devide the byte value by 1000000, you get the value in megabyte, but it doesn't match te value given in the console.

### Panel Version

1.8.1

### Wings Version

1.6.4

### Is there an existing issue for this?

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

Goksi:
Pretty sure its intended to work like that as MB  also can  be interpreted as 2^20 bytes. Term mebibyte might be confusing to end user

LOENS2:
It really confused me, because i had different values in Grafana (right megabyte conversion) and thought that my exporter made a wrong conversion, but upon realizing it was actually mebibytes in the panel, I now ask myself why it isn't converted correctly.

DaneEveritt:
> This value is shown in the console

What does this mean? We don't output anything into the console like this.

Clarkcj12:
I believe he is talking about the console graphs displayed on the servers console page.

LOENS2:
@DaneEveritt 
I'm talking about the server console in the panel for the docker container. On the left side of that window are the server stats I'm referring to. 
Sorry for the miscommunication

trenutoo:
[The math used for conversion is bytes / 1024](https://github.com/pterodactyl/panel/blob/develop/resources/scripts/helpers.ts#L3-L14), which indeed is a kilobyte (KiB) and mebibyte (MiB), and so forth. However, the values converted and displayed are something that is displayed to a human, and the majority of average users will not know the difference between KiB and kB as there's no acceptance in the industry for it, even though it's the accepted standard for base 2. 

[...]

PR #4419, opened 2022-10-09
fix: byte units

Cubxity:
This PR fixes inconsistency between byte units in the UI. Currently, Pterodactyl refers to units based on power of 10 and power of 2 as `mega-`, `giga-`, etc - which is wrong. According to [ISO IEC 80000](https://en.wikipedia.org/wiki/ISO/IEC_80000), binary prefixes such as `mebi-` and `gibi-` are used for units based on power of 2.

## Changes
- `bytesToString` formatter function
- Use MiB for memory's Y-axis in StatsGraph
- Use MiB for disk space in admin view (https://github.com/pterodactyl/wings/blob/develop/server/configuration.go#L71)


trenutoo:
This was already changed once in https://github.com/pterodactyl/panel/issues/4133 and reverted back again after complaints from end users https://github.com/pterodactyl/panel/issues/4183

Cubxity:
I suggest there should be a reconsideration regarding this issue. This greatly confuses users when dealing with other software that displays **correct** values.

**Examples:**
1. User has Grafana dashboard displaying memory usage on their Minecraft server. However, this doesn't match what's displayed on Pterodactyl.
2. User tries to allocated 2GB (2000 MB) on a server, however it's displayed differently in the console as "1.95GB" (which in reality, is 1.95GiB). This can cause confusion for many users.
3. User may think MB is MiB in the admin panel and allocate 1024 "MB", which is infact 976.5625 MiB. This may result in under-allocation for the application. (for example, Java with -Xmx1G)

Anyhow, the value displayed is completely wrong and differs from the real value. Furthermore, the use of the unit is inconsistent throughout the panel. I would highly recommend a reconsideration on this issue.

DaneEveritt (approved):
This is fine, it isn't actually changing any of the calculation, just adjusting the units to match the conversion factor.

matthewpi (approved):

PR #4420, opened 2022-10-10
Allow users to change the server description

### LINKED ISSUE #4412: Allow non-admins to change server description

Currently server owners and subusers with the `settings.rename` permission can only change the server name, not the description.

Grant server owners the ability to change the server description by default. It would also be useful to introduce a new permission separate from `settings.rename` so that subusers can also change the server description, in parallel with the `settings.rename` permission for subusers.

N/A

Boy132:
Closes #4412.

This allows users (with the permission `settings.rename`) to change the server desc, similar to the server name.

image


devnote-dev:
IMO "change-description" doesn't really fit for the permission name or endpoint... perhaps just "description" instead, to keep things concise?

Boy132:
> IMO "change-description" doesn't really fit for the permission name or endpoint... perhaps just "description" instead, to keep things concise?

You're right, I changed it now.
For some reason my brain thought the endpoint for the docker image is called "change docker image", so it felt right with the "change". :D

DaneEveritt:
Maybe just use the name permission? Seems weird to allow a user to only change one or the other, kinda unnecessarily granular IMO.

Also, description should probably be a textarea, not a text field?

Boy132:
> Maybe just use the name permission? Seems weird to allow a user to only change one or the other, kinda unnecessarily granular IMO.

I'm fine with both variants. I guess it's up to Matthew.

> Also, description should probably be a textarea, not a text field?

Since line breaks aren't displayed, I didn't see a reason for a text area:
image

DaneEveritt:
I was just thinking textarea more for readability — a long description is going to be annoying to edit when you have to scroll left/right IMO.

richiedevs:
I agree with Dane.

joshcannotcode:
I think it would be better if the name of the "Change Server Name" box was changed to "Change Server Details" and the description input is added to the "Change Server Name" box rather than being in a separate one, like this:
image

devnote-dev:
@notjoshmiles They're separate because they use 2 different endpoints, and I don't know if the form is capable of submitting 2 requests at the same time. If we were to go with Dane's suggestion though, then I agree that it would be better like this.

matthewpi:
If the endpoints are separate, then we should modify one to support updating both.  There should be one box for both these fields, not two individual ones.

Boy132:
> If the endpoints are separate, then we should modify one to support updating both. There should be one box for both these fields, not two individual ones.

Made some changes: it now uses the "rename" endpoint (and rename permission) to change both the server name and/ or description. Also, the input for the description is now a textarea.
image

app/Http/Controllers/Api/Client/Servers/SettingsController.php
```diff
@@ -34,8 +34,10 @@ public function __construct(
      */
     public function rename(RenameServerRequest $request, Server $server): JsonResponse
     {
+        $description = $request->input('description');
         $this->repository->update($server->id, [
             'name' => $request->input('name'),
+            'description' => $description !== null ? $description : '',
```
matthewpi: `$request->input('description') ?? ''`

matthewpi (changes requested):

matthewpi (approved):

PR #4421, opened 2022-10-10
Apply node maintenance mode to servers

### LINKED ISSUE #2705: Maintenance mode does not work
* Panel or Daemon: panel
* Version of Panel/Daemon: 1.1.1
* Server's OS: Debian 9
* Your Computer's OS & Browser: macOS, Safari 

**Describe the bug**
I'm not sure if you're aware of this, but Maintenance mode does not work. 

**To Reproduce**
After trying to enable it, and hitting the save button, it automatically disables after the page refresh.

**Expected behavior**
The Maintenance mode should enable and not let any users enter their server panel if they are on the node with mm enabled.

### LINKED ISSUE #4288: Can't update maintenance mode with application API
### Current Behavior

Endpoint: /api/application/nodes/X
Object "maintenance_mode" is ignored on PATCH request

### Expected Behavior

The endpoint should accept "maintenance_mode" update

Solution: Add 'maintenance_mode' to /app/Http/Requests/Api/Application/Nodes/StoreNodeRequest.php

return collect($rules ?? Node::getRules())->only([
            'public',
            'name',
            'location_id',
            'fqdn',
            'scheme',
            'behind_proxy',
            **'maintenance_mode',**
            'memory',
            'memory_overallocate',
            'disk',
            'disk_overallocate',
            'upload_size',
            'daemonListen',
            'daemonSFTP',
            'daemonBase',
        ])->mapWithKeys(function ($value, $key) {
            $key = ($key === 'daemonSFTP') ? 'daemonSftp' : $key;

### Steps to Reproduce

Make a PATCH request to /api/application/nodes/x 

**Maintenance Mode will not change**

with: 

                {
                "name": "NODETEST",
                "description": "TESTEXX",
                "location_id": 1,
                "fqdn": "nodetest.domain.com",
                "scheme": "https",
                "behind_proxy": false,
                "maintenance_mode": false,
                "memory": 6144,
                "memory_overallocate": -1,
                "disk": 750000,
                "disk_overallocate": -1,
                "upload_size": 100,
                "daemon_listen": 443,
                "daemon_sftp": 2022
                }

### Panel Version

1.10.1

### Wings Version

1.7.0

_No response_

### Docker Image

_No response_

### Error Logs

_No response_

Boy132:
Closes #2705. Closes #4288.

This applies the maintenance mode of the node to its servers.
When maintenance mode is active the server will be in a similar state as if it's suspended, e.g. it can't be started, files can't be accessed.

~~Also, this [changes the server transformer a bit](https://github.com/pterodactyl/panel/blob/95f0143335165aa8f45df03eb365c7504661c216/app/Transformers/Api/Client/ServerTransformer.php#L45:L48).~~

This also changes the color of the maintenance box in the admin area. (based on #3661)

For admins this message is displayed on the console page:
image

For all other pages (and for users on the console):
image


richiedevs:
This would be a nice addition

app/Models/Server.php
```diff
@@ -214,6 +214,11 @@ public function isSuspended(): bool
         return $this->status === self::STATUS_SUSPENDED;
     }
 
+    public function isUnderMaintenance(): bool
```
matthewpi: This function shouldn't be on the Server model, it should be on the node.  The name of it makes it seem like there is a server-level maintenance mode when it is clearly the node itself.  This should also make the rest of the code cleaner as it will more easily show that calling it requires the `node` relation to be loaded.

app/Transformers/Api/Client/ServerTransformer.php
```diff
@@ -42,7 +42,10 @@ public function transform(Server $server): array
             'internal_id' => $server->id,
             'uuid' => $server->uuid,
             'name' => $server->name,
-            'node' => $server->node->name,
+            'node' => [
```
matthewpi: I'm not sure if this is gonna fly.  This is a breaking API change and while I agree with the new schema, it may cause problems for users.  Maybe just a new field in the root named `is_node_under_maintenance` or something.  This should definitely be revisited in v2.

matthewpi (changes requested):

matthewpi (approved):

PR #4432, opened 2022-10-14
Fix rendered HTML in activity view

### LINKED ISSUE #4370: Erroneous display of activity logs containing closing html tag
### Current Behavior

Logs containing the closing HTML tag do not display correctly (blank space)
msedge_w2wGgXGJRI
msedge_CkAIXOgwdz

### Expected Behavior

Normal display for example `</abc>` in activity logs.

### Steps to Reproduce

- Go to server
- Change the name of the server to, for example `</abc>`.
- Check the activity logs

### Panel Version

1.10.1

### Wings Version

1.7.0

_No response_

### Docker Image

_No response_

### Error Logs

```bash
No errors
```

mrxbox98:
This fixes a bug where HTML tags would cause errors in the activity panel.
I have 0 idea why this solution actually works. I put some info I found about the issue on the issue thread, but I will repost it here:
I did some digging and this seems to be the result of the translation module used by Pterodactyl. Even when I turned off HTML tags in the i18n config, the server name HTML tags still tried to render while the ``<strong></strong>`` was not rendered. Not sure why this is happening because the strong tags are also passed in as values.

If anyone finds a better solution please let me know.

closes https://github.com/pterodactyl/panel/issues/4370


matthewpi:
I know at some point we will want to add proper translation support, but have you tried just removing i18n entirely?  Because right now it isn't even really wired up, just loaded and not even used.

mrxbox98:
> I know at some point we will want to add proper translation support, but have you tried just removing i18n entirely? Because right now it isn't even really wired up, just loaded and not even used.

My latest commit removes the React part of the i18n translation which fixes the bug. I haven't tested the other parts of the panel to see if anything has broken yet though.

DaneEveritt (changes requested):
I disagree with this change — I'm not sure I follow why we're deleting an entire package over some really janky rendering issues due to rare input values?

This seems like we're tossing a very reasonable approach for 99% of the codebase all because occasionally an activity log renders funky, rather than looking into ways to address that on the backend?

Why not just encode the actual values coming back in the activity log which would completely negate all of these issues without removing packages and implementing non-standard translation handler logic?

edit: @matthewpi the translation stuff is definitely loaded and used (notably, by this system) — and it is in a position where it can begin being used throughout the frontend without much additional work needed.

mrxbox98:
> Why not just encode the actual values coming back in the activity log which would completely negate all of these issues without removing packages and implementing non-standard translation handler logic?

Wdym?

matthewpi:
> @matthewpi the translation stuff is definitely loaded and used (notably, by this system) — and it is in a position where it can begin being used throughout the frontend without much additional work needed.

I was under the impression that we never started using the translation system, I didn't realize it was wired into anything, I remember the packages being there and some placeholder code but that was it.

mrxbox98:
> Why not just encode the actual values coming back in the activity log which would completely negate all of these issues without removing packages and implementing non-standard translation handler logic?

Do you mean like escaping certain values before they are passed to the Translate component?

matthewpi:
Gonna close this in favor of https://github.com/pterodactyl/panel/pull/4538

PR #4437, opened 2022-10-15
update mumble

QuintenQVD0:
This is a changed mumble egg as the do not provice a static binary anymore for the latest 1.4.x server version.
The current only way to get 1.4 running is you compile the source or run it from the alpine packages mumur what is offical by them.
This egg uses a image that has that mumble packages installed in the docker image
changes:
- The image
- startup cmd as it is now in PATH and does not need a ./ anymore and specify to use murmur.ini config
- File parser to set the sqlite database path to /home/container/murmur.sqlite
- Set the logfile to /home/container/murmur.log but notthing seeems to be writen to it
- remove the version variable as there is no version control as the binary is now in the docker image
- remove the logs part as it is not more used

EDIT
install script:
afbeelding

### DIFF
--- a/database/Seeders/eggs/voice-servers/egg-mumble-server.json
+++ b/database/Seeders/eggs/voice-servers/egg-mumble-server.json
@@ -10,7 +10,7 @@
     "description": "Mumble is an open source, low-latency, high quality voice chat software primarily intended for use while gaming.",
     "features": null,
     "docker_images": {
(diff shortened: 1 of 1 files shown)


QuintenQVD0:
I can join:
afbeelding

parkervcp (approved):
LGTM

trenutoo (approved):

Tealk:
I use this egg but i still get Mumble 1.3.4
<W>2023-01-21 13:52:38.604 Murmur 1.3.4 (1.3.4) running on X11: Linux 5.10.0-15-amd64: Booting server

PR #4451, opened 2022-10-20
Fix running phpunit tests on bare install

lancepioch:


lancepioch:
I don't like this solution. Instead we should switch the default test driver to logging.

PR #4455, opened 2022-10-21
Try out solutions from ignition

lancepioch:
This would only show when ignition shows, aka when debug is true

Sample:

image

### DIFF
--- a/app/Exceptions/ManifestDoesntExistException.php
+++ b/app/Exceptions/ManifestDoesntExistException.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Pterodactyl\Exceptions;
+
+use Exception;
+use Spatie\Ignition\Contracts\Solution;
+use Spatie\Ignition\Contracts\ProvidesSolution;
+
+class ManifestDoesntExistException extends Exception implements ProvidesSolution
+{
+    public function getSolution(): Solution
+    {
+        return new Solutions\ManifestDoesntExistSolution();
+    }
+}
--- a/app/Exceptions/Solutions/ManifestDoesntExistSolution.php
+++ b/app/Exceptions/Solutions/ManifestDoesntExistSolution.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Pterodactyl\Exceptions\Solutions;
+
+use Spatie\Ignition\Contracts\Solution;
+
+class ManifestDoesntExistSolution implements Solution
+{
+    public function getSolutionTitle(): string
+    {
+        return "The manifest.json file hasn't been generated yet";
+    }
+
+    public function getSolutionDescription(): string
+    {
+        return 'Run npm run build:production or yarn run build:production to build the frontend first.';
+    }
+
+    public function getDocumentationLinks(): array
+    {
+        return [
+            'Docs' => 'https://github.com/pterodactyl/panel/blob/develop/package.json',
+        ];
+    }
+}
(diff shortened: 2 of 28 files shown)


app/Exceptions/Solutions/ManifestDoesntExistSolution.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Pterodactyl\Exceptions\Solutions;
+
+use Spatie\Ignition\Contracts\Solution;
+
+class ManifestDoesntExistSolution implements Solution
+{
+    public function getSolutionTitle(): string
+    {
+        return "The manifest.json file hasn't been generated yet";
```
matthewpi: Come on man, can't even use PHP CS Fixer smh.  String should be using single quotes.

app/Exceptions/Solutions/ManifestDoesntExistSolution.php
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    public function getSolutionDescription(): string
+    {
+        return 'Run npm run build:production or yarn run build:production to build the frontend first.';
```
matthewpi: Only suggest `yarn build:production`.  Nobody should be using `npm` as we only provide a `yarn.lock` file.  This may change later down the road if we switch to something like pnpm, but that will definitely be a v2 only kind of change.

app/Exceptions/ManifestDoesntExistException.php
```diff
@@ -0,0 +1,15 @@
...
+
+use Exception;
+use Spatie\Ignition\Contracts\Solution;
+use Spatie\Ignition\Contracts\ProvidesSolution;
+
+class ManifestDoesntExistException extends Exception implements ProvidesSolution
+{
+    public function getSolution(): Solution
+    {
+        return new Solutions\ManifestDoesntExistSolution();
```
matthewpi: Is there any specific reason this is a relative import rather than being fully imported?

matthewpi (changes requested):

app/Exceptions/ManifestDoesntExistException.php
lancepioch: One less import and less characters. Honestly I should've done the same thing for Exception because it's only referenced once as the extended class.

app/Exceptions/Solutions/ManifestDoesntExistSolution.php
lancepioch: Didn't want to escape the apostrophe. But also double quotes are fine. It's a common myth that the string interpolation has any effect. Only after running strings in loops for millions of times do you see some millisecond differences between '' and "" strings.

app/Exceptions/Solutions/ManifestDoesntExistSolution.php
lancepioch: Good point!

app/Exceptions/Solutions/ManifestDoesntExistSolution.php
matthewpi: I completely missed the apostrophe there, my bad.  We should probably enforce a strict code style at some point, right now we have it setup for dev (and in CI) but it doesn't actually enforce it.

lancepioch:
I also tried doing the runnable solutions, but I couldn't get it to work, not sure if that's because of our code or not...

matthewpi (approved):

matthewpi:
Actually could you change the file names from `ManifestDoesntExist*` to  `ManifestDoesNotExist*`?  The `nt` prefix in a file name just feels weird.

PR #4465, opened 2022-10-22
Replace schedule and task repositories

lancepioch:


app/Http/Controllers/Api/Client/Servers/ScheduleTaskController.php
```diff
@@ -84,7 +83,7 @@ public function update(StoreTaskRequest $request, Server $server, Schedule $sche
             throw new HttpForbiddenException("A backup task cannot be created when the server's backup limit is set to 0.");
         }
 
-        $this->repository->update($task->id, [
+        Task::query()->update([
```
matthewpi: Are you sure this updates the task correctly?  I don't see an ID or model instance reference anywhere.

lancepioch: You're correct, fixed.

PR #4469, opened 2022-10-23
Replace allocation repository

lancepioch:


app/Http/Controllers/Admin/NodesController.php
```diff
@@ -125,11 +123,12 @@ public function allocationRemoveMultiple(Request $request, int $node): Response
      */
     public function allocationRemoveBlock(Request $request, int $node): RedirectResponse
```
matthewpi, line 124: Unrelated to this PR specifically, but does this request even have validation?

app/Http/Controllers/Admin/NodesController.php
```diff
@@ -140,14 +139,12 @@ public function allocationRemoveBlock(Request $request, int $node): RedirectResp
...
      *
-     * @throws \Pterodactyl\Exceptions\Model\DataValidationException
-     * @throws \Pterodactyl\Exceptions\Repository\RecordNotFoundException
      */
     public function allocationSetAlias(AllocationAliasFormRequest $request): \Symfony\Component\HttpFoundation\Response
     {
-        $this->allocationRepository->update($request->input('allocation_id'), [
-            'ip_alias' => (empty($request->input('alias'))) ? null : $request->input('alias'),
-        ]);
+        $allocation = Allocation::query()->findOrFail($request->input('allocation_id'));
```
matthewpi, line 145: Is there a reason to select the allocation just to end up updating it?  Wouldn't it be simpler to do a single update?

lancepioch: The idea is to replace the majority of these findOrFail calls with [implicit bindings](https://laravel.com/docs/9.x/routing#implicit-binding) in the next "round". For now, there will be a couple extra calls here and there, none of them should be in loops, so no N+1.

PR #4480, opened 2022-10-24
Replace database repository

lancepioch:
The reason I've skipped two tests is because the methods are no longer mocked, but actually called to create the databases. Because they are in a transaction, this causes issues since the database create in mysql automatically closes the transaction  in the testing framework.


hwalker928:
It appears that this pull request causes the database connection to be running on the panel database, and not the database host itself.

PR #4501, opened 2022-10-29
Removed unused IP

mrxbox98:
I couldnt find the reason why this was here and the only filter that can be used is event anyways.


robertdrakedennis:

PR #4503, opened 2022-10-29
server delete and other stuff

ProsameYT:
https://pteropaste.com/2tze6f


QuintenQVD0:
That are not pterodactyl logs but a fork

PR #4513, opened 2022-11-01
egg(teamspeak): fix database support

Wunderharke:
I know pull requests from new contributors aren't usually allowed, but this is a very small change that gave me a headache the other day.
It fixes database support for MariaDB by copying a necessary library. See https://forum.teamspeak.com/threads/132925-unable-to-load-database-plugin-library-libts3db_mariadb-so


trenutoo (changes requested):
Thanks for taking the time to contribute to the project. I'm afraid there are some breaking issues:

* This redirection breaks the tar step causing the whole installation to fail. You either want to use `&&` or perform the copy on a different line.

Wunderharke:
Sorry for the late response and the wrong pull request. I've updated the commit, and it should be working now.
I've moved the command to a new line and also escaped the slashes.

trenutoo (approved):
LGTM. Thanks for contributing :+1:

matthewpi:
If you are able to can you rebase your changes on top of `1.0-develop`?  When I switched the branch the PR will merge into, the commits develop has ahead of 1.0-develop persisted.

Wunderharke:
Should be correct and on top of `1.0-develop` now.

PR #4516, opened 2022-11-02
Create CodeOwners

lancepioch:


Doc94:
Its not better work with teams here?
this allow make changes in teams later and not in this file.

matthewpi:
I'll add a CODEOWNERS file once I have a better grasp of how I want it to be setup.  Likely we will also use teams instead of individual users to make maintenance of it easier.

lancepioch:
@matthewpi Yeah I do like the teams idea as well

PR #4523, opened 2022-11-04
Fix 'Launch SFTP' button

flaxeneel2:
This tiny commit fixes the 'Launch SFTP' button in the settings tab of a server
The issue in question:
image

Basically, just making sftp open the /home/container/ folder instead of opening the / that it would open by default.

### DIFF
--- a/resources/scripts/components/server/settings/SettingsContainer.tsx
+++ b/resources/scripts/components/server/settings/SettingsContainer.tsx
@@ -50,7 +50,7 @@ export default () => {
                                     </div>
                                 </div>
                                 <div css={tw`ml-4`}>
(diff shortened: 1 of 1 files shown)


Jcodeerd:
Ive never seen this happen though?

flaxeneel2:
> Ive never seen this happen though?

I guess I did not test with many SFTP clients, though it has been an issue at least for WinSCP users for a while, though looking at the uri, it would make sense if it was the same issue for others since it did not specify a path before.

flaxeneel2:
nvm it was a mistake on my part while i was changing other things with the panel and is not an issue with the panel itself

PR #4531, opened 2022-11-06
Handle reinstall failures differently from normal installation failures

### LINKED ISSUE #1994: Manually triggering a server reinstall that fails will brick the server instance
* Panel or Daemon: Panel
* Version of Panel/Daemon: `develop`

**Describe the bug**
If a user or an administrator triggers a server reinstallation for a server that has already installed successfully but the install fails, the server will enter a state which requires manual DB modification to get working again. This occurs because the system does not know that this is a manual reinstall, and just marks the server install state as being "2" incorrectly.

**Expected behavior**
The server should just "fail" the install and go back into a state where the user can interact with ti and try again.

Honestly, all failed installs should work this way, there is no reason not to do this now that the daemon just fetches the server from the API whenever the data is needed. Theres minimal risk of bad data or data corruption happening.

lancepioch:
Fixes #1994 with necessary flag sent by wings


matthewpi:
We will still need to implement logic in Wings to support this, but this at least supports the Panel backend side of things.  It would also be nice to use this state to show a notice in the client UI about a reinstallation failure that can be dismissed and reset the server status back to `NULL`.

matthewpi (approved):

PR #4532, opened 2022-11-06
Add Cancel Server Transfer modal

### LINKED ISSUE #4505: Failed transfer causes server deadlock
### Current Behavior

When transferring a server, if the connection to the node is interrupted, the transfer fails and the server is completely unusable: it cannot be accessed via the client or admin panel, nor can its state be overridden in the panel. The same occurs when trying to transfer a running server (see #4244). Attempting to change the status of a server in this state results in a HTTP 409 Conflict response.

### Expected Behavior

When a server transfer fails, there should be an option to recover the server. The transfer service does not modify the server or its contents meaning that it is still usable (and this can be proved if you manually resolve the state conflict). There should be a dedicated status for when a server is transferring and one for if that transfer fails. The current system relies on the `transfer` property of a server which isn't a reliable way of tracking a server's transfer progress and doesn't follow existing tracking systems (e.g. server installs, server suspension, etc). It also doesn't allow the server to be transferred again afterwards.

### Steps to Reproduce

1. Initiate the transfer process
2. Interrupt the node (i.e. terminate its connection, kill it, etc)

or alternatively:

1. Start a server (preferably a Minecraft server)
3. Initiate the transfer process

### Panel Version

1.10.4

### Wings Version

1.7.2

N/A

### Docker Image

N/A

### Error Logs

```bash
Unavailable
```

lancepioch:
We are planning for a better resolution overall, but for now this will help people immediately attempt to resolve their failed transfers.

Resolves #4505

### DIFF
--- a/app/Http/Controllers/Admin/Servers/ServerTransferController.php
+++ b/app/Http/Controllers/Admin/Servers/ServerTransferController.php
@@ -80,6 +80,25 @@ class ServerTransferController extends Controller
     }
 
     /**
+     * Cancels the specified server's transfer.
+     */
+    public function cancel(Request $request, Server $server): RedirectResponse
+    {
+        if (!$transfer = $server->transfer) {
+            $this->alert->danger(trans('admin/server.alerts.transfer_does_not_exist'))->flash();
+
+            return redirect()->route('admin.servers.view.manage', $server->id);
+        }
+
+        $transfer->successful = true;
+        $transfer->save();
+
+        $this->alert->success(trans('admin/server.alerts.transfer_cancelled'))->flash();
+
+        return redirect()->route('admin.servers.view.manage', $server->id);
+    }
+
+    /**
      * Assigns the specified allocations to the specified server.
      */
     private function assignAllocationsToServer(Server $server, int $node_id, int $allocation_id, array $additional_allocations)
--- a/app/Models/Server.php
+++ b/app/Models/Server.php
@@ -2,8 +2,11 @@
 
 namespace Pterodactyl\Models;
 
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Notifications\Notifiable;
 use Illuminate\Database\Query\JoinClause;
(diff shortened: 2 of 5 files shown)


matthewpi:
While adding this endpoint is useful, I'd rather not handle it like this.  I know there is a need for a feature to cancel transfers, but this still needs to be wired into Wings.  All this change does is unblock the panel which is only useful if both nodes have no knowledge of the transfer anymore.  I feel people will use this new action and break their server even more if we don't spend time implementing it properly.

LyrinoxTechnologies:
Are we able to get instructions on how to recover servers stuck in an infinite transfer loop in the meantime then?

I have a breached node in my network I need to nuke for security and there is one of my servers stuck in an infinite transfer as I didn't know it automatically started

LyrinoxTechnologies:
> Are we able to get instructions on how to recover servers stuck in an infinite transfer loop in the meantime then?
> 
> I have a breached node in my network I need to nuke for security and there is one of my servers stuck in an infinite transfer as I didn't know it automatically started

For anyone viewing this, while they are working on the official fix for this, here are instructions for recovering your server from the state (if you don't already know)

**USE WITH CAUTION**
If your server is actually just transferring and *not actually stuck* this WILL result in file deletion!!
---

Step 1) log into your pterodactyl database (typically named `panel` )

Step 2) Navigate to the `server_transfers` table

Step 3) Find the any row that says `NULL` in it

Step 4) Delete the row

The transfer has now been aborted and you can restart the transfer.

To ensure it completes, make sure the server is off and Wings won't get interrupted and the panel won't get disconnected from the new node while it's happening.

If your server gets stuck again, repeat these steps.

matthewpi:
If you would like, this can now be wired up to Wings.

There are two endpoints in Wings for cancelling transfers.
<https://github.com/pterodactyl/wings/blob/develop/router/router.go#L60> for the destination/target node. (`DELETE /api/transfers/:server`)
<https://github.com/pterodactyl/wings/blob/develop/router/router.go#L81> for the source node. (`DELETE /api/servers/:server/transfer`)

PR #4537, opened 2022-11-08
Show warning in allocation management for running servers

### LINKED ISSUE #4266: The Panel Allows to Modify Ports of Running Servers Leading to Incorrect State on Nodes
### Current Behavior

If you have a server running on example port `27003` and for some reason change to remove this port from this server or change the main port to any other, you will have problems with this allocation later on.
It is important to say that the server must be running when you remove/change its port.
And finally, when you create a new server using the removed allocation of the other server that was online, you get an already allocated port error.

### Expected Behavior

If the server is running, you cannot remove the main port or the extra ports.

### Steps to Reproduce

- Remove/change the main port of a running server.
- Try creating a new server with this port
- _Voiala!_

### Panel Version

1.9.2

### Wings Version

1.6.4

Minecraft

### Docker Image

_No response_

### Error Logs

```bash
Wings: https://ptero.co/rocywatopu
```

- [x] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

lancepioch:
Resolves #4266

image


matthewpi:
I wonder if this change might cause problems if a node is offline or the connection times out as it will block the page for whatever the HTTP timeout is.  The ideal solution would be to make the request to the node in JavaScript once the page is loaded, but that has it's own complexities.

lancepioch:
@matthewpi Yeah I thought of that. I think we can just reduce the timeout in this particular instance to maybe a couple seconds or less. Thoughts?

matthewpi:
> @matthewpi Yeah I thought of that. I think we can just reduce the timeout in this particular instance to maybe a couple seconds or less. Thoughts?

It would be a bit of a pain to wire-up through the code, but it is definitely an option.  I guess we can keep it as is with a lower timeout and when the new admin comes out we will do it in the browser rather than before rendering the page.

PR #4538, opened 2022-11-08
Fix activity logs and prevent them from removing text

### LINKED ISSUE #4370: Erroneous display of activity logs containing closing html tag

lancepioch:
The correct way to fix this is to update the i18n package from React and utilize this new behavior: https://github.com/i18next/react-i18next/issues/1323

Resolves #4370

Basically the translation is removing the tags completely instead of escaping them. By default the strong tag is allowed though which is why it still works.

### DIFF
--- a/resources/scripts/components/elements/activity/ActivityLogEntry.tsx
+++ b/resources/scripts/components/elements/activity/ActivityLogEntry.tsx
@@ -19,7 +19,7 @@ interface Props {
 
 function wrapProperties(value: unknown): any {
     if (value === null || typeof value === 'string' || typeof value === 'number') {
-        return `<strong>${String(value)}</strong>`;
+        return `${String(value)}`;
     }
 
     if (isObject(value)) {


matthewpi:
Is there a reason why we don't just update the dependency?

matthewpi:
Let's update the dependency to fix this.  We will likely end up doing it later anyways, so doing it now and tracking it with a proper issue is gonna be the best in the long-run.

PR #4542, opened 2022-11-09
Change MB to MiB on admin pages

Jcodeerd:
This PR changes MB to MiB which was already changed on the frontend, but not on the admin pages.


matthewpi:
If we are going to change these, you should change the rest of the usages of `B` suffixes.  Specifically on the node pages as well.

Jcodeerd:
Ah, I forgot those. I'll change those tomorrow!

Jcodeerd:
I think those are all, unless I missed any.

matthewpi:
I just made these changes myself (https://github.com/pterodactyl/panel/commit/2f4a60c9615af0f06b025c3936ade89dc87ba974), I'm not sure how you were making these changes but it is always easier to use the find and replace functionality of an editor.

PR #4544, opened 2022-11-09
fix(transformers): force object type for properties

### LINKED ISSUE #4543: Inconsistent API activity properties type
### Current Behavior

Same as #4296 with the "properties" field in the activity logs object.

### Expected Behavior

The "properties" field should always be an object, never an array.

### Steps to Reproduce

Request the account or server activity logs via the client API. In particular, the `user:two-factor.delete` activity will always return an array, as supposed to other activities which may or may not include the IP, user-agent, or additional properties.

### Panel Version

1.10.4

### Wings Version

1.7.2

N/A

### Docker Image

N/A

### Error Logs

```bash
N/A
```

devnote-dev:
Fixes #4543.


matthewpi:
It would probably be a good idea to add a test for this to ensure this logic doesn't break.  If you are willing and able to add that, I'd appreciate it.  I'll end up merging this regardless though.

matthewpi (approved):

PR #4541, opened 2022-11-09
Use Node 16 over 14 in Dockerfile; add ESlint workflow

camwhi-te:

PR #4545, opened 2022-11-09
Use Node 16 over 14 in Dockerfile; add ESlint workflow

camwhi-te:
Repeat of #4541, using the `develop-1.0` branch as intended. (sorry!)


matthewpi:
Node is only used to build the assets, I'm going to be leaving this as is for now.

PR #4546, opened 2022-11-10
egg(teamspeak): port and argument updates

TaktischerSpeck:
Teamspeak automatically binds the RAW, SSH and HTTP query port to the default values.
If you want to have more than one Teamspeak on the same node all Teamspeak servers will try / will bind these ports on the same default values.
(Default Teamspeak egg)
`|INFO    |Query         |   |listening for ssh query on 0.0.0.0:10022, [::]:10022`
`|INFO    |Query         |   |listening for http query on 0.0.0.0:10080, [::]:10080`

You can decide on which ports the server will bind, also which protocols it should use.
I added these variables to the egg config and Startup:
- Query protocols
- Query SSH port
- Query HTTP port


database/Seeders/eggs/voice-servers/egg-teamspeak3-server.json
```diff
@@ -53,7 +55,38 @@
             "default_value": "10011",
             "user_viewable": true,
             "user_editable": false,
-            "rules": "required|integer|between:1,65535"
+            "rules": "required|integer|between:1,65535",
```
trenutoo: Ports below 1024 can't be used as they require root. This applies to all of the variables below. TS3 also doesn't try to bind to ports below 1024.

database/Seeders/eggs/voice-servers/egg-teamspeak3-server.json
```diff
@@ -53,7 +55,38 @@
...
+            "field_type": "text"
+        },
+        {
+            "name": "Query Protocols",
+            "description": "Comma separated list of protocols that can be used to connect to the ServerQuery | \r\nPossible values are raw, ssh and http | \r\nE.g.: raw,ssh,http",
+            "env_variable": "QUERY_PROTOCOLS_VAR",
+            "default_value": "raw",
```
trenutoo: The default value in `ts3server.ini` is enabling all except HTTPS `raw,http,ssh`. I'd say disabling webinterface is not the best idea, especially when `raw` is planned to be disabled by default in the near future.

trenutoo (changes requested):
The webquery and server query ports are optional and shouldn't affect running multiple servers at once if those ports are not assigned or exposed. They also can be changed in the `ts3server.ini` similar to the majority of options.

It doesn't hurt to make it a configurable option. SSH option by itself is useless without `query_ssh_rsa_host_key`.

database/Seeders/eggs/voice-servers/egg-teamspeak3-server.json
TaktischerSpeck: When / where did they say `raw` will be disabled?
Last update was in 06/22, you sure they will push this soon?

I changed it to the default of the documentation, I would not remove raw in my experience a lot when not most users use this.

> Default: The default starts a raw or "classic" ServerQuery, and an encrypted ServerQuery using SSH

database/Seeders/eggs/voice-servers/egg-teamspeak3-server.json
TaktischerSpeck: Changed

TaktischerSpeck:
> It doesn't hurt to make it a configurable option. SSH option by itself is useless without `query_ssh_rsa_host_key`.

`query_ssh_rsa_host_key` only gives a path to the file which contains the key, if not set it will just create a file and key which you can then exchange for your key.
So yes it is useless, but the required step (or wished step for multiple TeamSpeak on one host) to use the SSH Query

database/Seeders/eggs/voice-servers/egg-teamspeak3-server.json
trenutoo: In their Forum announcements when posting about new HTTP(S) webquery. Disabling by default in the near future, not removing it.

TaktischerSpeck: Ah okay, but would you just remove it on default because of this?
I could imagine a bunch of people wondering.

trenutoo: No, I would set the default values to match the current default, which in `ts3server.ini` by default was everything except https.

TaktischerSpeck: I checked with a new TeamSpeak, default for the ini is `raw,http,ssh`
~~Without the ini the default should be `raw,ssh`~~, nvm they did not update the documentation sorry

So you would prefer  `raw,http,ssh` right?

TaktischerSpeck:
> > It doesn't hurt to make it a configurable option. SSH option by itself is useless without `query_ssh_rsa_host_key`.
> 
> `query_ssh_rsa_host_key` only gives a path to the file which contains the key, if not set it will just create a file and key which you can then exchange for your key. So yes it is useless, but the required step (or wished step for multiple TeamSpeak on one host) to use the SSH Query

@Software-Noob what is about the `query_ssh_rsa_host_key`, it does only specify the location of the key, do you want this to be configurable too?

Is there something else I should change?

trenutoo (approved):
LGTM. Thanks for contributing :+1:

PR #4551, opened 2022-11-11
Fix Undefined array key

NotYusta:
Fixes an error issue below:
[2022-11-11 17:30:58] production.ERROR: Undefined array key 0 {"userId":2,"exception":"[object] (ErrorException(code: 0): Undefined array key 0 at /var/www/pterodactyl/console-panel/app/Repositories/Wings/DaemonFileRepository.php:38) [stacktrace]

### DIFF
--- a/app/Repositories/Wings/DaemonFileRepository.php
+++ b/app/Repositories/Wings/DaemonFileRepository.php
@@ -35,7 +35,9 @@ class DaemonFileRepository extends DaemonRepository
             throw new DaemonConnectionException($exception);
         }
 
-        $length = (int) $response->getHeader('Content-Length')[0] ?? 0;
+        $fileLength = $response->getHeader('Content-Length');
+        
+        $length = (int) $fileLength ? $fileLength[0] : 0;
 
         if ($notLargerThan && $length > $notLargerThan) {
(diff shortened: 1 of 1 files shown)


NotYusta:
Basically the content-length is undefined so you can't access the array, you gotta check if the content length value is null or not.

app/Repositories/Wings/DaemonFileRepository.php
```diff
@@ -35,7 +35,9 @@ public function getContent(string $path, int $notLargerThan = null): string
             throw new DaemonConnectionException($exception);
         }
 
-        $length = (int) $response->getHeader('Content-Length')[0] ?? 0;
+        $fileLength = $response->getHeader('Content-Length');
+        
+        $length = (int) $fileLength ? $fileLength[0] : 0;
```
lancepioch, line 40: The exception is because the cast has a higher precedence than the null coalesce. This is what we want:

```suggestion
        $length = (int) ($response->getHeader('Content-Length')[0] ?? 0);
```

matthewpi:
https://github.com/pterodactyl/panel/commit/c3521e022161606c115d8a900040cf2cae190a96

PR #4559, opened 2022-11-18
Set cache & session drivers in .env.example to recommended value

Boy132:
Follow up of #3915.

Currently it still uses filesystem as default driver because the values in the example .env are "file":
image


trenutoo:
See the commit history as to why it was reverted

https://github.com/pterodactyl/panel/commit/3e55a79439bc91071a7ef690744d9d68b05149e8

Boy132:
> See the commit history as to why it was reverted
> 
> [3e55a79](https://github.com/pterodactyl/panel/commit/3e55a79439bc91071a7ef690744d9d68b05149e8)

Ah, didn't see that mb!

PR #4565, opened 2022-11-22
Update ARK Survival Evolved egg for fix some mistakes

iblinder:
Fixed rcon command syntax for `saveworld` and `doExit` commands,  option '-c' not exists, from rcon help:
```
   To run single mode type commands after options flags. Example:
   rcon -a 127.0.0.1:16260 -p password command1 command2

   To run terminal mode just do not specify commands to execute. Example:
   rcon -a 127.0.0.1:16260 -p password
```

Do not send to rcon ^C on exit, use :q to exit with exit code 0 for correct exit from `until` cycle.  
Do not use trap command, on correct exit from cycle, command placed behind it will work.

### DIFF
--- a/database/Seeders/eggs/source-engine/egg-ark--survival-evolved.json
+++ b/database/Seeders/eggs/source-engine/egg-ark--survival-evolved.json
@@ -15,12 +15,12 @@
         "quay.io\/parkervcp\/pterodactyl-images:debian_source"
     ],
     "file_denylist": [],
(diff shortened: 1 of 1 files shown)


melvyn2:
This still needs to be merged, Ark server stopping is completely broken by now. Have to manually `DoExit` and `:q`.

Boy132:
> This still needs to be merged, Ark server stopping is completely broken by now. Have to manually `DoExit` and `:q`.

The default ark egg is out of date in general. (as well as this PR)
Please use the [egg from the egg repo](https://github.com/parkervcp/eggs/tree/master/game_eggs/steamcmd_servers/ark_survival_evolved) instead until the stock egg gets updated.

PR #4568, opened 2022-11-22
Fix timeout during decompress

### LINKED ISSUE #4415: Timeout when try to decompress a tar.xz file
### Current Behavior

.tar.xz decompress via the file manager results in a timeout.
.tar.gz works fine.

### Expected Behavior

.tar.xz files should not timeout when you try to decompress it.

### Steps to Reproduce

1. upload a .tar.xz file to the server
2. Try decompress it via the file manager

### Panel Version

1.10.1

### Wings Version

1.7.0

_No response_

### Docker Image

_No response_

### Error Logs

```bash
Panel:
[2022-10-08 13:18:40] production.WARNING: GuzzleHttp\Exception\ConnectException: cURL error 28: Operation timed out after 5001 milliseconds with 0 bytes received (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://xxxxxx:8080/api/servers/2011f34f-2609-4169-b724-89ba03c993f1/files/decompress in /var/www/pterodactyl/vendor/guzzlehttp/guzzle/src/Handler/CurlFactory.php:210

wings:
INFO: [Oct  8 11:30:25.879] starting file decompression file=1.19.2.tar.xz request_id=936bc0d0-46aa-44c6-a258-9ccec8ace6af root_path=/ server_id=2011f34f-2609-4169-b724-89ba03c993f1
```

VibeGAMESNL:
Fix CURL 5 second timeout when calling the decompress endpoint

Closes #4415


matthewpi:
This is not a proper solution to this problem.  Wings should be doing this operation in the background just like remote file downloads, reinstalls, transfers, etc.  For the file manager, all tasks should be async and progress should be displayed in the UI.  For the API, options to do requests in the background or foreground should be provided.

VibeGAMESNL:
> This is not a proper solution to this problem. Wings should be doing this operation in the background just like remote file downloads, reinstalls, transfers, etc. For the file manager, all tasks should be async and progress should be displayed in the UI. For the API, options to do requests in the background or foreground should be provided.

The same "fix" seems to be applied on the compress function. https://github.com/pterodactyl/panel/commit/95e8492c5dea75f62562ce333d09c6fe3d7b4966

I understand you want to run this async in wings but at the moment this is not implemented yet.

Nyabsi:
Can confirm, this also solved the issue for me I tried to extract a .zip file over 5 gigabytes but it kept doing timeout every 15 seconds which essentially caused it always stop half way but after applying this commit it works like butter.

matthewpi (approved):

PR #4574, opened 2022-11-25
feat: update admin ui + various text

wdhdev:


yesBad:
you're aware that the whole admin ui is getting revamped?

matthewpi:
Thanks for trying to contribute, however none of these changes are useful and some of the proposed changes make parts of the software more confusing rather than clarifying anything.  The admin is being entirely replaced soon as well, so I don't see the value in changing the existing one other than fixing major issues.

wdhdev:
> you're aware that the whole admin ui is getting revamped?

I did not know that.

PR #4575, opened 2022-11-25
Update README.md

wdhdev:


TrixterTheTux:
It would probably help to provide reasoning on the pull requests, as for example converting the readme to HTML makes reading the file locally a lot harder than just using markdown (and unclear why that is needed?). There's also still markdown mixed in.

wdhdev:
> It would probably help to provide reasoning on the pull requests, as for example converting the readme to HTML makes reading the file locally a lot harder than just using markdown (and unclear why that is needed?). There's also still markdown mixed in.

Using HTML there allows the content to be centered as you cannot do that with markdown.

lokerhp:
> > It would probably help to provide reasoning on the pull requests, as for example converting the readme to HTML makes reading the file locally a lot harder than just using markdown (and unclear why that is needed?). There's also still markdown mixed in.
> 
> Using HTML there allows the content to be centered as you cannot do that with markdown.

In my opinion it looked better without the centering.
Maybe you can discuss this things in the discord so other people can also give their opinion/discuss changes? (Note: i’m not pterodactyl staff)

Jcodeerd:
I don't see a reason to center text really, have it left to right is easier to read and find stuff.

matthewpi:
Thanks for trying to contribute, but I don't see any reason to change this and it is intentionally not centered.

wdhdev:
@matthewpi I removed the centering and HTML.

PR #4586, opened 2022-11-27
Update view to support node->getConnectionAddress() call + Edit node getConnectionAddress() function to support node behind a reverse proxy

graille:
A bug appears when placing a node behind a Reverse Proxy (Traefik in my case) managing SSL certificates.

In this case, the port of the wings on the panel is 80, such has the node adress for the API was https://node.my-website/.

But, in that case Pterodactyl was constantly trying to reach https://node.my-website:80/, which, in addition to being a nonsense, raise a CORS error, a SSL error, and eventually prevents the connection between the wings instance and the panel.

Moreover, the view was recreating the node adress despite the fact that it was already done by  getConnectionAddress() method. So I updated it.

Tested on my local instance. Works like a charm.


Gauvino:
If I think you did the thing I think you did, I love you bro.

Cause I'm using CloudFlare Tunnel and my problem was that I have to open to a different port for pterodactyl like:

https://ptero-api.myptero.com:8443

and 

https://ptero.myptero.com:80

and making a little handling in the ptero.conf to make it work so if with your thing, we just have to open like one domain and pterodactyl redirect on the node you saving my life and probably life of other people.

graille:
> If I think you did the thing I think you did, I love you bro.
> 
> Cause I'm using CloudFlare Tunnel and my problem was that I have to open to a different port for pterodactyl like:
> 
> https://ptero-api.myptero.com:8443
> 
> and
> 
> https://ptero.myptero.com:80
> 
> and making a little handling in the ptero.conf to make it work so if with your thing, we just have to open like one domain and pterodactyl redirect on the node you saving my life and probably life of other people.

Well, I'm not sure this PR adress all your issues, but it will probably help, I guess x).

Gauvino:
I hope so !!

matthewpi:
I'm not quite sure I understand what you are trying to achieve with this.

I understand omitting the port if it is the schemes default (hide port `80` when `http` is being used or `443` when `https` is being used).  Is that what you are trying to achieve?  If not, please try to clarify your setup and current issues so I can better understand the problems you are facing.

graille:
> I'm not quite sure I understand what you are trying to achieve with this.
> 
> I understand omitting the port if it is the schemes default (hide port `80` when `http` is being used or `443` when `https` is being used). Is that what you are trying to achieve? If not, please try to clarify your setup and current issues so I can better understand the problems you are facing.

This is exactly what I'm trying to do. When :80 is explicitly used at the end of an URL when using a Reverse proxy, it just doesn't work.

matthewpi:
> > I'm not quite sure I understand what you are trying to achieve with this.
> > I understand omitting the port if it is the schemes default (hide port `80` when `http` is being used or `443` when `https` is being used). Is that what you are trying to achieve? If not, please try to clarify your setup and current issues so I can better understand the problems you are facing.
> 
> This is exactly what I'm trying to do. When :80 is explicitly used at the end of an URL when using a Reverse proxy, it just doesn't work.

Ah.  Then instead of always just checking for port 80, you should check for scheme http and port 80, and scheme https and port 443.  So your code would look like `if ((scheme === 'http' && port === 80) || (scheme === 'https' && port === 443)) { return scheme and host } else { return sprintf with scheme, host, and port }`.

graille:
So, yes, it would work, I'll update that PR to take your remark into account.

But after a long thought, I think the core issue is way more tricky.

When you configure your node container, if you know that you are going to put it behind a reverse proxy, you have several choices (taking the example of Traefik)
- Set the node API port (swing) to 80 (not 443 since it doesn't handle SSL certificates directly, it's up to the reverse proxy to do that), and tell the reverse proxy to point https://node.mywebsite.com to the node container
- Set the node API port (swing) to any port, for example 8080, and add a load balancing option on Traefik to indicate that https://node.mywebsite.com is actually pointing to the container port 8080.

Moreover, Traefik has the ability to redirect the requests from a port to another.

In any case, in the end, you only talk to https://node.mywebsite.com, there is no port involved anymore. In my example I just made a little hack to make it work. But the problem seems to be bigger than just removing the ports when talking with the node. 

Finally, when creating a configuration, the config.yml configuration proposed inside the panel specify the port the panel want to make requests to, not the port the wing is listening to...

**Maybe the real solution would be to completely disassociate the port the node listens to, from the port the panel makes requests to in order to be fully compatible with reverse proxies.**

I'll have a deeper thought on that and maybe come back with some idea in the future.

matthewpi:
> So, yes, it would work, I'll update that PR to take your remark into account.
> 
> But after a long thought, I think the core issue is way more tricky.
> 
> When you configure your node container, if you know that you are going to put it behind a reverse proxy, you have several choices (taking the example of Traefik)
> 
>     * Set the node API port (swing) to 80 (not 443 since it doesn't handle SSL certificates directly, it's up to the reverse proxy to do that), and tell the reverse proxy to point https://node.mywebsite.com to the node container
> 
>     * Set the node API port (swing) to any port, for example 8080, and add a load balancing option on Traefik to indicate that https://node.mywebsite.com is actually pointing to the container port 8080.
> 
> 
> Moreover, Traefik has the ability to redirect the requests from a port to another.
> 
> In any case, in the end, you only talk to https://node.mywebsite.com, there is no port involved anymore. In my example I just made a little hack to make it work. But the problem seems to be bigger than just removing the ports when talking with the node.
> 
> Finally, when creating a configuration, the config.yml configuration proposed inside the panel specify the port the panel want to make requests to, not the port the wing is listening to...
> 
> **Maybe the real solution would be to completely disassociate the port the node listens to, from the port the panel makes requests to in order to be fully compatible with reverse proxies.**
> 
> I'll have a deeper thought on that and maybe come back with some idea in the future.

I separated the ports in the v2 version of the panel.  For now, you can just manually edit the `config.yml` to use whatever port you need, but make sure to set the public facing port on the panel first, otherwise it could overwrite the one you specified in the `config.yml`.

Separating the ports will solve the reverse proxy problem, however if you want to for example use separate URIs for Panel -> Node communication vs User -> Node communication, nothing listed here will solve that; which should be fine for 99% of users.

PR #4603, opened 2022-12-08
Fix mail settings not saving

### LINKED ISSUE #4602: Mail settings will not be saved
Hey, since the new update of the Ptero panel, the mail settings will not be saved anymore :( I tried it multiple times, the default settings will be restored anytime. Please fix that!

Boy132:
With the recent Laravel 9 update the config names for the mail settings changed. https://github.com/pterodactyl/panel/commit/e49ba657094fc009387b567fe5d8c8a97f4a40ee partly dealt with it.
The current problem is that the [config values aren't set](https://github.com/pterodactyl/panel/blob/develop/app/Providers/SettingsServiceProvider.php#L105) because the [key names](https://github.com/pterodactyl/panel/blob/develop/app/Providers/SettingsServiceProvider.php#L39:L46) no longer match the config values. So the correct values are saved in the `settings` table on the database but the actual config value is never set.
This also results in displaying the "wrong" values on the panel when editing the mail settings. ("Mail settings don't save" problems)

I'm not sure about backwards compatibility tho.
I tested it on my panel and it worked. But that was probably because I also set the values in my `.env` file, so it pulled the right defaults for me.
Maybe it would be better to edit the `SettingsServiceProvider`'s `boot` function to transform the key values?

Fixes #4602

### DIFF
--- a/app/Http/Controllers/Admin/Settings/MailController.php
+++ b/app/Http/Controllers/Admin/Settings/MailController.php
@@ -57,8 +57,8 @@ class MailController extends Controller
         }
 
         $values = $request->normalize();
-        if (array_get($values, 'mail:password') === '!e') {
-            $values['mail:password'] = '';
+        if (array_get($values, 'mail:mailers:smtp:password') === '!e') {
+            $values['mail:mailers:smtp:password'] = '';
         }
 
         foreach ($values as $key => $value) {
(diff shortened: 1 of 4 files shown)


matthewpi:
Hey, thanks for this.  I forgot about this before I started making all the changes myself.

https://github.com/pterodactyl/panel/commit/63a179d7f3e80e9c4c930fbaada7ea42d9467022
https://github.com/pterodactyl/panel/commit/4626118d77e03244d51ecc7e0f641080b79bccee

Issue #2483, opened 2020-10-12
A separate server address for SFTP in node condig


Doc94:
**Is your feature request related to a problem? Please describe.**
Actually for the conection of nodes i use a subdomain in cloudflare, this subdomain are in proxy, then only work with SSL in for HTTP requests, in the case of SFTP details show the same subdomain but not work in conections for the proxy.

**Describe the solution you'd like**
A optional variable for sftp server address in any node. in cases where the node is in proxy then can set a sftp connection (like ftp.domain.com or the IP).

Homie20006:
I can totally relate to this.
Exact same setup and users can't use the visible node address for SFTP.

This extra option for a custom address would simply resolve the problem. 👍

elitetheespeon:
Adding to this: I have a 1:1 NAT setup with internal IPs assigned to external IPs and while I can alias the game servers IP addresses, SFTP still shows the local address in the panel to users. Being able to customize this as an IP or FQDN would be helpful.

TekExplorer:
What if its akin to the aliases for ip:port allocations? or would it be better to have that specified in the node config?

Doc94:
> What if its akin to the aliases for ip:port allocations? or would it be better to have that specified in the node config?

Delay...

aliases for ip:port allocations not think can be great.. because for example in my case i use the aliases for set a SRV DNS for the game and this not work with the SFTP. its better a aliase for SFTP connection in node.

maxlingenfelter:
This would be a very helpful feature. Currently, the only way to have a proxied node but allow sftp that I know of is to modify the panel's code to point to a different place for sftp.

BIackHornet:
This is a needed feature for those using free Cloudflare proxy services. Please give us an alternate SFTP hostname field that will change the web links and perhaps work with the transfer server feature.

ghost:
Agreed this feature is needed, especially with some reverse proxies where the SFTP address is already on a different port.

tomlyo:
Just wanted to chime in to say that this is still very relevant, I'm using Cloudflare as well as many others in here who could benefit from specifying an SFTP URL.

maxlingenfelter:
> Just wanted to chime in to say that this is still very relevant, I'm using Cloudflare as well as many others in here who could benefit from specifying an SFTP URL.

I completely agree and could really use this feature!

tschlienger:
[...]

PR #4607, opened 2022-12-11
Adding option to sftp ip on node.

mundotv789123:
This option is because when someone use cloudflare proxy on node the sftp connection is impossible.


Doc94:
If any maintainer can link this PR to the issue #2483 (?

Northwave:
Would love to see this in. Actually helped me to find and do the change in ServerTransformer.php manually

drysius:
This helps other problems too, such as the dashboard schedule that when the cloudflare proxy is used, the schedule stops working because it needs sftp, which ends up giving error in the free cloudflare plan that only allows a proxy port "8443".This is just a suggestion but I think it was very strange to get ip on the side of the door, in my case I thought it best to change to "FQDN Sftp" and "FQDN Wings" 

image

Juloos:
Any updates ? That doesn't seems really hard to implement and yet very useful for a better workflow control

PR #4608, opened 2022-12-11
Update security.md

Jcodeerd:
This PR updates the values that are supported to the latest releases


Jcodeerd:
Not the MB to mib ones...

matthewpi:
For future reference, don't make Revert commits on a separate branch.  Always make sure you have a clean tree to make changes from, otherwise you complicate the commit history if it is merged without being squashed.

I made these changes myself since the changes you made don't take into account that the previous release series is still supported.

PR #4667, opened 2023-01-27
add: server startup password as variable

Panda260:
# Description

you can now start yout teamspeak server with a serverquery password defined under startup


parkervcp (commented):
It would be better if you exported the egg from the panel instead of a manual edit.

Panda260:
> It would be better if you exported the egg from the panel instead of a manual edit.

okey and what difference would that make? just asking

parkervcp:
It makes sure the entire egg gets updated and time stamped.

Panda260:
?

parkervcp (approved):

matthewpi:
This PR needs to go into `1.0-develop` not `develop`.

Panda260:
should i open a new PR?

PR #4669, opened 2023-01-28
Wait timeout decompression for Cloudflare

NotYusta:
Source: https://community.cloudflare.com/t/default-timeout-limit-increase/240042

Cloudflare only has 100 seconds timeout limit, so if you are proxying the panel with cloudflare, decompressing a large archive would be a problem. and this happens to me several times, so i have to access the VPS to extract the archive.

I have a custom panel myself, committing a change for a help.

### DIFF
--- a/app/Http/Controllers/Api/Client/Servers/FileController.php
+++ b/app/Http/Controllers/Api/Client/Servers/FileController.php
@@ -192,7 +192,7 @@ class FileController extends ClientApiController
      */
     public function decompress(DecompressFilesRequest $request, Server $server): JsonResponse
     {
-        set_time_limit(300);
+        set_time_limit(99);
 
         $this->fileRepository->setServer($server)->decompressFile(
             $request->input('root'),


Aeris1One:
Uh, I think this limit should be 99 only when there's a Cloudflare proxy in use. IMO, by default it should stay 300. 

Maybe a configuration option like "BEHIND_CLOUDFLARE" would be better

devnote-dev:
@Aeris1One Agreed, merging this PR runs the risk of breaking/disrupting many panel's requests to Wings for decompression just because it's slightly longer than 99 seconds.

Nodes have a `behind_proxy` option which could be used here instead, but that doesn't necessarily solve the issue either. I think it would be better if the Wings endpoint for this were asynchronous given that it doesn't return anything anyway, and after the initial checks the only errors that can be returned are decompression failure errors which can be streamed into the server websocket instead.

danny6167:
As mentioned this PR has too much impact on those who aren't using Cloudflare and doesn't really provide any real good benefits.
Long term, making the API asynchronous really makes the most sense to me.

PR #4733, opened 2023-03-24
fix(admin): settings routes

itsvic-dev:
The settings routes did not properly render due to their path properties not being relative.

This PR fixes that by making the paths in the routes relative.
image


robertdrakedennis:

PR #4734, opened 2023-03-25
Add support for Carbon, a rust modding framework

jbrazio:
Not to create additional vars, one for each mod and then needing to validate that the user didn't activated both of them at the same time.. the existing Oxide variable was renamed to "Modding Framework" and accepts three values: "carbon", "oxide" and "vanilla" (default). The names should be self-explanatory about which framework will get deployed on the server instance.

The `entrypoint.sh` file on the rust's docker container was also updated to allow the selection of the modding framework to be used. Carbon uses `UnityDoorstop` as the injector thus it needs additional env context to be setup before starting the dedicated server.

### DIFF
--- a/database/Seeders/eggs/rust/egg-rust.json
+++ b/database/Seeders/eggs/rust/egg-rust.json
@@ -4,7 +4,7 @@
         "version": "PTDL_v2",
         "update_url": null
     },
-    "exported_at": "2023-02-04T14:45:12-05:00",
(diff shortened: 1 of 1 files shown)


matthewpi (approved):

matthewpi:
Added in https://github.com/pterodactyl/panel/commit/29783ed24002168bc939f4e07b283db0dd3fde54 to 1.0-develop

PR #4744, opened 2023-04-04
Create a Database Host with artisan

Thomas5300:
This is the same idea as creating a node via artisan, but for a database host as this didn't exist yet and can be quite useful sometimes.


Boy132:
Duplicate of #4062

Thomas5300:
There is a bug in that PR code that the password gets the host instead of the password and some people don't always want their database to be linked to a node and the that code can't do that
`$data['password'] = $this->option('host') ?? $this->secret('Enter the password for the defined account');`

This PR has solved all that.

Boy132:
You should leave a note/ review on the other PR and not make a duplicate.

PR #4751, opened 2023-04-13
Update index.php

CodingRule:
I removed the comments that were unnecessary or redundant. I also reorganized the code to follow a more logical flow. Finally, I simplified some of the variable assignments to make the code more concise.

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -5,69 +5,25 @@
  *
  * @author   Taylor Otwell <taylor@laravel.com>
  */
+
+// Set the start time for the script
 define('LARAVEL_START', microtime(true));
 
-/*
-|--------------------------------------------------------------------------
-| Check If Application Is Under Maintenance
-|--------------------------------------------------------------------------
-|
-| If the application is maintenance / demo mode via the "down" command we
-| will require this file so that any pre-rendered template can be shown
-| instead of starting the framework, which could cause an exception.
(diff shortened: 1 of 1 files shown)


public/index.php
```diff
@@ -5,69 +5,25 @@
...
-*/
-
+// Handle the incoming request and send the response back to the browser
 $kernel = $app->make(Illuminate\Contracts\Http\Kernel::class);
-
-$response = $kernel->handle(
-    $request = Illuminate\Http\Request::capture()
-);
-
+$response = $kernel->handle(Illuminate\Http\Request::capture());
```
devnote-dev, line 25: Have you tested this? `$request` isn't a named parameter, it's a variable defined in the method call which is used below.

devnote-dev (commented):
I don't think this PR is actually necessary, it can be quite useful to people familiarizing themselves with Pterodactyl's codebase. Regardless, one small point below.

public/index.php
CodingRule: ohhh

Jcodeerd:
I don't see the reason to remove most of the documentation/comments in the file.

matthewpi:
This is copied directly from Laravel's template.  I don't see any reason to change any of this, we shouldn't be modifying this file anyways.

PR #4754, opened 2023-04-17
Make wording in prune-backup command clearer

Boy132:
This makes the wording of prune-backup command description a bit clearer.

### DIFF
--- a/app/Console/Commands/Maintenance/PruneOrphanedBackupsCommand.php
+++ b/app/Console/Commands/Maintenance/PruneOrphanedBackupsCommand.php
@@ -10,7 +10,7 @@ class PruneOrphanedBackupsCommand extends Command
 {
     protected $signature = 'p:maintenance:prune-backups {--prune-age=}';
 
-    protected $description = 'Marks all backups that have not completed in the last "n" minutes as being failed.';
+    protected $description = 'Marks all backups older than "n" minutes that have not yet completed as being failed.';
 
     /**
      * PruneOrphanedBackupsCommand constructor.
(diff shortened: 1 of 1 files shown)


Aeris1One:
Pruning backups that are older than the pruning date is what it is intended for 🤔 
The goal of pruning is the get rid of older backups...

Boy132:
> Pruning backups that are older than the pruning date is what it is intended for 🤔 The goal of pruning is the get rid of older backups...

Not according the command description:
> Marks all backups that have not completed in the last "n" minutes as being failed.

Aeris1One:
Oh, sorry, misunderstood what you were talking about.

> Marks all backups that have not completed in the last "n" minutes as being failed.

could be rephrased as
> Marks all backups that have not completed and were created more that "n" minutes ago as being failed.

I agree the wording is kind of bad, but the command is about marking as failed all backups that are stuck for more than "n" minutes. 
I can't see the point of a command that mark all backups started less than "n" minutes ago and aren't completed as failed. That would mean that even backup that are not failed and still ongoing (because launched some seconds ago maybe, so perfectly normal they are still running) would be marked as failed.

Boy132:
> I agree the wording is kind of bad, but the command is about marking as failed all backups that are stuck for more than "n" minutes. I can't see the point of a command that mark all backups started less than "n" minutes ago and aren't completed as failed. That would mean that even backup that are not failed and still ongoing (because launched some seconds ago maybe, so perfectly normal they are still running) would be marked as failed.

I see your point. And I agree that makes more sense. I changed the PR to only change the wording so it's clearer.
Thanks for pointing that out!

PR #4761, opened 2023-04-21
do not show strings with password in notification text

djohts:


Boy132:
The password is shown on the screen anyways, so why hide it in the little notification?

djohts:
> The password is shown on the screen anyways, so why hide it in the little notification?

at least because the notification doesn't look really good on phones when copying JDBC connection string

matthewpi (approved):

PR #4781, opened 2023-05-14
Fix file uploads when using cloudflare access

liorsl:
When wingsl sits behind Cloudflare access (a zero trust solution offered by Cloudflare), file uploads from panel are broken. The reason for that is that axios doesn't send Cloudflare's own cookies, which makes them send 302 to their own login page.
Since wings is deemed trusted I don't think there is a problem with passing cookies to it, so I had to set ``withCredentials`` to true in the request options. Alternatively we could add a list of cookies to pass along to wings, to be set by the user in the .env file.


robertdrakedennis:
Will be looking at this later to see if this is still an issue

liorsl:
@robertdrakedennis appreciate it, honestly I haven't used this setup recently, but if there is interest in this PR I'll be happy to fix it up since it seems something went wrong with the git history

PR #4788, opened 2023-05-25
fix(4752): Can't set server description to empty from client UI

### LINKED ISSUE #4752: Can't remove server description
### Current Behavior

You can't remove the server description via the Rename button in the Settings tab. When you hit save it looks like the description got removed but that isn't the case. Refreshing the page or switching the tab shows the old value again;
gif

https://github.com/pterodactyl/panel/commit/e43da311fe83d32bf04e16888a8bbb51d22b0ad2 causes this.

### Expected Behavior

Correctly remove the description.

### Steps to Reproduce

1. Set the description of a server.
2. Try to remove it via the "Rename" button in the Settings tab
3. Refresh the page and see that the description is still there

### Panel Version

1.11.3

### Wings Version

1.11.5

_No response_

### Docker Image

_No response_

### Error Logs

_No response_

robertnisipeanu:
Fixes #4752 

Issue is that the middleware `ConvertEmptyStringsToNull` is enabled by default by Laravel.

Request:
image
Values:
image

New behavior after fix:
If description is not included in the request it will keep the old value.
If description is included but has a value of `null` or empty string it gets converted to `null` by the middleware, and then casted to empty string using a `(string)` cast.
If description is included and not empty, it will save that value.

### DIFF
--- a/app/Http/Controllers/Api/Client/Servers/SettingsController.php
+++ b/app/Http/Controllers/Api/Client/Servers/SettingsController.php
@@ -35,7 +35,7 @@ class SettingsController extends ClientApiController
     public function rename(RenameServerRequest $request, Server $server): JsonResponse
     {
         $name = $request->input('name');
-        $description = $request->input('description') ?? $server->description;
+        $description = $request->has('description') ? (string) $request->input('description') : $server->description;
         $this->repository->update($server->id, [
             'name' => $name,
             'description' => $description,


Mubeen142 (commented):
The suggested code can be simplified by doing

```php
$request->input('description', $server->description);
```

Instead of

```php
$request->has('description') ? (string) $request->input('description') : $server->description
```

Boy132:
> The suggested code can be simplified by doing
> 
> ```
> $request->input('description', $server->description);
> ```
> 
> Instead of
> 
> ```
> $request->has('description') ? (string) $request->input('description') : $server->description
> ```

No. With `$request->input('description', $server->description);` you still can't set the description to empty. robertnisipeanu's fix works fine.

PR #4804, opened 2023-06-14
Update StartupContainer.tsx

NotYusta:
Only show variables title or text when there is variables available.


QuintenQVD0:
your to branch should be 1.0-develop

PR #4805, opened 2023-06-14
Update StartupContainer.tsx

NotYusta:
Update variables title visibility, it will only be shown if there are variables available.

### DIFF
--- a/resources/scripts/components/server/startup/StartupContainer.tsx
+++ b/resources/scripts/components/server/startup/StartupContainer.tsx
@@ -124,7 +124,7 @@ const StartupContainer = () => {
                     )}
                 </TitledGreyBox>
             </div>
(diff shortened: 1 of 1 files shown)


resources/scripts/components/server/startup/StartupContainer.tsx
```diff
@@ -124,7 +124,7 @@ const StartupContainer = () => {
                     )}
                 </TitledGreyBox>
             </div>
-            <h3 css={tw`mt-8 mb-2 text-2xl`}>Variables</h3>
+            {data.variables.length != 0 && <h3 css={tw`mt-8 mb-2 text-2xl`}>Variables</h3>}
```
matthewpi: Please use `> 0`.  Also, if you are ever writing comparisons in the future, please use type-strict comparisons (`!==`, `===`, etc)

Issue #4723, opened 2023-03-14
CSRF required even with API key


LindaJuffermans:
### Current Behavior

In a frontend (browser) application, `POST https://pterodactyl.file.properties/api/client/servers/1a7ce997/schedules/2` to activate/deactivate a schedule. Include a valid API key in the headers.

returns a 419 status with "CSRF token mismatch."

### Expected Behavior

The API call should be successfull.

A CSRF threat exists when the authentication validation is based on a cookie; a 3rd party website can POST to the API with the browser automatically sending the (authentication) cookie details.
If authentication validation is based on a header (API key), this is not automatically injected by the browser and cannot be used by a 3rd party website to piggyback onto.

### Steps to Reproduce

In a frontend (browser) application, `POST https://pterodactyl.file.properties/api/client/servers/1a7ce997/schedules/2` to activate/deactivate a schedule. Include a valid API key in the headers.
Doing this through CURL or Postman is not a valid test as it doesn't inject the Origin header which a browser should always send.

### Panel Version

1.11.3

### Wings Version

1.11.0

### Games and/or Eggs Affected

Not relevant

### Docker Image

Not relevant

### Error Logs

```bash
Request:

POST /api/client/servers/8a16850b/schedules/284 HTTP/1.1
Accept: Application/json
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US,en;q=0.9,nl;q=0.8
Authorization: Bearer <removed>
Cache-Control: no-cache
Connection: keep-alive
Content-Length: 108
Content-Type: application/json
Host: <removed>
Origin: http://localhost:3000
Pragma: no-cache
Referer: http://localhost:3000/
Sec-Fetch-Dest: empty
Sec-Fetch-Mode: cors
Sec-Fetch-Site: cross-site
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/110.0.0.0 Safari/537.36
sec-ch-ua: "Chromium";v="110", "Not A(Brand";v="24", "Google Chrome";v="110"
sec-ch-ua-mobile: ?0
sec-ch-ua-platform: "Windows"

Body:
{
    "name": "Restart",
    "minute": "30",
    "hour": "05,11,17,23",
    "day_of_month": "*",
    "day_of_week": "*",
    "is_active": false
}

Response:
{
    "errors": [
        {
            "code": "HttpException",
            "status": "419",
[...]

PR #4837, opened 2023-08-12
fixing CSRF required even with API key

filpecosta:


devnote-dev:
This PR would remove CRSF checks for browsers entirely which is a big security issue. I would also like to point out (as previously mentioned from the initial discussion in the Discord server) that **you should not be making such requests from the browser if you do not want to deal with CSRF/cookies.** Using API keys for requests in the browser is _not_ the way to go about structuring your application, and there are more than enough resources available on how to do so properly. I don't believe #4723 is valid here and should be closed as it is not a panel issue.

filpecosta:
Actually in my case I don't use the panel for the public, but without it, the error is occurring not in the Browser, but in the API call even in Postman :(

filpecosta:
Actually in my case I don't use the panel for the public, but without it, the error is occurring not in the Browser, but in the API call even in Postman :(

ghost:
> Actually in my case I don't use the panel for the public, but without it, the error is occurring not in the Browser, but in the API call even in Postman :(

If you don't use it in public, you can still use your modified Version of the Panel, just keep it in your Fork, if People actually need this Change, they can just build the Panel from your Fork-Source, because the way you are implementing it presents a serious security issue since... well People sometimes (surprisingly) update their Software.

Sewdohe:
I am also having this issue when making API requests from postman as well as the curl application. Is this the only fix?

PR #4869, opened 2023-09-22
fix(imports): admin ui button imports

alexevladgabriel:
This change made the UI styles feel more consistent.

Before:
image
After:
image


voidvalue-dev:
@DaneEveritt The files don't even exist anymore. Shouldn't it be closed?

PR #4872, opened 2023-09-25
feat(node-auto-deploy): Node Auto Deploy

alexevladgabriel:
Screenshots:
image
image

Preview:
https://www.loom.com/share/9535e5445d014d89b9791b440eefab4b?sid=974afd7b-c0b2-4688-be00-d2f9df059672

### DIFF
--- a/app/Http/Controllers/Api/Application/Nodes/NodeAutoDeployController.php
+++ b/app/Http/Controllers/Api/Application/Nodes/NodeAutoDeployController.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace Pterodactyl\Http\Controllers\Api\Application\Nodes;
+
+use Illuminate\Http\Request;
+use Pterodactyl\Models\Node;
+use Pterodactyl\Models\ApiKey;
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Api/Application/Nodes/NodeAutoDeployController.php
```diff
@@ -0,0 +1,64 @@
...
+                'memo' => 'Automatically generated node deployment key.',
+                'allowed_ips' => [],
+            ], ['r_nodes' => 1]);
+        }
+
+        return new JsonResponse([
+            'debug' => config('app.debug'),
+            'node' => $node->id,
+            'token' => $key->identifier . $this->encrypter->decrypt($key->token),
+            'remote' => route('index'),
```
Boy132: IMO it's better to use the `APP_URL` as the remote

alexevladgabriel: Yes, we could, I've seen `route('index')` used by Matthew in the past. I will do a commit with new changes.

PR #4875, opened 2023-09-28
fix: navigation link

alexevladgabriel:
Due to using `a` tag, made the page reload between pages.
Sincerely it's annoying to see the white blank page when working in the night.

### DIFF
--- a/resources/scripts/components/NavigationBar.tsx
+++ b/resources/scripts/components/NavigationBar.tsx
@@ -77,9 +77,9 @@ export default () => {
 
                     {rootAdmin && (
                         <Tooltip placement="bottom" content="Admin">
-                            <a href="/admin" rel="noreferrer">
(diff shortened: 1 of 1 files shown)


AidenOnyenwenu:
> Due to using `a` tag, made the page reload between pages. Sincerely it's annoying to see the white blank page when working in the night.

Good idea 👀

rmartinoscar:
Hey this is a good idea but i can't do it that way, it will redirect to /admin on the client side which is a 404, i'd have to add something like [this](https://reactrouter.com/en/main/components/link#reloaddocument) which requires a react update.

matthewpi:
For anyone commenting on this PR:

This is a change for v2 where the admin area has been rewritten in React from PHP Blade templates.  This change will break navigation to the admin area in v1 as the page needs to be reloaded for the Blade templates to be rendered.

PR #4890, opened 2023-10-12
Fixed sum up issue with node viability function

UndercoverNL:
Fixed sum up issue with node viability function, in which `sum_memory` and `sum_disk` always where `0`

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -11,7 +11,7 @@ return [
     | change this value if you are not maintaining your own internal versions.
     */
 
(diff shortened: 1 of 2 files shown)


config/app.php
```diff
@@ -11,7 +11,7 @@
     | change this value if you are not maintaining your own internal versions.
     */
 
-    'version' => 'canary',
+    'version' => '1.11.4',
```
Boy132: The version should stay on `canary`

PR #4905, opened 2023-10-22
Update CHANGELOG.md

Rushaway:
Update changelog related to **v1.11.5**


Boy132:
`1.0-develop` already has the changelog. So there is no need to change it on `develop`.

PR #4916, opened 2023-11-07
Change Avatar.tsx to use gravatar instead of BoringAvatar

Miniontoby:
Hi,

Using this PR, the `<Avatar.User>` will now use gravatar for the avatar instead of BoringAvatar avatar.

Gravatar is already used in the admin view, so why not add it to the regular view as well?

Kind regards,
- Miniontoby

### DIFF
--- a/resources/scripts/components/Avatar.tsx
+++ b/resources/scripts/components/Avatar.tsx
@@ -1,5 +1,6 @@
 import BoringAvatar, { AvatarProps } from 'boring-avatars';
 import { useStoreState } from '@/state/hooks';
+import crypto from 'crypto';
 
 const palette = ['#FFAD08', '#EDD75A', '#73B06F', '#0C8F8F', '#587291'];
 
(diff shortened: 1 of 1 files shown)


yesBad:
It was literally just changed to not use Gravatar as they are migrating away from it.

Miniontoby:
Oh I didn't know that

PadowYT2 (commented):
`crypto` is not installed by default and you have made a PR to the wrong branch

Miniontoby:
> `crypto` is not installed by default and you have made a PR to the wrong branch

Uhmm, `crypto` is a built-in nodejs package

And about the branch, I guess the develop branch shouldn't be the 'default' branch, then...

And yeah as yesBad said, it will not be added, because they are removing it... So I am going to close this...

PadowYT2:
> > `crypto` is not installed by default and you have made a PR to the wrong branch
> 
> Uhmm, `crypto` is a built-in nodejs package

My bad, thought of `crypto-js`

PadowYT2:
> And about the branch, I guess the develop branch shouldn't be the 'default' branch, then...

Well it is Pterodactyl v2 and currently is the main development version, but this kind of feature should be done in `1.0-develop`

Miniontoby:
But if they want to move off gravatar, then should I still open a new one for the `1.0-develop` or not?

Miniontoby:
> It was literally just changed to not use Gravatar as they are migrating away from it.

what will be the new plan? Or will it just all be the `BoringAvatar`'s?

TekExplorer:
> > It was literally just changed to not use Gravatar as they are migrating away from it.
> 
> what will be the new plan? Or will it just all be the `BoringAvatar`'s?

You can always re-implement Gravatar in your own fork of the panel if you want it. As far as i understand it, `BoringAvatar` is just what it's going to be. (at least in the client (public-facing) side)

You can always change it for your own panel ;)

PR #3821, opened 2021-12-18
Add sort by file size/date modified

mrxbox98:
This closes https://github.com/pterodactyl/panel/issues/845 by adding sorting for date last modified and size. This also adds being able to reverse the search. (e.i. z->a instead of a->z).

PR #4951, opened 2023-12-18
Server Sorting: Sort servers from A-Z or Z-A.

nullure:
Be able to sort servers from A-Z or Z-A which makes navigating servers easier.

### DIFF
--- a/resources/scripts/components/dashboard/DashboardContainer.tsx
+++ b/resources/scripts/components/dashboard/DashboardContainer.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { Server } from '@/api/server/getServer';
 import getServers from '@/api/getServers';
 import ServerRow from '@/components/dashboard/ServerRow';
@@ -13,6 +13,7 @@ import useSWR from 'swr';
 import { PaginatedResult } from '@/api/http';
 import Pagination from '@/components/elements/Pagination';
 import { useLocation } from 'react-router-dom';
+import Select from '../elements/Select';
 
 export default () => {
     const { search } = useLocation();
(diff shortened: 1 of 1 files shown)


QuintenQVD0:
IF this is a PR for the current releases then it is the wrong branch develop is v2, 1.0-develop are the current releases

PadowYT2:
Also seems to be a duplicate of pterodactyl/panel#3821

resources/scripts/components/dashboard/DashboardContainer.tsx
```diff
@@ -13,20 +13,43 @@ import useSWR from 'swr';
 import { PaginatedResult } from '@/api/http';
 import Pagination from '@/components/elements/Pagination';
 import { useLocation } from 'react-router-dom';
+import Select from '../elements/Select';
```
PadowYT2, line 16: You should be using proper paths, e.g.
```tsx
import Select from '@/components/elements/Select';
```

resources/scripts/components/dashboard/DashboardContainer.tsx
```diff
@@ -13,20 +13,43 @@ import useSWR from 'swr';
...
 export default () => {
     const { search } = useLocation();
     const defaultPage = Number(new URLSearchParams(search).get('page') || '1');
 
     const [page, setPage] = useState(!isNaN(defaultPage) && defaultPage > 0 ? defaultPage : 1);
     const { clearFlashes, clearAndAddHttpError } = useFlash();
-    const uuid = useStoreState(state => state.user.data!.uuid);
-    const rootAdmin = useStoreState(state => state.user.data!.rootAdmin);
+    const uuid = useStoreState((state) => state.user.data!.uuid);
+    const rootAdmin = useStoreState((state) => state.user.data!.rootAdmin);
```
PadowYT2, line 25: Should be reverted I believe, just use prettier

resources/scripts/components/dashboard/DashboardContainer.tsx
```diff
@@ -58,8 +81,14 @@ export default () => {
                     <Switch
                         name={'show_all_servers'}
                         defaultChecked={showOnlyAdmin}
-                        onChange={() => setShowOnlyAdmin(s => !s)}
+                        onChange={() => setShowOnlyAdmin((s) => !s)}
                     />
+                    <div css={tw`mx-3`}>
```
PadowYT2, line 86: Shouldn't it be `ml-3`? How does it currently look?

Boy132:
~~There is already an open PR for this functionality: #3821 (which also adds more sorting modes)~~

PadowYT2:
> There is already an open PR for this functionality: #3821 (which also adds more sorting modes)

That's what I wanted to say, but the PR is for sorting servers, not files

Boy132:
> > There is already an open PR for this functionality: #3821 (which also adds more sorting modes)
> 
> That's what I wanted to say, but the PR is for sorting servers, not files

Ah, that's my bad. Then nevermind my comment.

TekExplorer:
Isn't sorting server-side?

PadowYT2:
> Isn't sorting server-side?

There's no sorting in the server list. Well there is, just that it is hardcoded

PadowYT2:
Althougjh I don't really like to have this. The current order system will be just messed up

TekExplorer:
No, I mean that most sorting in the api is done via query parameters and done serverside. [this code](https://github.com/pterodactyl/panel/blob/develop/app/Http/Controllers/Api/Client/ClientController.php) would need to add sorting options to do this correctly.

Notice [here in the application side.](https://github.com/pterodactyl/panel/blob/develop/app/Http/Controllers/Api/Application/Servers/ServerController.php#L48) This is how sorting is handled, and any pr that adds new sorting options should do the same, or else you will end up with local sorts, which will miss items on the next pages that should now be front and center.

nullure:
Is this mergable? Or it needs improvements

resources/scripts/components/dashboard/DashboardContainer.tsx
nullure: Looks fine.

PadowYT2: Need a screenshot. Because the switch doesn't have a margin right (without your PR) and you added it to your element

TekExplorer:
No. As I said, sorting is supposed to be server-side. Refer to my previous comment

PR #4959, opened 2023-12-31
fix: gravatar url

alexevladgabriel:
Self explanatory
Used: https://docs.gravatar.com/gravatar-images/php/

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -222,7 +222,7 @@ class User extends Model implements
     public function avatarUrl(): Attribute
     {
         return Attribute::make(
-            get: fn () => 'https://www.gravatar.com/avatar/' . $this->md5 . '.jpg',
+            get: fn () => 'https://www.gravatar.com/avatar/' . $this->md5,
         );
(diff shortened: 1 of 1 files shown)


QuintenQVD0:
This branch is out-of-date with the base branch

and if you want to fix this for the v1 releases then you selected the wrong to branch.  if it is for v2 then it is fine

Boy132:
Won't gravatar be removed in v2 anyways? At least that was Danes plan I believe.

alexevladgabriel:
It's the correct branch, it's an update for Admin UI. Gravatar is used on the sidebar, and it's not removed.
image

Boy132:
> [...] Gravatar is used on the sidebar, and it's not removed.

Not yet but Danes plan was to remove gravatar entirely. That's why the client area already uses boring avatars.

prplwtf:
> [...] That's why the client area already uses boring avatars.

Subusers still appear to be using Gravatar, might have been a oversight when switching to boring avatars.

PR #5055, opened 2024-04-04
feature: Closes #1831 - Add an artisan command to create an API key from the command line

### LINKED ISSUE #1831: Ability to create an application API key from the CLI
I would like to automate the creation of an application API key. This automation is used when deploying a new Pterodactyl installation with an external application that communicate with the API. The ability to automatically create an application API key is to avoid asking the user to manually generate an application API key and make the external application integration with Pterodactyl seamless to the user.

An artisan command like `php artisan p:application-api:create-key`.
To make it easy to release this feature, a first implementation would create an application API key with all permissions.

There is currently none.

GreatSymphonia:
This PR add an artisan command that allows for someone to create an API key from the command line.

This allows the creation of both a (deprecated) `ApiKey::TYPE_APPLICATION` and a `ApiKey::TYPE_ACCOUNT`.

The `TYPE_APPLICATION` key may be needed in cases where someone would want to automate the deployment of the panel and Wings instances without using the Panel which is the case when you want to deploy the project using a tool like Ansible, thus why it was implemented even though Deprecated.


Boy132:
You aren't actually checking the user password.

IMO the command should only be able to create application api keys.
Also I don't think application api keys are meant to be deprecated, not sure why there is a [comment](https://github.com/pterodactyl/panel/blob/develop/app/Models/ApiKey.php#L74-L75) saying so.

GreatSymphonia:
I do understand, I am going to make the changes towards the authentication of the user.
I do agree that only application keys should be able to be created using that, I was confused with the "deprecated" statement regarding them, I will make changes regarding that too.

GreatSymphonia:
Actually, I am questionning if I should be authenticating the user here because of the fact that it's especially for the admin panel on the installation.  It will most likely used right after the creation of the admin user for the Panel and would assing the token to it.

If I am loggin in the user, should I use the `Pterodactyl\Http\Controllers\Auth\LoginController` or should I simply validate the password in a login request?  I am really trying to learn about best practices for that PR.

Boy132:
> Actually, I am questionning if I should be authenticating the user here because of the fact that it's especially for the admin panel on the installation. It will most likely used right after the creation of the admin user for the Panel and would assing the token to it.

Yeah, for just application api keys I don't see a reason for any auth.

danny6167:
> Also I don't think application api keys are meant to be deprecated, not sure why there is a [comment](https://github.com/pterodactyl/panel/blob/develop/app/Models/ApiKey.php#L74-L75) saying so.

Don't know if new management has decided to reverse the decision, but they were deprecated by Dane, the comment is no mistake.
https://discord.com/channels/122900397965705216/936744899787784282/1018218699909505157

PR #5074, opened 2024-04-16
Update SECURITY.md

danny6167:
Remove wings 1.10.x from supported list as per https://discord.com/channels/122900397965705216/376449434654998539/1217554125550518363


matthewpi:
Manually made that change into 1.0-develop then updated develop as well, thanks for the reminder.

PR #5147, opened 2024-07-01
Update Java Version Model Feature to add Minecraft 1.19 error

danny6167:
Adds the `minecraft 1.19 requires running the server with java 17 or above` variant of the error message to the Java Version model


Jcodeerd:
Might also want to add the errors for 1.20 and 1.21?

danny6167:
Paper 1.19, 1.20, and 1.21 all say that same error message with the version as 1.19.
Currently waiting for my internet to behave to confirm what Vanilla says.

danny6167:
Vanilla doesn't produce this message at all and is already caught by the existing match lines. This PR is good to go as is.

PR #5157, opened 2024-07-11
Clear allocation notes on server deletion

giomxx:
Currently when a user sets notes for an allocation, it's saved to the DB, then never touched again. When the server is deleted, the notes on that allocation are preserved. 

This means that the DB can get filled with allocations with note entries, and a NULL server ID, and when the allocation eventually get's reused for another server, the note is displayed for the allocations on the new server. 

My PR fixes this by ensuring the notes are removed from the allocations during the server deletion process.


rmartinoscar:
Hey that would work for server deleted since the change but older allocations will still have their notes we need to use a migration as well, something like this should do the job

`database/migrations/2024_07_12_170235_clear_notes_on_unused_allocations.php`
```suggestion
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Support\Facades\DB;

return new class extends Migration
{
    /**
     * Run the migrations.
     */
    public function up(): void
    {
        DB::table('allocations')
            ->where('server_id', null)
            ->whereNot('notes', null)
            ->update(['notes' => null]);
    }

    /**
     * Reverse the migrations.
     */
    public function down(): void
    {
        // Reverse not needed
    }
};
```

giomxx:
@RMartinOscar Thanks! I had completely forgot about that. I've added the migration script now. 😅

app/Services/Servers/ServerDeletionService.php
```diff
@@ -77,7 +77,12 @@ public function handle(Server $server): void
                 }
             }
 
+            // clear any allocation notes for the server
+            $server->allocations->each(function ($allocation) {
```
DaneEveritt:
```suggestion
            $server->allocations()->update(['notes' => null]);
```

Doesn't execute `n+1` update queries.

database/migrations/2024_07_13_091852_clear_unused_allocation_notes.php
```diff
@@ -0,0 +1,26 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        DB::table('allocations')
+            ->where('server_id', null)
```
DaneEveritt:
```suggestion
            ->whereNull('server_id')
```

database/migrations/2024_07_13_091852_clear_unused_allocation_notes.php
```diff
@@ -0,0 +1,26 @@
...
+            ->whereNot('notes', null)
```
DaneEveritt:
```suggestion
            ->whereNotNull('notes')
```

I'd actually just remove this entirely so it can be a simpler query on the `server_id` index, it doesn't matter if we updated a bunch of rows where the `notes` value is already null anyways,.

DaneEveritt (changes requested):

DaneEveritt (approved):

DaneEveritt:
@giomxx you have some (all) failing tests.

DaneEveritt (approved):

PR #5163, opened 2024-07-18
Fixed 'Node -> Settings' Location Dropdown selecting first location after save

MackenzieMolloy:
Currently, when you click the save button with the Node Settings page open, the location drop-down displays (select) the first location (one with the lowest id) even if the node location property wasn't edited and had a different location set.

For example, if location with id 2 was set and you clicked save then the dropdown displays the location with id 1.

This bug was caused by the Laravel Old() method always returning a String whereas the location id property is always an Integer, so old and current property comparison would always return false when saving.


danny6167:
Tested, working, and ready to merge.
Thank you for your contribution.

DaneEveritt (approved):

Issue #5175, opened 2024-07-31
Application API - Not showing API keys created by other admins


JustinPooters:
### Current Behavior

When I create an Application API key from the Admin dashboard, other admin's can't see the keys I made (and I can't see theirs). I've also confirmed this on another server. Same issue.

### Expected Behavior

I'd expect on the application api page every admin sees all API keys.

### Steps to Reproduce

Create an Application API key.
Login with another user
It's not there.

### Panel Version

 1.11.7

### Wings Version

1.11.13

### Games and/or Eggs Affected

None

### Docker Image

None

### Error Logs

```bash
Not relevant.
```

### Is there an existing issue for this?

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

MackenzieMolloy:
I believe this is intentional however I agree that all Application API keys should be visible to all Admins since it does pose the potential for malicious users to generate keys without other panel admins noticing.

In the meantime, you can view the api_keys table to see all API Keys - this table does also include normal User API Keys.

JustinPooters:
That was exactly my worry. We use Pterodactyl for our game servers. If we dismiss someone who had admin rights and they still have an active API key, it could be exploited for malicious purposes.

MackenzieMolloy:
I'll make a PR later today to adjust this behaviour for you.

I'm not sure if the Pterodactyl team will merge it as, like I said, I think the current functionality is intentional.

If you don't know how to build the source files, feel free to reach out to me on Discord ('wackenzie' is my tag). 

Happy to help.

PR #5176, opened 2024-07-31
Added support for viewing and deleting of any Application API Keys

MackenzieMolloy:
This PR closes Issue #5175.

For context, I am aware that Application API Keys are deprecated in favour of Client API Keys however they are still operational within Pterodactyl and thus, not fully removed.

Currently in Pterodactyl, as an Admin, you can only view your Application API Keys on the Admin Panel. In this PR, I made it so all Application Keys are visible and deletable. The key strings are obfuscated if the key does not belong to the user viewing.

image

The reason for adding this is primarily so other admin users can be aware of and delete another admin user's Application API keys from the UI. This functionality is useful in the event of a malicious user compromising an admin account, creating some API Keys to continue their attacks and the owner of the compromised admin account being unaware of Application API Keys. In this instance, even after a password reset, the attack could continue via the Application API without the admin realising it.  

I've tested the creation and deleting of keys along with using keys via the Application API to ensure no breakages have occurred.


app/Http/Controllers/Admin/NodeAutoDeployController.php
```diff
@@ -32,8 +32,9 @@ public function __construct(
     public function __invoke(Request $request, Node $node): JsonResponse
     {
         /** @var ApiKey|null $key */
-        $key = $this->repository->getApplicationKeys($request->user())
+        $key = $this->repository->getApplicationKeys()
```
DaneEveritt: I would keep `getApplicationKeys` as supporting a `?User $user = null` parameter, or even just `?User $user` and explicitly passing `null` or the user depending on the context so it is clear we're intentionally requesting all, or only requesting user specific ones.

resources/views/admin/api/index.blade.php
```diff
@@ -1,103 +1,113 @@
...
+                        <th>Last Used</th>
+                        <th>Created</th>
+                        <th>Created by</th>
+                        <th></th>
+                    </tr>
+                    @foreach($keys as $key)
+                    <tr>
+                        <td><code>
+                                @if(Auth::user()->id != $key->user->id)
+                                {{ $key->identifier . str_repeat('*', strlen(decrypt($key->token)))}}
```
DaneEveritt: Lets just do `{{ $key->identifier }}****` no reason to do a pointless decrypt and str_repeat here. The point is made sufficiently with `****`.

DaneEveritt (changes requested):

app/Http/Controllers/Admin/NodeAutoDeployController.php
```diff
@@ -34,6 +34,9 @@ public function __invoke(Request $request, Node $node): JsonResponse
         /** @var ApiKey|null $key */
         $key = $this->repository->getApplicationKeys($request->user())
             ->filter(function (ApiKey $key) {
+                if ($key->user->id != $request->user()->id) {
```
DaneEveritt: This filtering is unnecessary now, correct?

app/Repositories/Eloquent/ApiKeyRepository.php
```diff
@@ -28,13 +28,33 @@ public function getAccountKeys(User $user): Collection
...
+            $instance->where('user_id', $user->id);
+        }
+
+        return $instance->get($this->getColumns());
+    }
+
+    /**
+     * Delete an application API key. If a user is provided, ensure it belongs to them.
+     */
+    public function deleteApplicationKey(?User $user = null, string $identifier): int
```
DaneEveritt:
```suggestion
    public function deleteApplicationKey(?User $user, string $identifier): int
```

Shouldn't have a defaulted value before a required parameter.

DaneEveritt: Also is this function defined twice in this file or am I losing it?

DaneEveritt (changes requested):

app/Repositories/Eloquent/ApiKeyRepository.php
MackenzieMolloy: Possibly, yeah - I was losing it too while trying to work out what was going on. I also need to test these changes of which I haven't gotten around to doing it just yet.

DaneEveritt (approved):

PR #5191, opened 2024-08-17
Dashboard server sorting

MrLiam2614:
The servers will now be sorted based on their description.
If it starts with "[$number]" then the number will be used. (1 to 10000)
If the description has no number or multiple server have the same one then the server will be ordered Alphabetical (A to Z) and putted on the bottom of the List

Example:
image

### DIFF
--- a/resources/scripts/components/dashboard/DashboardContainer.tsx
+++ b/resources/scripts/components/dashboard/DashboardContainer.tsx
@@ -29,8 +29,30 @@ export default () => {
         () => getServers({ page, type: showOnlyAdmin && rootAdmin ? 'admin' : undefined })
     );
 
+    const [sortedServers, setSortedServers] = useState<Server[]>([]);
+
     useEffect(() => {
         if (!servers) return;
+
+        const sorted = servers.items.slice().sort((a, b) => {
+            const aDescription = a.description || '';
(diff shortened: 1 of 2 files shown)


resources/scripts/components/dashboard/DashboardContainer.tsx
```diff
@@ -29,8 +29,30 @@ export default () => {
...
+            const bDescription = b.description || '';
+
+            const aMatch = aDescription.match(/^\[(\d+)\]/);
+            const bMatch = bDescription.match(/^\[(\d+)\]/);
+
+            const aNumber = aMatch ? parseInt(aMatch[1], 10) : Number.MAX_SAFE_INTEGER;
+            const bNumber = bMatch ? parseInt(bMatch[1], 10) : Number.MAX_SAFE_INTEGER;
+
+            if (aNumber === bNumber) {
+                // Se i numeri sono uguali, ordina per nome
```
heytkowskii, line 48: English please

Skjolberg:
has this been forgotten?

lajczi:
Do not worry; Dane will maybe review it within weeks.

voidvalue-dev:
I like the current system. Instead of this, I suggest implementing a system that allows users to place it however they want.

PR #5233, opened 2024-10-08
[ Enhancement ] @class directive

ibrahimBougaoua:
**used @class directive instead if statement.**


camwhi-te:
Considering there's a lot more @if statements in the codebase it's probably worth replacing the lot in one go.

ibrahimBougaoua:
> Considering there's a lot more @if statements in the codebase it's probably worth replacing the lot in one go.

Yes,
If that's ok, I'll complete the rest of the changes.

PR #5242, opened 2024-11-05
fix: added loading of .env file to entrypoint.sh

mlunkeit:
Fix: Load environment variables from /app/.env in entrypoint.sh

When using the Pterodactyl panel with Docker, environment variables such as DB_PORT and DB_HOST were not being loaded in entrypoint.sh. This commit resolves the issue by loading the /app/.env file using xargs, ensuring that all necessary environment variables are available at runtime.

### DIFF
--- a/.github/docker/entrypoint.sh
+++ b/.github/docker/entrypoint.sh
@@ -29,6 +29,12 @@ else
   ln -s /app/var/.env /app/
 fi
 
+## load .env file
+if [ -f /app/.env ]; then
+  echo "Loading environment variables from /app/.env"
+  export $(grep -v '^#' /app/.env | xargs)
+fi
+
 echo "Checking if https is required."
 if [ -f /etc/nginx/http.d/panel.conf ]; then
   echo "Using nginx config already in place."


WeeskyBDW:
Correct me if i'm wrong but normally docker load .env file that is located at the same location of the compose (docker-compose.yml) file, so it shouldn't be necessary to add that. I already used pterodactyl with docker and i never had to make that, also because thoses variables you mentionned are specified in the docker-compose (see [the example compose file located at project's root](https://github.com/pterodactyl/panel/blob/de3b1efb6991c9039d2e86bb395072ab5c30d90c/docker-compose.example.yml#L72) ligne 72 and 73).

If docker dont load the .env i would recommand to check if you have up-to-date versions of both docker and docker compose. Again if i'm wrong feel free to correct me.

mlunkeit:
Thank you for your response,
I created a volume in the docker-compose.yml and linked `panel.env` to `/app/.env`, I think this is a possible reason why the `.env` wasn't recognized by the `entrypoint.sh` file. But it also works with committed code.
I guess I'm going to close my pull request since it was my bad naming the file `panel.env` and not `.env` on my host system.

WeeskyBDW:
If it can help you, you can set `env_file` with the name of your dotfile (see [docker-compose's env_file doc](<https://docs.docker.com/reference/compose-file/services/#env_file>)) and it's simplier than mounting the .env

edit: make the comment more "friendly" with better doc link and a less "rude" beginning

mlunkeit:
Thank you, I didn't know that.

PR #5271, opened 2025-01-06
Add header-based authentication support

### LINKED ISSUE #4026: Authentication using http header

It would be nice to allow authentication via an HTTP-header containing the username. This would allow using proxy authentication via sso providers like [Authelia](https://github.com/authelia/authelia) or [Authentik](https://github.com/goauthentik/authentik) and by this, it would also be possible to authenticate using ldap or oidc.

- New environment variables for:
  - Whether the header authentication should be enabled
  - The header containing the username of the user
  - The header containing the email of the user
- Automatic account creation for new users, this can happen transparently in my option

Http header authentication basically works like this:
- You put a proxy in front of your application
- This proxy does not let anything through to the application until the user authenticates with the proxy
- As soon as the user is authenticated, the proxy passes the traffic to the application including a header containing their username, email and whatever else you configure

I am aware that other requests for ldap (#594), saml (#2635) and oidc (#3990) have been dismissed for being too time-consuming to implement. 
Using an HTTP-header, however, would be much simpler to do.

Many other services do this as well, for example [Paperless](https://paperless-ngx.readthedocs.io/en/latest/configuration.html?highlight=PAPERLESS_ENABLE_HTTP_REMOTE_USER#hosting-security)
and [Firefly III](https://docs.firefly-iii.org/firefly-iii/advanced-installation/authentication/#remote-user)

I think, this would be a good solution for all the users who wanted to have ldap, saml or oidc. And it should not be too complicated to implement :slightly_smiling_face: What do you think?

banddude:
This PR adds support for header-based authentication, allowing users to authenticate via HTTP headers. This is particularly useful for proxy authentication via SSO providers like Authelia or Authentik.

Features:
- New HeaderAuthentication middleware
- Configurable header names for username and email
- Optional automatic user creation
- Comprehensive test suite
- SQLite and MySQL compatibility

The feature can be enabled via environment variables:
```
AUTH_HEADER_ENABLED=true
AUTH_HEADER_AUTO_CREATE=true
AUTH_HEADER_USERNAME=X-Auth-Username
AUTH_HEADER_EMAIL=X-Auth-Email
```

## Reverse Proxy Configuration
When using this feature with a reverse proxy, it is important to exclude the /api route from header authentication to prevent issues with API requests. Here is an example Nginx configuration:

```nginx
location /api {
    proxy_pass http://panel:80;
    # Do not add auth headers for API routes
}

location / {
    proxy_pass http://panel:80;
    # Add auth headers here
    proxy_set_header X-Auth-Username $user;
    proxy_set_header X-Auth-Email $email;
}
```

This implementation provides a simple way to integrate with existing SSO solutions without requiring complex LDAP, SAML, or OIDC implementations. The proxy handles the authentication, and the panel trusts the headers it receives.

All tests are passing, and the implementation is compatible with both MySQL and SQLite databases.

Fixes #4026

### DIFF
--- a/config/auth.php
+++ b/config/auth.php
@@ -55,9 +55,8 @@ return [
             'driver' => 'session',
             'provider' => 'users',
         ],
-
-        'api' => [
-            'driver' => 'token',
+        'sanctum' => [
+            'driver' => 'sanctum',
             'provider' => 'users',
         ],
     ],
@@ -126,4 +125,11 @@ return [
     */
 
     'password_timeout' => env('AUTH_PASSWORD_TIMEOUT', 10800),
+
+    'header' => [
+        'enabled' => env('AUTH_HEADER_ENABLED', false),
+        'username_header' => env('AUTH_HEADER_USERNAME', 'X-Auth-Username'),
+        'email_header' => env('AUTH_HEADER_EMAIL', 'X-Auth-Email'),
+        'auto_create' => env('AUTH_HEADER_AUTO_CREATE', false),
+    ],
 ];
--- a/config/auth.php
+++ b/config/auth.php
@@ -55,9 +55,11 @@ return [
             'driver' => 'session',
             'provider' => 'users',
         ],
-        'sanctum' => [
-            'driver' => 'sanctum',
+
+        'api' => [
+            'driver' => 'token',
             'provider' => 'users',
+            'hash' => false,
         ],
     ],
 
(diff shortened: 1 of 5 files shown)


Zaryu:
You have to exclude the api route from pterodactyl on your reverse proxy, to bypass the api route, otherwise youll get an error with: "Request header to big"

banddude:
Thanks for the feedback! You are right - I have updated the PR description to include recommended reverse proxy configuration. Users should exclude the `/api` route from header authentication to prevent issues with API requests.

For example, with Nginx:

```nginx
location /api {
    proxy_pass http://panel:80;
    # Do not add auth headers for API routes
}

location / {
    proxy_pass http://panel:80;
    # Add auth headers here
    proxy_set_header X-Auth-Username $user;
    proxy_set_header X-Auth-Email $email;
}
```

This ensures API functionality remains unaffected while still allowing header authentication for the web interface.

config/auth.php
```diff
@@ -59,6 +59,7 @@
         'api' => [
             'driver' => 'token',
             'provider' => 'users',
+            'hash' => false,
```
Wouter0100, line 62: Curious to the reasoning on setting `hash` to `false` here? I don't think that is part of this feature, and this seems to be the default value according to older Laravel docs:

https://laravel.com/docs/5.8/api-authentication#hashing-tokens 

Maybe good to keep removed?

PR #5283, opened 2025-02-04
Add basic logic to block actions to subusers after removing them from server

MinerPL:
This pull request is frontend implementation for https://github.com/pterodactyl/wings/pull/213


Jelcoo:
I would recommend also implementing this logic when the sftp access permission is revoked.

MinerPL:
> I would recommend also implementing this logic when the sftp access permission is revoked.

Done

Issue #5237, opened 2024-10-27
Schedules Running multiple times during timezone change dates


haraldmbs:
### Current Behavior

Schedules will randomly run several times more than expected (Have had it run 10x what it is supposed to do, and what it normally does)

### Expected Behavior

Schedule runs once or the same amount it normally would.

### Steps to Reproduce

Create a schedule. Have a timezone change happen (In this case CEST timezone changes 1 hour back)

### Panel Version

1.11.7

### Wings Version

1.11.13

### Games and/or Eggs Affected

Assuming all, tested: Forge, Paper

### Docker Image

ghcr.io/pterodactyl/yolks:java_21 and ghcr.io/pterodactyl/yolks:java_18

### Error Logs

```bash
No relevant information, I believe this to be a problem either with the scheduling system itself, or cron, if it only triggered twice (1 hour apart) I wouldn't think much of it, but it spam did the schedule for around an hour span. and then again about 30 min ago it spammed a different schedule.
```

### Is there an existing issue for this?

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

sprocketaudio:
Same problem here for me again (happened last year too and I ended up reinstalling the OS, and Panel/Wings!!)

 Panel  1.11.7
 Wings 1.11.13
 
 .env file set to Europe/Berlin
 config.yml file set to Europe/Berlin
 Linux OS timezone set to Europe/Berlin
 MYSQL time set to 'SYSTEM'

Subthy:
same issue happening here.
Got America/Chicago time zone

ninco90:
The same thing happens here. The task keeps repeating itself every minute 1 hour before the scheduled time and until it reaches the scheduled time it does not stop repeating itself.

My colleague has already restarted the wings, the panel,... verified the time settings. We don't know what else to try.

sprocketaudio:
The following was posted in Discord and has worked for me finally!


Run in SSH:
**sudo nano /var/www/pterodactyl/.env**
check that this is the correct timezone in APP_TIMEZONE
**cd /var/www/pterodactyl && php artisan config:clear  (This is the important bit I think!!!)
sudo systemctl restart ngin**x
Restart the whole panel server with reboot

ninco90:
[...]

PR #5295, opened 2025-02-27
fix: timezone is lost during the creation of CarbonImmutable in Schedule

execsuroot:
Problem: The schedule was running multiple times instead of once because the next run date was being set to UTC, ignoring the specified timezone for Pterodactyl.

Solution: Changed the code to use CarbonImmutable::instance instead of CarbonImmutable::createFromTimestamp. This preserves the timezone information from the CronExpression's getNextRunDate method, ensuring the schedule runs at the correct local time.


execsuroot:
Fixes the #5237 issue.

execsuroot:
This has been open for a long time only for a duplicated to get merged instead, shows how dead this project is, so closing.

PR #5309, opened 2025-03-16
Add language portuguese

opastorello:


matthewpi:
The old translation system is slated to be removed.  We are not accepting new translations at this time.

Issue #5051, opened 2024-03-30
500 Error on Node settings edit


F0x06:
### Current Behavior

500 SERVER ERROR

### Expected Behavior

Working normaly.

### Steps to Reproduce

1. Start provided docker-compose.example.yml file
2. Create admin account with `docker compose exec panel php artisan p:user:make`
3. Log-in
4. Create Location:
     - Short Code: my-location
6. Create Node:
     - Name: test
     - Total Memory: 2048
     - Memory Over-Allocation: 0
     - Total Disk Space: 2048
     - Disk Over-Allocation: 0
     - FQDN: localhost
7. Go to the "Settings" tab of your node and try to change anything, then click "Save Changes"
8. "500 SERVER ERROR"

### Panel Version

1.11.5

### Wings Version

n/a

### Error Logs

```bash
Log files are empty.
```

### Is there an existing issue for this?

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

Boy132:
This is not a bug, please visit the Discord for configuration issues.

F0x06:
> This is not a bug, please visit the Discord for configuration issues.

Thank you for the response. I understand that this might be stemming from a configuration issue. In my experience, unhandled 500 errors often point towards areas where error handling could potentially be enhanced.

Could you provide more detailed information on common causes for this error, or any specific configurations that are often overlooked?. 

I encountered this issue when reinstalling my whole setup I had for years. Everything works except node edition, as a workaround I had to change the settings on the database. 

Thanks

F0x06:
Apparently this was caused by a permission error on the log files folder, by default it's owned by `root` when using local path mounts, for example `- "/srv/pterodactyl/logs:/app/storage/logs"`, the issue is not present when using docker volumes. fixed by using this command `docker compose exec panel chown -R nginx: /app/storage/logs/`, found on discord https://discord.com/channels/122900397965705216/493443683685892096/1150639872198070282

Basically if anything gone wrong on the node settings that involves a log message, the error is triggered.

A possible fix could be to ensure correct ownership of `/app/storage/logs/` on the entrypoint script of the container ?

TBG-FR:
[...]

Issue #4330, opened 2022-08-20
Strange 500 Server Error


StanislasSiweck:
### Current Behavior

Hello,
I have a problem of 500 Server Error when I want to delete a node or a server.
I don't know where it comes from, I have no log and no information about why and how it doesn't work.
All in docker.
I have a traefik, I think it could be that.
URL fail :
- /admin/nodes/view/{nodeID}/settings
- /admin/servers/view/{serversID}/delete

I have no more information, I put you my docker-compose in pass ^^

```yml
version: '3.8'
services:
  traefik:
    image: traefik:latest
    container_name: traefik
    restart: unless-stopped
    security_opt:
      - no-new-privileges:true
    networks:
      - proxy
    ports:
      - 80:80
      - 443:443
    environment:
      - "OVH_ENDPOINT=ovh-eu"
      - "OVH_APPLICATION_KEY="
      - "OVH_APPLICATION_SECRET="
      - "OVH_CONSUMER_KEY="
      # - CF_API_KEY=YOU_API_KEY
      # be sure to use the correct one depending on if you are using a token or key
    volumes:
      - /etc/localtime:/etc/localtime:ro
      - /var/run/docker.sock:/var/run/docker.sock:ro
      - /home/stan/traefik/data/traefik.yml:/traefik.yml:ro
      - /home/stan/traefik/data/acme.json:/acme.json
      - /home/stan/traefik/data/config.yml:/config.yml:ro
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.traefik.entrypoints=http"
      - "traefik.http.routers.traefik.rule=Host(`traefik.domain.net`)"
      - "traefik.http.middlewares.traefik-auth.basicauth.users="
      - "traefik.http.middlewares.traefik-https-redirect.redirectscheme.scheme=https"
      - "traefik.http.middlewares.sslheader.headers.customrequestheaders.X-Forwarded-Proto=https"
      - "traefik.http.routers.traefik.middlewares=traefik-https-redirect"
      - "traefik.http.routers.traefik-secure.entrypoints=https"
      - "traefik.http.routers.traefik-secure.rule=Host(`traefik.domain.net`)"
      - "traefik.http.routers.traefik-secure.middlewares=traefik-auth"
      - "traefik.http.routers.traefik-secure.tls=true"
      - "traefik.http.routers.traefik-secure.tls.certresolver=ovh"
      - "traefik.http.routers.traefik-secure.tls.domains[0].main=domain.net"
[...]

PR #5324, opened 2025-03-31
Fix log permissions, HASHIDS env var generation, and the two errors that have been showing up since the v1.12.0 update

Loki-101:
Tested on branch v1.11.10, v1.12.0, and v1.12.1


Loki-101:
This PR prevents issues like https://github.com/pterodactyl/panel/issues/5051 and https://github.com/pterodactyl/panel/issues/4330

Loki-101:
This PR has been renamed to reflect all current known issues with the image that are being addressed.
Details on the two new issues this PR was just patched to include fixes for:

The v1.12.0 update comes with two breaking changes to the Docker image.

First error:
Creating migration table ...................................... 16.96ms DONE

 INFO  Loading stored database schemas.

database/schema/mysql-schema.sql .............................. 15.18ms FAIL

 Process.php line 275:

The command "mysql  --user="${:LARAVEL_LOAD_USER}" --password="${:LARAVEL_L
OAD_PASSWORD}" --host="${:LARAVEL_LOAD_HOST}" --port="${:LARAVEL_LOAD_PORT}
" --database="${:LARAVEL_LOAD_DATABASE}" < "${:LARAVEL_LOAD_PATH}"" failed.

Exit Code: 127(Command not found)

Working directory: /app

Output:
================


Error Output:
================
sh: mysql: not found

Second error:

Once the mysql client is added to the image, a second error shows up if encryption is not enabled, as is the case for many single-node deployments where the database is an internal-only service:

TLS/SSL error: SSL is required, but the server does not support it

Loki-101:
Solves https://github.com/pterodactyl/panel/issues/5521

Loki-101:
~~The error "TLS/SSL error: SSL is required, but the server does not support it" no longer seems to show up in v1.12.1; the part of the PR that addressed that has been removed.~~

It's still there. I've added the patches back.

Loki-101:
The SSL/TLS error appears to go away when MariaDB's version is set to 11; I'll remove that part of the patch again and just update the example compose instead.

PR #5329, opened 2025-04-22
Fix typo in config key in p:info command

EpicPlayerA10:


KingIronMan2011:
I dont think that is a typo since thats the correct plural of proxy

EpicPlayerA10:
> I dont think that is a typo since thats the correct plural of proxy

I mean the config key is named `trustedproxy`, not `trustedproxies` https://github.com/pterodactyl/panel/blob/1.0-develop/config/trustedproxy.php

KingIronMan2011:
> > I dont think that is a typo since thats the correct plural of proxy
> 
> I mean the config key is named `trustedproxy`, not `trustedproxies` https://github.com/pterodactyl/panel/blob/1.0-develop/config/trustedproxy.php

ok then its probably a typo

Issue #5343, opened 2025-05-12
[GDPR] Don't use googleapi for loading fonts.


prplwtf:
This issue was reported by @sirzento to [BlueprintFramework/framework#121](https://github.com/BlueprintFramework/framework/issues/121).

> I don't know where this file is being used exactly but it loads fonts via googleapi and thats something we want to prevent inside the EU because of GDPR reasons. ([Random site](https://cookie-script.com/blog/google-fonts-and-gdpr) that explains a bit about google fonts and GDPR)
>
> File: https://github.com/BlueprintFramework/framework/blob/ffeb5d092a52fc981085c750f2f13c239296d49e/resources/views/templates/wrapper.blade.php
> 
> Code:
> 
> ```php
> <style>
>     @import url('//fonts.googleapis.com/css?family=Rubik:300,400,500&display=swap');
>     @import url('//fonts.googleapis.com/css?family=IBM+Plex+Mono|IBM+Plex+Sans:500&display=swap');
> </style>
> ```

Issue refers to the following code in the [pterodactyl/panel](https://github.com/pterodactyl/panel) repository and applies to the client view wrapper:
https://github.com/pterodactyl/panel/blob/ccea09c9d94cba49ae52f9c7348a3f4316ec8068/resources/views/templates/wrapper.blade.php#L34-L37

We'd like to first give our upstream (Pterodactyl) a chance to fix this issue, otherwise we'll go ahead and fix it on our side.

PR #5347, opened 2025-05-23
Changes for GDPR compliance

YeahJenni:
Mentioning @prplwtf 's issue with GDPR compliance, #5343. Fonts have been stored locally in public/assets/fonts.

This does need testing, I have not been able to yet, but I believe this *should* work just fine.


YeahJenni:
I've done it wrong, help would be nice xD

YeahJenni:
Closing since I am NOT the right person to be attempting this lmfao

Lovinoes:
why not just switch from google fonts to bunny fonts instead?...

prplwtf:
> why not just switch from google fonts to bunny fonts instead?...

You'd still end up serving fonts through a third party provider.

Issue #5380, opened 2025-07-06
Scheduled Tasks Run Every Minute Due to Incorrect Timezone in `getNextRunDate`


Fallen-Breath:
## Description

After performing maintenance on the operating system hosting Pterodactyl, I encountered an issue where all scheduled tasks are executing every minute, regardless of their configured cron expressions

I searched for related issues in this repository and found #3351 and #4536, but neither provided a viable solution. Therefore, I decided to investigate the issue myself

## Environment

- **System Timezone**: UTC+8, Asia/Shanghai
  ```
  # timedatectl
                 Local time: Mon 2025-07-07 02:27:34 CST
             Universal time: Sun 2025-07-06 18:27:34 UTC
                   RTC time: Sun 2025-07-06 18:27:34
                  Time zone: Asia/Shanghai (CST, +0800)
  System clock synchronized: yes
                NTP service: active
            RTC in local TZ: no
  ```

- **Panel Version**: 1.11.11.0, installed directly on the system (note: I'm using a fork from https://github.com/pterodactyl-china/panel, but my analysis indicates this issue is unrelated to the fork)
- **Wings Version**: 1.11.14.0 (though this issue is unrelated to Wings)
- **Timezone Configuration**: All relevant configurations in the system and Pterodactyl panel explicitly set the timezone to `Asia/Shanghai`, including `.env` with `APP_TIMEZONE=Asia/Shanghai`

## Analysis

I first checked the database and found anomalies in the `next_run_at` field of scheduled tasks:

```sql
MariaDB [panel]> select last_run_at, next_run_at from schedules;
+---------------------+---------------------+
| last_run_at         | next_run_at         |
+---------------------+---------------------+
| 2025-07-07 02:33:02 | 2025-07-06 18:40:00 |   // crontab expr: 0,20,40 * * * *
| 2025-07-07 02:33:03 | 2025-07-06 18:50:00 |   // crontab expr: 10,30,50 * * * *
+---------------------+---------------------+
2 rows in set (0.000 sec)
```

This query was executed at 2025-07-07 02:33 UTC+8. Notably, the `next_run_at` values are earlier than `last_run_at`. Additionally, `next_run_at` is approximately 8 hours behind the current time, suggesting that `next_run_at` is being set using the UTC+0 timezone instead of the local timezone (UTC+8, Asia/Shanghai)

Then I looked into the code. Note that I am not familiar with PHP, so please point out any oversights in my analysis

[...]

PR #5381, opened 2025-07-06
Fix timezone issue in `Schedule::getNextRunDate`

Fallen-Breath:
See #5380


RaphaelIT7:
Isn't this the same solution as in #5295 but just formatted differently?

Fallen-Breath:
it seems to be

rowansc1:
Thanks for the PR! Seems to have come up after daylight saving time. Shall get @matthewpi to have a look :)

ramonmateofrancisco-sys:
Macro

ramonmateofrancisco-sys:
Hollé macro pliz

Issue #5078, opened 2024-04-24
Folders not being prioritized in file manager, with large amount of files after wings 1.11.x


LukasW1337:
### Current Behavior

After upgrading to 1.11.x wings, folders are no longer prioritized in file manager when it limits to 250 files. 
Before the folders in this directory would be on top, and then the files below are the ones that gets cut away.
![image](https://github.com/pterodactyl/panel/assets/33436545/7967d386-3260-481b-a676-011e230824e3)

### Expected Behavior

When we go over 250 files, make sure that folders are shown first, then limit the files after that to 250. This makes sure that you can still access configuration folders etc. when games has a lot of data files in certain folders, but you need to navigate through them into sub folders.

### Steps to Reproduce

Install a game that has a lot of files in one directory, then try to access folders in the same directory.

CS2 has a lot of .vpk files in root folder. But same directory has CFG, Addons etc. that you still need to access. You can only do this through SFTP now, and not the panel file-manager.

### Panel Version

1.11.5

### Wings Version

1.11.11

### Is there an existing issue for this?

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

PR #5387, opened 2025-07-21
Sort directories before files in file manager listing

BeamHost:
Ensures all folders are listed before files when displaying directory contents. This prevents folders from being hidden when the file display cap (250) is reached, allowing users to always navigate directories regardless of total file count.


danny6167:
Wings already sorts the file listing.
Please see https://github.com/pterodactyl/panel/issues/5078 and https://github.com/pterodactyl/wings/pull/188

PR #5424, opened 2025-11-04
openjdk deprecated docker images, change to eclipse-temurin

alanescarcha:
OpenJDK has deprecated the Docker images `openjdk:<any>-jdk-slim` and any other jdk-slim version, so if you install or reinstall any server with eggs that use that Docker image, the installation will fail because those OpenJDK Docker images cannot be obtained. Switch to eclipse-temurin images

Wings error:
```
Stacktrace:
Error response from daemon: manifest for openjdk:17-jdk-slim not found: manifest unknown: manifest unknown
failed to pull updated installation container image for server
```

Also check https://hub.docker.com/_/openjdk#deprecation-notice (DEPRECATION NOTICE)


DaneEveritt (approved):

DaneEveritt:
@alanescarcha can you rebase this against the `1.0-develop` branch so that it runs with the updated required check suite, and then I can get it merged.

PR #5433, opened 2025-11-25
Update minimum Node.js version to 22, updates dependencies

DaneEveritt:
See https://github.com/pterodactyl/panel/pull/5433#issuecomment-3650039400


package.json
```diff
@@ -85,58 +87,55 @@
...
     "scripts": {
         "clean": "cd public/assets && find . \\( -name \"*.js\" -o -name \"*.map\" \\) -type f -delete",
         "test": "jest",
+        "tsc": "tsc --noEmit",
         "lint": "eslint ./resources/scripts/**/*.{ts,tsx} --ext .ts,.tsx",
         "watch": "cross-env NODE_ENV=development ./node_modules/.bin/webpack --watch --progress",
         "build": "cross-env NODE_ENV=development ./node_modules/.bin/webpack --progress",
-        "build:production": "yarn run clean && cross-env NODE_ENV=production ./node_modules/.bin/webpack --mode production",
-        "serve": "yarn run clean && cross-env WEBPACK_PUBLIC_PATH=/webpack@hmr/ NODE_ENV=development webpack-dev-server --host 0.0.0.0 --port 8080 --public https://pterodactyl.test --hot"
+        "build:production": "yarn run clean && cross-env NODE_ENV=production NODE_OPTIONS=--openssl-legacy-provider ./node_modules/.bin/webpack --mode production",
```
PadowYT2: As webpack was updated, this change is no longer needed

```
[nix-shell:~/repos/panel]# node -v
v22.20.0

[nix-shell:~/repos/panel]# yarn build:production
yarn run v1.22.22
$ yarn run clean && cross-env NODE_ENV=production ./node_modules/.bin/webpack --mode production
...
Done in 13.62s.
```

.github/workflows/build.yaml
```diff
@@ -32,6 +32,6 @@ jobs:
 
       - name: Install dependencies
         run: yarn install --frozen-lockfile
-
-      - name: Build
-        run: yarn build:production
+      - run: yarn tsc
+      - run: yarn lint
+      - run: yarn build:production
```
rmartinoscar, line 37:
```suggestion
      - name: Install & Build dependencies
        run: |
          yarn install --frozen-lockfile
          yarn tsc
          yarn lint
          yarn build:production
```

.github/workflows/release.yaml
```diff
@@ -18,12 +18,14 @@ jobs:
...
+          node-version: 22
           cache: yarn
 
       - name: Install dependencies
         run: yarn install --frozen-lockfile
 
+      - run: yarn tsc
+      - run: yarn lint
       - name: Build
         run: yarn build:production
```
rmartinoscar, line 30:
```suggestion
      - name: Install & Build dependencies
        run: |
          yarn install --frozen-lockfile
          yarn tsc
          yarn lint
          yarn build:production
```

package.json
```diff
@@ -30,6 +30,7 @@
         "i18next": "^21.8.9",
         "i18next-http-backend": "^1.4.1",
         "i18next-multiload-backend-adapter": "^1.0.0",
+        "pathe": "^2.0.3",
```
rmartinoscar, line 33: Why use an external library when you already have `node:path` and you are already using it ?
https://github.com/pterodactyl/panel/blob/6d0818dad118b4b5f20b1e3ce65eaad707fa9120/webpack.config.js#L1
https://github.com/pterodactyl/panel/blob/6d0818dad118b4b5f20b1e3ce65eaad707fa9120/webpack.config.js#L18

PadowYT2: A package is required here as the `node:` protocol isn't accessible on the frontend. You would need to polyfill it which needs an external library and in result the asset will be much bigger than just using `pathe`

DaneEveritt: Yeah - Webpack 5 stopped automatically polyfilling these.

https://webpack.js.org/configuration/resolve/#resolvefallback

.github/workflows/release.yaml
DaneEveritt: I prefer these as separate steps because it is quicker to see what specific thing failed in the UI rather than having to scroll and find the error output. 👍🏻

package.json
DaneEveritt:
```suggestion
        "build:production": "yarn run clean && cross-env NODE_ENV=production ./node_modules/.bin/webpack --mode production",
```

laraproto:
The uuid package seems to only be used in the client code once for generating a v4 uuid, every major browser now supports crypto.randomUUID, the uuid package isn't particularly required anymore

DaneEveritt:
It has become abundantly clear that my time and effort is better spent making updates and maintaining the software on an independent fork. It is disappointing to see such a large and popular piece of software languish without a modicum of effort put into even basic maintenance for long term stability.

My limited time is better spent working on a fork where I can make updates and not see them languish for weeks at a time.

I wish this project well.

:v:

finnie2006:
> It has become abundantly clear that my time and effort is better spent making updates and maintaining the software on an independent fork. It is disappointing to see such a large and popular piece of software languish without a modicum of effort put into even basic maintenance for long term stability.
> 
> My limited time is better spent working on a fork where I can make updates and not see them languish for weeks at a time.
> 
> I wish this project well.
> 
> ✌️

I completely agree with you on this, Dane. There is no commitment whatsoever from the current maintainers. It's a shame that they don't want to commit to your work. Plus 1 star for your fork.

PR #5461, opened 2025-12-25
Fix admin server manage boxes misalignment when suspending a server

voidvalue-dev:
The PR is a bit basic, but the interface problem will be fixed. And it doesn't have any problems with other screens.

old:
image
image

new:
image


DaneEveritt (approved):

voidvalue-dev:
Thx :3

PR #5469, opened 2025-12-26
Enable Image Preview in Pterodactyl Files via Clickable Modal

voidvalue-dev:
Currently, images in the Pterodactyl file manager cannot be viewed directly.
This pull request adds support for previewing image files by opening them in a modal when clicked.

The new behavior improves usability by allowing users to quickly view image contents without downloading them or leaving the file manager. The implementation is lightweight and integrates seamlessly with the existing UI.

Preview:
image
image
(Idea taken from an add-on from builtbybit)

### DIFF
--- a/resources/scripts/components/server/files/FileObjectRow.tsx
+++ b/resources/scripts/components/server/files/FileObjectRow.tsx
@@ -8,50 +8,36 @@ import FileDropdownMenu from '@/components/server/files/FileDropdownMenu';
 import { ServerContext } from '@/state/server';
 import { NavLink, useRouteMatch } from 'react-router-dom';
 import tw from 'twin.macro';
-import isEqual from "react-fast-compare";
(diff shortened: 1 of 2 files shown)


lajczi (commented):
Thanks for your contribution!

Can you please format code with Prettier?
I just recommend installing Prettier extension in your IDE (JetBrains/VSCode/etc)

Further review will be by @DaneEveritt :D

lajczi:
Oh and also please fix lint :sob:

voidvalue-dev:
> Thanks for your contribution!
> 
> Can you please format code with Prettier? I just recommend installing Prettier extension in your IDE (JetBrains/VSCode/etc)
> 
> Further review will be by @DaneEveritt :D

I format it with prettier and it gives me 50 more things to change 😭
But as such it would already be, only strange things come out now.
If you would like to suggest any changes, I would be delighted to accept them.

PR #5473, opened 2025-12-26
Change server link button to open in new tab

voidvalue-dev:
This is better, because now you have a wrench and it's not very clear what it does if you don't click on it.

now:
image
with the updated:
image

### DIFF
--- a/resources/views/admin/servers/index.blade.php
+++ b/resources/views/admin/servers/index.blade.php
@@ -61,7 +61,7 @@
                                     @endif
                                 </td>
                                 <td class="text-center">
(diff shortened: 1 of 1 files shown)


AndyIsHereBoi:
i dont know if a pull request was really needed but its here now, i do agree that the wrench looked like a configure button when just clicking the server name does that instead. the open in new tab icon describes the action better as well i feel

voidvalue-dev:
thx

iiAhmedYT:
I SUPPORT THIS. You have no idea how many times this button fooled me.

PR #5474, opened 2025-12-26
Add copy configuration button in admin node configurations

voidvalue-dev:
It is useful
image
image


KingIronMan2011:
In my opinion you should make it just a little copy symbol in the corner of the config text and not a huge button under the text where the config needs to go

voidvalue-dev:
> In my opinion you should make it just a little copy symbol in the corner of the config text and not a huge button under the text where the config needs to go

I had thought about it, and I did something similar to what you suggested, but in my own way. I have updated the photos.

PR #5476, opened 2025-12-27
Scope the remote node token to limit the servers it can manage for backups/transfers

DaneEveritt:
Improves the security posture of things more by limiting the servers that a node can even communicate about with the Panel.

### DIFF
--- a/app/Http/Controllers/Api/Remote/Servers/ServerDetailsController.php
+++ b/app/Http/Controllers/Api/Remote/Servers/ServerDetailsController.php
@@ -40,7 +40,16 @@ class ServerDetailsController extends Controller
         Assert::isInstanceOf($node = $request->attributes->get('node'), Node::class);
 
         $server = $this->repository->getByUuid($uuid);
-        if (! $server->node->is($node)) {
+        $transfer = $server->transfer;
+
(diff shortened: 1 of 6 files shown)


QuintenQVD0:
Have you checked that transfers still work?

DaneEveritt:
@QuintenQVD0 are you able to test that for me, I don't have a good setup for that. So far as I can tell they should be fine since the logic checks that either the source/dest node making the request are associated with the transfer.

app/Http/Controllers/Api/Remote/Servers/ServerDetailsController.php
```diff
@@ -34,7 +37,12 @@ public function __construct(
      */
     public function __invoke(Request $request, string $uuid): JsonResponse
     {
+        Assert::isInstanceOf($node = $request->attributes->get('node'), Node::class);
+
         $server = $this->repository->getByUuid($uuid);
+        if (! $server->node->is($node)) {
+            throw new HttpForbiddenException('Requesting node does not have permission to access this server.');
```
Boy132, line 53: This will break transfers: the destination node will call this endpoint and then error because technically the server still belongs to the source node.

For reference: https://github.com/pelican-dev/panel/issues/1680

Issue #2484, opened 2020-10-12
Sending Symbols and message in console produce a symbols over the message


Doc94:
**Background (please complete the following information):**
* Panel or Daemon: Panel
* Version of Panel/Daemon: 1.0.0
* Server's OS: Ubuntu 18 LTS
* Your Computer's OS & Browser: Opera GX

**Describe the bug**
A especial symbols have incorrect padding

**To Reproduce**
1. Go to a server panel (minecraft server)
2. Show Console
3. InGame send a message with symbol like "\u22d9 Hello People" to show in console
4. Look Console

**Expected behavior**
Show the message with the symbol, but actually show the message and the symbol over the message, with not spaces.
Note: in old panel all the symbols like this show correctly.

**Image**
![image](https://user-images.githubusercontent.com/3602279/95695140-f8023580-0c0b-11eb-91e4-9f09fdb17c64.png)

matthewpi:
Are you still able to reproduce this issue?

Doc94:
> Are you still able to reproduce this issue?

Yeah still happen, now testing in last version of panel 1.4.2 and still happen in console, this can be reproduced by send in chat a text like `۞ test` (or using plugins) and see console.
You can get the simbols using webs like https://minecraft.tools/en/tellraw.php?tellraw=۞%20test&selector=%40a (for tellraw or motd)
This not happen with all special characters..
![image](https://user-images.githubusercontent.com/3602279/125860192-f2583aa0-b34e-4fe5-ba2d-49349ad59c06.png)
![image](https://user-images.githubusercontent.com/3602279/125860613-011d7f97-d741-4c6f-abfb-07f7c78d1f10.png)
raw console (selected and copy from console)
`[21:27:34 INFO]: ۞ MrDoc94 fue a dormir.`
`[21:28:25 INFO]: MrDoc94 issued server command: /tellraw @a {"text":"\u06de test"}`
`[21:28:42 INFO]: <[ADMIN] MrDoc94> ۞ test`
`[21:30:00 INFO]: <[ADMIN] MrDoc94> ۞ test`
`[21:31:21 INFO]: <[ADMIN] MrDoc94> ► conectado`
`[21:31:42 INFO]: <[ADMIN] MrDoc94> ᚏ bue`
`[21:32:38 INFO]: <[ADMIN] MrDoc94> ⁂ ss`

PR #5523, opened 2026-01-11
Fix Unicode character width rendering using Unicode11Addon in xterm.js

voidvalue-dev:
TO FIX:
https://github.com/pterodactyl/panel/issues/2484

image


voidvalue-dev:
I'm going to do it again. I had several commits that are unrelated to this.

Issue #5533, opened 2026-01-16
System Information Version checker


lighscent:
### Current Behavior

System information show 1.11.11 and says Panel is up to date
<img width="518" height="152" alt="Image" src="https://github.com/user-attachments/assets/f08d38a0-925d-4577-b816-3d2e1a6b02e2" />

### Expected Behavior

System information should show current version is 1.11.11 and latest is 1.12.1 and panel need to be updated
(like all version before)

### Steps to Reproduce

idk, install panel in 1.11.11 and u will see

### Panel Version

1.11.11

### Wings Version

1.11.13 (daemon)

### Games and/or Eggs Affected

none

### Docker Image

none

### Error Logs

```bash
https://ptero.co/mksLk9
```

### Is there an existing issue for this?

- [x] I have searched the existing issues before opening this issue. I understand that maintainers may close this issue without communication if I have not provided sufficient information.

voidvalue-dev:
same

AndyIsHereBoi:
it was doing this for me too on 4 panels

AndyIsHereBoi:
<img width="444" height="195" alt="Image" src="https://github.com/user-attachments/assets/b0c0c92f-2700-419d-81b5-dfcdeaa916f9" />
currently still does on one of them

Liam-Broome:
Created a [PR](https://github.com/pterodactyl/panel/pull/5537).

DaneEveritt:
fixed in https://github.com/pterodactyl/panel/commit/5f1bfd109ddb6cb35b75e88a8ae6d7ce9f5b0e25

PR #5537, opened 2026-01-18
[Issue][5533] - Version comparison.

Liam-Broome:
Issue - [5533](https://github.com/pterodactyl/panel/issues/5533).

`version_compare` does not like leading 'v'.
Solution was to trim the leading 'v' in the comparison.

image

### DIFF
--- a/app/Services/Helpers/SoftwareVersionService.php
+++ b/app/Services/Helpers/SoftwareVersionService.php
@@ -65,7 +65,11 @@ class SoftwareVersionService
             return true;
         }
 
-        return version_compare(config('app.version'), $this->getPanel()) >= 0;
+        return version_compare(
+            ltrim(config('app.version'), 'vV'),
+            ltrim($this->getPanel(), 'vV'),
+            '>='
+        );
     }
 
     /**


AndyIsHereBoi:
I'm pretty sure they were told they can remove it, I bet it was and wasn't yet deployed

https://discord.com/channels/122900397965705216/122902526751145984/1457888588564791412

Looks like he said "yes" replying but maybe wasn't done yet?

Liam-Broome:
@AndyIsHereBoi 
I could recreate the issue, so the JSON change likely wasn't deployed yet 👀 

Of course if the source of the version string is fixed this PR is redundant unless we wanna keep the additional handling to prevent this from occurring again.

DaneEveritt:
This is actually just due to us shipping `v1.12.0` with the leading `v` which wasn't supposed to be included.

This was resolved in https://github.com/pterodactyl/panel/commit/5f1bfd109ddb6cb35b75e88a8ae6d7ce9f5b0e25 which will fix itself upon the next release of the software. 👍🏻

PR #5548, opened 2026-02-01
Add support for stripe-style identifiers on existing models with UUIDs

DaneEveritt:
This is a partial implementation to begin moving towards stripe-style identifiers for resources in the system. Any models with an existing `uuid` column can easily be updated to return an identifier in the format of `prfx_xyz` where `prfx` is a four character prefix, and `xyz` is the UUID, encoded using base-32.

These are quite easy to use within the API layer because we just need to do one quick transformation to extract the UUID for those models. This PR implements that logic for servers in the `SubstituteClientBindings` logic.

A future PR will need to come through and handle identifiers for models that _don't_ currently use UUIDs for reference that we want to expose to clients. In those cases it is easier to just generate base-32 encoded UUID7s that get stored in the database and indexed. They follow the same base approach, but you don't need to do any transformations in the code (other than stripping the prefix, unless we decide to store the prefix).

There is also now a `PTERODACTYL_USE_SERVER_IDENTIFIERS` environment variable, that when set to true, updates the front-end and API response to use this new identifier in place of the `uuidShort` value.

### DIFF
--- a/app/Models/Egg.php
+++ b/app/Models/Egg.php
@@ -2,7 +2,9 @@
 
 namespace Pterodactyl\Models;
 
+use Pterodactyl\Contracts\Models\Identifiable;
 use Illuminate\Database\Eloquent\Relations\HasMany;
+use Pterodactyl\Models\Traits\HasRealtimeIdentifier;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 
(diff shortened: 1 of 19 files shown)


app/Models/Egg.php
```diff
@@ -47,10 +49,12 @@
  * @property Egg|null $scriptFrom
  * @property Egg|null $configFrom
  */
-class Egg extends Model
+#[Attributes\Identifiable('eegg')]
```
DaneEveritt, line 52: I couldn't think of how to make this four letters. It was `eegg` or `eggg`. 🤷🏻

voidvalue-dev: eggs?

anthonyphysgun: that might just be big brain

DaneEveritt: `oeuf` — sorry, I'm French.

voidvalue-dev: `huev`o I'm spanish

PR #5573, opened 2026-02-15
Fix transfer status permission checks

### LINKED ISSUE #5572: Transfers get stuck processing due to wrong node permission check on successful transfer status update
### Current Behavior

Currently, the panel attempts to check if the successful transfer status update is coming from the new node by checking if the servers node is the new node in the transfer, before the server has been updated to the new node, always failing.

### Expected Behavior

The panel should check if the node making the request is the new node in the transfer.

### Steps to Reproduce

Attempt to transfer a server.

### Panel Version

1.12.1

### Wings Version

1.12.1

N/A

### Docker Image

N/A

### Error Logs

```bash
ERROR: [...] failed to set transfer status on panel error=Error response from Panel: HttpForbiddenException: Requesting node does not have permission to access this server. (HTTP/403) server=... status=true subsystem=transfer
```

MrSoulPenguin:
Before, the new node was checked against the servers current node (the old node) before the server is updated to the new node, causing the check to always fail.

Now the requesting node is checked against the new node.

Resolves [#5572](https://github.com/pterodactyl/panel/issues/5572)

### DIFF
--- a/app/Http/Controllers/Api/Remote/Servers/ServerTransferController.php
+++ b/app/Http/Controllers/Api/Remote/Servers/ServerTransferController.php
@@ -34,7 +34,7 @@ class ServerTransferController extends Controller
      *
      * @throws \Throwable
      */
-    public function failure(string $uuid): JsonResponse
+    public function failure(Request $reqest, string $uuid): JsonResponse
     {
         $server = $this->repository->getByUuid($uuid);
         $transfer = $server->transfer;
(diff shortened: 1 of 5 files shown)


MrSoulPenguin:
Noticed that there was also an issue with the failed transfer status update node checks. Checked against the servers current node instead of the requesting node.

So the check always passed when the servers current node was check against the old node in the transfer, since it wouldn't have been updated due to the failed transfer.

MrSoulPenguin:
Went ahead and added some test cases to cover this in the future, and fixed a typo I didn't notice before.

Space-Banane:
bump, same issues

MrSoulPenguin:
> bump, same issues

If you're saying you've experienced the same issue this pr addresses, then it's probably best to do so on the linked issue.

Doing it on the pr seems like you're stating that you're still experiencing the issue after testing the fix.

Issue #5570, opened 2026-02-14
Runtask broken in new update


minexnode:
### Current Behavior

RunTaskJob dont work

### Expected Behavior

Runtakjob Works

### Steps to Reproduce

Run a scedle/task

### Panel Version

1.12.1

### Wings Version

1.12

### Error Logs

```bash
/var/www/pterodactyl/app/Jobs/Schedule/RunTaskJob.php

From:
$this->dispatch((new self($nextTask, $this->manualRun))->delay($nextTask->time_offset));

To:
self::dispatch($nextTask, $this->manualRun)->delay($nextTask->time_offset);


log:[2026-02-14 22:21:42] production.ERROR: Pterodactyl\Jobs\Schedule\RunTaskJob::__construct(): Argument #1 ($task) must be of type Pterodactyl\Models\Task, Pterodactyl\Jobs\Schedule\RunTaskJob given, called in /var/www/pterodactyl/vendor/laravel/framework/src/Illuminate/Foundation/Bus/Dispatchable.php on line 19 {"userId":1,"exception":"[object] (TypeError(code: 0): Pterodactyl\\Jobs\\Schedule\\RunTaskJob::__construct(): Argument #1 ($task) must be of type Pterodactyl\\Models\\Task, Pterodactyl\\Jobs\\Schedule\\RunTaskJob given, called in /var/www/pterodactyl/vendor/laravel/framework/src/Illuminate/Foundation/Bus/Dispatchable.php on line 19 at /var/www/pterodactyl/app/Jobs/Schedule/RunTaskJob.php:21)
```

### Is there an existing issue for this?

- [x] I have searched the existing issues before opening this issue. I understand that maintainers may close this issue without communication if I have not provided sufficient information.

minexnode:
this is basicly the same as https://github.com/pterodactyl/panel/issues/5258

quick fix: /var/www/pterodactyl/app/Jobs/Schedule/RunTaskJob.php

From:
$this->dispatch((new self($nextTask, $this->manualRun))->delay($nextTask->time_offset));

To:
self::dispatch($nextTask, $this->manualRun)->delay($nextTask->time_offset);

voidvalue-dev:
> this is basicly the same as [#5258](https://github.com/pterodactyl/panel/issues/5258)
> 
> quick fix: /var/www/pterodactyl/app/Jobs/Schedule/RunTaskJob.php
> 
[...]

PR #5586, opened 2026-02-20
Refactor task dispatching to use self::dispatch

minexnode:
Fixing the schedule bug


minexnode:
lets see if i have done this corect now

Dimoka113:
I confirm that this fix works.
It also resolves issue #5570
(After implementing this fix, the schedules started working correctly.)

PR #5610, opened 2026-03-30
Rename toVueObject -> toReactObject on User model

camwhi-te:
Simple change renaming Eloquent helper in accordance with actual UI stack.

Also, who decided to use Vue initially? ew.


robertdrakedennis:
@camwhi-te Can you refactor this to `toJsObject`

PR #5611, opened 2026-03-30
fix(security): block SSRF in file pull URL validation

aymenelouadi:
Validate that the URL provided to the file pull endpoint does not resolve to an internal, loopback, link-local, or otherwise reserved IP address, preventing Server-Side Request Forgery (SSRF) attacks.

Checks performed:
- Hostname is parsed from the URL before any resolution.
- If the host is an IP literal it is validated directly.
- If the host is a domain name, gethostbynamel() resolves all A records so DNS-rebinding tricks are also caught.
- FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE blocks: 10.x, 172.16-31.x, 192.168.x (RFC-1918) 127.x (loopback), 169.254.x (link-local / cloud metadata) and other IANA-reserved ranges.

### DIFF
--- a/app/Http/Requests/Api/Client/Servers/Files/PullFileRequest.php
+++ b/app/Http/Requests/Api/Client/Servers/Files/PullFileRequest.php
@@ -16,11 +16,55 @@ class PullFileRequest extends ClientApiRequest implements ClientPermissionsReque
     public function rules(): array
     {
         return [
-            'url' => 'required|string|url',
+            'url' => ['required', 'string', 'url', function ($attribute, $value, $fail) {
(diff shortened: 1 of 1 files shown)


matthewpi:
This is already protected against inside of Wings.

This code also won't prevent redirects to internal IPs.

This isn't necessary.

aymenelouadi:
You're right that HTTP redirects can bypass DNS-level pre-validation that's a real gap. If Wings already handles this correctly including after following redirects, then a partial check at the Panel level that misses redirect-based bypasses would give false confidence rather than real protection. I'll defer to Wings implementation for this.

Issue #5630, opened 2026-05-04
Compress activity log shows `{{file}}` instead of the archive filename


marioenache:
### Current Behavior

When compressing files, the activity log entry for a single-file compress renders the filename placeholder literally instead of resolving it:

> Compressed `/{{file}}`

### Expected Behavior

Expected:

> Compressed `/file.txt`

### Steps to Reproduce

1. Open a server's file manager
2. Select one or more files and compress them
3. Navigate to the server's Activity Log
4. Observe the compress entry, the filename shows as `{{file}}`

In `resources/lang/en/activity.php`, the `compress_one` string references the placeholder `:file` (singular):

Fix > resources/lang/en/activity.php (70):

- 'compress_one' => 'Compressed :directory:file',
+ 'compress_one' => 'Compressed :directory:files.0',

### Panel Version

1.12.2

### Wings Version

1.12.1

### Error Logs

```bash

```

### Is there an existing issue for this?

- [x] I have searched the existing issues before opening this issue. I understand that maintainers may close this issue without communication if I have not provided sufficient information.

danny6167:
Closed by https://github.com/pterodactyl/panel/commit/4de50090141f549967fa715ef42aaaee2b4fc983

PR #5637, opened 2026-05-27
Fixed compress activity log showing {{file}} instead of the archive filename

REGEX777:
Fix for #5630 

The compress_one string in resources/lang/en/activity.php referenced :file as the placeholder, but the backend passes the filename under a files array. This caused the activity log to render {file}} instead of the actual filename

Changed :file to :files.0 to correctly resolve the first filename from the array.

Solution was identified by the issue reporter (@marioenache).

### DIFF
--- a/resources/lang/en/activity.php
+++ b/resources/lang/en/activity.php
@@ -67,7 +67,7 @@ return [
             'delete' => 'Deleted database :name',
         ],
         'file' => [
-            'compress_one' => 'Compressed :directory:file',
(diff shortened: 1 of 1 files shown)


robertdrakedennis:
hey @REGEX777 can you submit a screenshot and re-target `develop` please?

REGEX777:
Sorry I couldnt get back to you sooner.
Here,

image
Before ^

image
After ^

PR #5652, opened 2026-06-15
Wings Update Alert

itzlinaton:
## Changes:
* Introduced an update alert similar to the panel update alert for Wings in the directory: `/resources/views/admin/nodes/view/index.blade.php`
E4E5B2D4-C51A-4DA2-BAB1-49AD2E26B698

### DIFF
--- a/resources/views/admin/nodes/view/index.blade.php
+++ b/resources/views/admin/nodes/view/index.blade.php
@@ -1,0 +1,176 @@
+@extends('layouts.admin')
+
+@section('title')
+    {{ $node->name }}
+@endsection
+
+@section('content-header')
+    <h1>{{ $node->name }}<small>A quick overview of your node.</small></h1>
+    <ol class="breadcrumb">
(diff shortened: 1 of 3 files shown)


AndyIsHereBoi:
i dont think the update guide link is needed, especially because it makes 3 new lines. the wings is not up to date and its version  can just replace the original message because its saying the same thing. You probably could just say "**Wings is not up-to-date!**" and put that in front of the current text and it would be fine

resources/views/admin/nodes/view/index.blade.php
```diff
@@ -157,12 +164,48 @@ function escapeHtml(str) {
...
+            (function () {
+                const box = $('[data-attr="info-version"]').closest('.box');
+                const status = $('#wings-status');
+
+                if (!box.length) return;
+
+                if (current !== latest) {
+                    box.removeClass('box-primary').addClass('box-danger');
+
+                    status.html(
```
lajczi: sorry but this implementation is unacceptable. Please try to follow our code standards

PR #5665, opened 2026-06-23
Fix: Node auto deploy using key with insufficient permissions

### LINKED ISSUE #5664: Node auto configuration token uses token with insufficient permissions
### Current Behavior

When clicking on "Generate Token" the panel generates a wings auto config URL that includes an application token.
It finds a token by first searching existing tokens with read access to node config, and then generating a fresh application token with node config read permissions if it couldn't find one.

As of https://github.com/pterodactyl/panel/commit/f04d2f52f1d60053dc35b86b1207220673f94f38 this is insufficient.

### Expected Behavior

Token found/generated with required permissions

### Steps to Reproduce

- Generate Token on node config page.
- Attempt to use configure command
- Command errors out due to insufficient access.

### Panel Version

1.14.0

### Wings Version

N/A

_No response_

### Docker Image

_No response_

### Error Logs

```bash

```

danny6167:
You now require write access to the node endpoint to pull the config.
This change updates the token fetching/generation to use a token with R&W permissions

Fixes #5664

### DIFF
--- a/app/Http/Controllers/Admin/NodeAutoDeployController.php
+++ b/app/Http/Controllers/Admin/NodeAutoDeployController.php
@@ -22,8 +22,8 @@ class NodeAutoDeployController extends Controller
     }
 
     /**
-     * Generates a new API key for the logged-in user with only permission to read
-     * nodes, and returns that as the deployment key for a node.
+     * Generates a new API key for the logged-in user with read and write permission
+     * to nodes, and returns that as the deployment key for a node.
      *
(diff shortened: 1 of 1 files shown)


robertdrakedennis:
thanks for tracking this down

i'm gonna close this and fix it in a separate commit. i don't think we want to solve this by giving the generated key full node write perms since that's broader than the configure flow needs.

as it's definitely a regression from a previous security patch 👍

PR #5668, opened 2026-06-30
feat(pocketmine): auto-detect PHP version and fix installation

qsaiwolf:
- Replace hardcoded PHP download URL with dynamic version detection via PMMP API
- Add symlink bin/php7 -> actual PHP folder for future-proof startup command
- Auto-download PocketMine-MP.phar during installation
- Simplify startup command to single line
- Installation now works without any manual intervention


qsaiwolf:
## Problem
The current installation script uses a hardcoded PHP download URL 
which breaks when PMMP requires a new PHP version.

## Solution
- Dynamically detect required PHP version via PMMP API
- Create symlink bin/php7 for future-proof startup command
- Auto-download PocketMine-MP.phar during installation

PR #5670, opened 2026-06-30
Fix backup restore activity event names

ShiroKSH:
This PR fixes activity log event names for backup restores.

Failed restore attempts were logged with an event name that used the wrong separator, while reset-state was checking for an event name that is never created anywhere. As a result, restore-related activity logs could not be matched consistently.

This change only updates activity log event naming. It does not change the restore logic itself.

Checks:

* php -l on both modified files
* git diff --check

### DIFF
--- a/app/Http/Controllers/Api/Remote/Backups/BackupStatusController.php
+++ b/app/Http/Controllers/Api/Remote/Backups/BackupStatusController.php
@@ -102,7 +102,7 @@ class BackupStatusController extends Controller
 
         $model->server->update(['status' => null]);
 
-        Activity::event($request->boolean('successful') ? 'server:backup.restore-complete' : 'server.backup.restore-failed')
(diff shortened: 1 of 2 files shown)


ShiroKSH:
﻿@pterodactyl We're ready for verification when you have a chance.

ShiroKSH:
@DaneEveritt - could you take a look at the backup restore event-name fix in this PR when you have a chance, or route it to the right reviewer? I can make any requested changes or add tests. Thanks!

ShiroKSH:
@DaneEveritt Friendly reminder: we'd appreciate it if you could take a look at this pull request when you have a chance.

ShiroKSH:
Hi maintainers, this PR has been open for over two weeks. When you have a moment, I would appreciate a review. Thank you.

PR #5672, opened 2026-07-02
Improve danger text readability

egg400:
When you try to delete a server on the panel, you'll see warnings that say stuff like "Deleting a server is an irreversible action. All server data...". This is useful, however, the text is extremely hard to read due to the background. A user would need to highlight the text to make it readable.

I've tried changing the text color to other colors, but they didn't provide enough contrast either. I eventually decided to make the background of the text red and make the text white. I've tried to keep the design as close as possible to the Pterodactyl UI.

This change is only 4 lines in CSS. Check attached images for before & after.

Before
image

After
image

### DIFF
--- a/public/themes/pterodactyl/css/pterodactyl.css
+++ b/public/themes/pterodactyl/css/pterodactyl.css
@@ -561,7 +561,10 @@ body {
 }
 
 .text-danger {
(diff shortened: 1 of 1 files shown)


Lovinoes:
I like this

PR #5679, opened 2026-07-11
Add Psalm static analysis (security/taint scanning)

alies-dev:
Adds Psalm ([psalm-plugin-laravel](https://github.com/psalm/psalm-plugin-laravel)) for taint (security) scanning in CI: SQL injection, XSS, SSRF, command injection, path traversal. Larastan already covers types here, so this fills the gap it can't: cross-function dataflow tracking that catches injection bugs PHPStan's architecture structurally can't reach. Findings are baselined so CI is green on merge; new taint issues fail the build and land in the Security tab.

Also bumps `laravel/framework` to `^11.54` (from `^11.31.0`). Composer's advisory audit blocks installing the plugin against older 11.x releases, and three of the flagged CVEs have no fix anywhere on the 11.x line, only 12.60+/13.10+/12.61.1+, so they're ignored in `composer.json` with reasons. Laravel 11 not getting these backports is worth keeping an eye on.

Also fixes a real finding from the taint scan itself: `ApiKey::findToken()` compared a decrypted API token against the caller-supplied token with `===`. That model is wired as Sanctum's personal-access-token model, so the comparison runs on every authenticated API request, and `===` on strings short-circuits at the first mismatched byte, so comparison time leaks how many leading bytes of a guess are correct (CWE-208). Swapped it for `hash_equals()`, the constant-time comparison PHP provides for exactly this case. Straight drop-in: the decrypted token is always a plain string (`KeyCreationService` builds it from `str_random()`, and `ApiKey` declares `@property string $token`).

Also fixes a real bug the type-mode side of Psalm caught: two 2017 migrations (`RemoveDaemonSecretFromServersTable`, `RemoveDaemonSecretFromSubusersTable`) still reference `Pterodactyl\Contracts\Repository\DaemonKeyRepositoryInterface`, deleted in 703f55271 along with the rest of the daemon-key system. Both migrations still run on any fresh install (pre-2020 migration history), so `php artisan migrate` from scratch fatals on a class that no longer exists. Inlined the constant's value (`'i_'`, pulled from the pre-deletion source) so both migrations behave exactly as before.

A handful of other type-mode findings (28) turned up too, but this PR is taint-only in CI, so they're not baselined here. A plain `vendor/bin/psalm` run (no `--taint-analysis`) will surface them for whoever wants to dig in later. Listed them in a comment below for anyone who wants a starting point.


alies-dev:
Full list of the 28 type-mode findings from a plain `vendor/bin/psalm` run (`errorLevel="8"`, no `--taint-analysis`). None of these are wired into CI or baselined, this is just a starting point if anyone wants to dig in.

**Accessor/mutator visibility (18)**, `app/Models/`: Eloquent only dispatches `getXAttribute()`/`setXAttribute()` through `__get()`/`__set()`, never by name, so these can be `protected` instead of `public`. Checked, no direct calls anywhere in the codebase.
- `Allocation.php:89` `getHashidAttribute()`
- `Allocation.php:97` `getAliasAttribute()`
- `Allocation.php:105` `getHasAliasAttribute()`
- `Egg.php:159` `getCopyScriptInstallAttribute()`
- `Egg.php:172` `getCopyScriptEntryAttribute()`
- `Egg.php:185` `getCopyScriptContainerAttribute()`
- `Egg.php:197` `getInheritConfigFilesAttribute()`
- `Egg.php:209` `getInheritConfigStartupAttribute()`
- `Egg.php:221` `getInheritConfigLogsAttribute()`
- `Egg.php:233` `getInheritConfigStopAttribute()`
- `Egg.php:246` `getInheritFeaturesAttribute()`
- `Egg.php:259` `getInheritFileDenylistAttribute()`
- `EggVariable.php:82` `getRequiredAttribute()`
- `Schedule.php:129` `getHashidAttribute()`
- `Subuser.php:61` `getHashidAttribute()`
- `Task.php:130` `getHashidAttribute()`
- `User.php:226` `setUsernameAttribute()`
- `User.php:234` `getNameAttribute()`

**Missing template params (4)**: these classes extend a generic base without declaring the template, so Psalm can't check them against the base's type.
- `app/Extensions/Illuminate/Database/Eloquent/Builder.php:7` extends `Illuminate\Database\Eloquent\Builder`, expects 1 param
- `app/Extensions/Laravel/Sanctum/NewAccessToken.php:9` extends `Illuminate\Contracts\Support\Arrayable`, expects 2 params
- `app/Models/Filters/AdminServerFilter.php:8` extends `Spatie\QueryBuilder\Filters\Filter`, expects 1 param
- `app/Models/Filters/MultiFieldServerFilter.php:9` same, expects 1 param

**Possible dead code (3)**, flagged as `NoValue`, worth a manual look rather than a mechanical fix:
- `app/Http/Controllers/Admin/UserController.php:153`
- `app/Services/Allocations/FindAssignableAllocationService.php:88` (two separate reports on the same line)

**One-offs (3):**
- `app/Exceptions/Http/Connection/DaemonConnectionException.php:4`: the `@method \GuzzleHttp\Exception\GuzzleException getPrevious()` docblock override needs a return type Psalm can check against the parent signature.
- `bootstrap/tests.php:37`: `env()` called outside `config/`; returns null once config is cached.
- `public/index.php:22`: can't find `storage/framework/maintenance.php` at scan time, that file only exists once `php artisan down` has run. Likely a plugin false positive on a standard Laravel skeleton pattern rather than an app issue, most Laravel apps will hit this.

alies-dev:
Hey @anthonyphysgun 

Can I know a reason why it's closed?

anthonyphysgun:
Not interested in the package, thanks!

Issue #5683, opened 2026-07-16
Fresh Setup Database Migrations fails  [v1.14.1]


deB4SH:
### Current Behavior

Hi all,
I'm currently running into a setup issue with ptreodactyl in version 1.14.1.
The database is fresh without any setup. Image is provided the credentials via environment variables. The service starts and wants to seed the database but runs into an issue while running migrations.

```
ln: /var/log/panel/logs: File exists
external vars exist.
Checking if https is required.
Checking if letsencrypt email is set.
No letsencrypt email is set using http config.
Removing the default nginx config
Checking log folder permissions.
Fixing log folder permissions.
Checking database status.
IP_OF_DATABASE_HOST (IP_OF_DATABASE_HOST:3306) open
Migrating and Seeding D.B

   INFO  Preparing database.  

  Creating migration table ...................................... 65.97ms DONE

   INFO  Running migrations.  

  2016_01_23_195641_add_allocations_table ....................... 58.46ms DONE
  2016_01_23_195851_add_api_keys ................................ 55.87ms DONE
  2016_01_23_200044_add_api_permissions ......................... 93.91ms DONE
  2016_01_23_200159_add_downloads .............................. 147.13ms DONE
  2016_01_23_200421_create_failed_jobs_table .................... 70.78ms DONE
  2016_01_23_200440_create_jobs_table .......................... 158.59ms DONE
  2016_01_23_200528_add_locations .............................. 124.91ms DONE
  2016_01_23_200648_add_nodes .................................. 123.36ms DONE
  2016_01_23_201433_add_password_resets ........................ 186.32ms DONE
  2016_01_23_201531_add_permissions ............................. 69.07ms DONE
  2016_01_23_201649_add_server_variables ....................... 105.50ms DONE
  2016_01_23_201748_add_servers ................................ 315.13ms DONE
  2016_01_23_202544_add_service_options ......................... 56.83ms DONE
  2016_01_23_202731_add_service_varibles ........................ 58.11ms DONE
  2016_01_23_202943_add_services ................................ 91.20ms DONE
  2016_01_23_203119_create_settings_table ...................... 162.44ms DONE
  2016_01_23_203150_add_subusers ............................... 172.23ms DONE
  2016_01_23_203159_add_users .................................. 202.67ms DONE
  2016_01_23_203947_create_sessions_table ...................... 276.58ms DONE
  2016_01_25_234418_rename_permissions_column ................... 52.39ms DONE
[...]

PR #5688, opened 2026-07-25
fix: Removed backticks from migration script to fix issues with MariaDB connection type.

Sanic5238:
### Summary
Quick one line fix that fixes the migration failure reported in #5683

Most installs load the database/schema/mysql-schema.sql and never have to replay these migrations, since the environment variable db_connection=mariadb was set it doesn't use the squashed schema and has to replay all migrations from 2016. This connection variable is supported though it seems as per
https://github.com/pterodactyl/panel/blob/1.0-develop/config/database.php#L63 but it also defaulted above to mysql

This field was updated to include the backticks as part of https://github.com/pterodactyl/panel/commit/8ca098940afb497b8527bc2cfe1bb2bd340ad721

but I'm unable to confirm why, if I've missed something I apologize  but other references in this file seem to also use the barename only this one was quoted.

### Testing

Built from the current v1.14.1 release branch Dockerfile. With the schema dump removed to force a full replay for testing with the mysql database connection aswell.

- MariaDB 11.8.8 - All migrations passed
- MySQL - All Migrations Passed

Installs with the dump present are also unaffected by this modification. 

[Apologies for bad formatting, i'm terrible with markdown]

### DIFF
--- a/database/migrations/2020_04_10_141024_store_node_tokens_as_encrypted_value.php
+++ b/database/migrations/2020_04_10_141024_store_node_tokens_as_encrypted_value.php
@@ -25,7 +25,7 @@ class StoreNodeTokensAsEncryptedValue extends Migration
             $table->char('uuid', 36)->after('id');
             $table->char('daemon_token_id', 16)->after('upload_size');
 
-            $table->renameColumn('`daemonSecret`', 'daemon_token');
+            $table->renameColumn('daemonSecret', 'daemon_token');
         });
 
         Schema::table('nodes', function (Blueprint $table) {


rowansc1:
Thanks for the MR!

Sanic5238:
heya team,
took a quick gander at the CI failures in-case i caused a linting error (somehow on a 2 character change).
seems like PR#5583 may have came through without the linter and it wants some cosmetic changes to tests/Integration/Api/Client/AccountControllerTest.php, otherwise CI will continue to fail for future PR's

lajczi:
https://github.com/pterodactyl/panel/pull/5301

PR #5696, opened 2026-08-04
Disable reinstall action during normal server status if "skip egg install scripts" setting is enabled

### LINKED ISSUE #5306: Ignore Skip egg install script option for reinstall
### Current Behavior

While the skip egg install script is toggled in the startup section, reinstalling will not work.

### Expected Behavior

The reinstall runs instead of immediately being stopped. There should be either be a way of showing you that this option is enabled, or just ignore it altogether.

### Steps to Reproduce

1. Under startup check

### Panel Version

1.11.10

### Wings Version

1.11.13

Minecraft (Paper)

### Docker Image

ghcr.io/pterodactyl/yolks:java_21

### Error Logs

```bash
Should not be required, if needed I will provide logs per request.
```

- [x] I have searched the existing issues before opening this issue.
- [x] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [x] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

SkyMulley:
### Description

You're able to enable a "skip egg install scripts" per server, while still being able to reinstall a server. With this setting enabled, reinstalling a server has no functional use, the server is locked and shutdown, then simply unlocked immediately afterwards after a success signal from wings. This PR disables the reinstall button when skip install scripts is enabled for both the user and admin, with a little component telling you why you cannot reinstall. The reinstall button is not disabled if the server is in a status that reinstall could resolve (eg. failed install).

### Resolved issues:
1. [x] resolves pterodactyl/panel#5306

### DIFF
--- a/app/Exceptions/DisplayException.php
+++ b/app/Exceptions/DisplayException.php
@@ -49,7 +49,7 @@ class DisplayException extends PterodactylException implements HttpExceptionInte
      */
     public function render(Request $request): JsonResponse|RedirectResponse
     {
-        if ($request->expectsJson()) {
+        if ($request->expectsJson() || $request->is('api/*')) {
             return response()->json(Handler::toArray($this), $this->getStatusCode(), $this->getHeaders());
         }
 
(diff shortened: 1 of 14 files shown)


SkyMulley:
https://github.com/user-attachments/assets/3435d383-2451-4c67-aa3c-547f57a6f754

PR #5699, opened 2026-08-10
Add Velocity egg with Java 25 support

iamthealfaandtheomega:
Adds a game egg for Velocity (Minecraft proxy) that did not exist in the panel's egg collection.

Highlights:
- Runs on Java 25 and Java 21 images (JDK 25 is already in the other Minecraft eggs)
- Installs the latest stable release build via the Fill v3 API, skipping snapshot/experimental trains
- Verifies the jar's SHA-256 against the value published by Fill before promoting it into place; on mismatch the install fails cleanly and no partial jar is left behind
- Uses Wings' `file` parser to rewrite the proxy bind address (TOML parsing is not supported by Wings and would be a no-op)
- Fails with a clear message when the API is unreachable

Tested locally: successful install (jar checksum verified), tampered download (clean abort), snapshot-only feed (clean abort), API outage (clean error), and config rewrite applied at runtime.


SerenModz21:
It's worth noting that a lot of eggs are available at https://eggs.pterodactyl.io/, including Velocity.

- https://eggs.pterodactyl.io/egg/games-velocity/
- Source: https://github.com/pterodactyl/game-eggs/blob/main/minecraft/proxy/java/velocity/egg-velocity.json
- Java 25 PR: https://github.com/pterodactyl/game-eggs/pull/602

iamthealfaandtheomega:
x1
Java 25 is not there.

SerenModz21:
I included the following PR at the end of my original comment:

 - https://github.com/pterodactyl/game-eggs/pull/602
 
 My apologies for not making it clearer.

PR #5701, opened 2026-08-11
Generate HASHIDS_SALT for existing installs

### LINKED ISSUE #5700: Generation of HASHIDS_SALT does not work for existing installs
### Current Behavior

Currently the code for generating HASHIDS_SALT if not provided are guarded within an else block of the following clause: if [ -f /app/var/.env ];

Meaning if .env exists from previous installations, the HASHIDS_SALT gets never generated and requires users to delete the .env in order for the startup scrip to generate the variable.

### Expected Behavior

HASHIDS_SALT should be generated when missing whether it is the first run or not.

### Steps to Reproduce

from an existing installation, remove HASHIDS_SALT from /app/var/.env and make sure it is not set in docker compose. Restart the container. This should cause "Database" tab of a game server in the customer UI to throw an error 500

### Panel Version

1.15.0

### Wings Version

1.13.2

_No response_

### Docker Image

_No response_

### Error Logs

```bash

```

jondycz:
Simply moves the code for HASHIDS_SALT outside of the else block guarded by clause checking for existence of .env - this breaks generation of HASHIDS_SALT for existing installs lacking the env variable causing error 500 on the user dashboard in "Database" tab.

Additionally checks for existence of HASHIDS_SALT in /app/var/.env in order to not inflate the file size by redefining the variable on every restart.

fixes #5700

### DIFF
--- a/.github/docker/entrypoint.sh
+++ b/.github/docker/entrypoint.sh
@@ -26,18 +26,18 @@ else
     echo -e "APP_KEY=$APP_KEY" > /app/var/.env
   fi
 
-  ## generate a random salt for hashids if not provided
-  if [ -z $HASHIDS_SALT ]; then
-     echo -e "Generating hashids salt."
-     HASHIDS_SALT=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9!@#$%^&*()_+?><~' | fold -w 20 | head -n 1)
-     echo -e "Generated hashids salt: $HASHIDS_SALT"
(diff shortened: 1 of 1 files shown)


danny6167:
This looks like it would append an additional HASHIDS_SALT= line to the .env file every time the container starts, unless you declare the HASHIDS_SALT in the containers environment.

jondycz:
> This looks like it would append an additional HASHIDS_SALT= line to the .env file every time the container starts, unless you declare the HASHIDS_SALT in the containers environment.

Just noticed that. Will update the PR.

PR #5705, opened 2026-08-15
fix: make p:upgrade survive its own dependency swap

svaningelgem:
The upgrade guide currently tells people not to use `p:upgrade` ("The self-upgrade is currently in-operable due to issues with some dependencies we make use of") and lists manual steps that are the same steps this command already runs.

`composer install` replaces the autoloader and every class under `vendor/` while the command that started it is still running, and PHP cannot unload classes it has already loaded.
Every `$this->call()` after that line runs against a mix of the definitions in memory and the files now on disk.
The `require bootstrap/app.php` on `1.0-develop` rebuilds the container, but the container is not the stale part.

So the upgrade is now split over two processes.
The first runs from the installed code and stops once the new code is on disk.
The second is a fresh `artisan p:upgrade --finalize` that boots the code it is about to migrate.

`composer.json` already does this for the same reason, running `@php artisan package:discover` as its own process in `post-autoload-dump`.

The rest is fallout from actually reading the command:

- No exit status was ever checked. All eleven `$process->run()` results were discarded (ref: #3159).
- `artisan down` ran after the archive was unpacked. Now it runs before, so new code is never served against the old schema.
- `--user` and `--group` were only tested with `is_null()` and their values never assigned, so `--user=nginx` still produced `chown -R www-data:www-data`.
- The PHP version guard called `$this->error()` and then carried on with no `return`.
- `composer check-platform-reqs`: Composer evaluates its own constraints and catches missing extensions too. A hardcoded `8.2.0` would have passed on PHP 8.4, where `composer install` then fails, since `require` is `^8.2 || ^8.3`.
- The archive is downloaded to a temporary file instead of piped straight into `tar`
- `chown -R` targets `.` instead of `*`, which skipped every dotfile including `.env`.
- Pre-flight checks the required binaries, a writable tree, free disk space and a reachable database before anything is touched.
- A failure before the Panel goes offline says so and changes nothing. A failure after it goes offline leaves maintenance mode on deliberately and prints how to resume, rather than exposing a half upgraded tree.

Not addressed: unpacking over the top still never deletes files removed upstream. Fixing that means release directories and an atomic symlink swap, which would restructure the webroot of every existing installation. The manual guide has the same gap.

Tests cover the command end to end: both halves, the pre-flight failures, checksum match and mismatch, the platform check passing and failing, the interactive prompts, and --url and --release. That is 98.7% of lines and 99.25% of branches; what is left is the free disk space guard, which cannot be provoked without adding a seam to the command.

### DIFF
--- a/app/Console/Commands/UpgradeCommand.php
+++ b/app/Console/Commands/UpgradeCommand.php
@@ -6,6 +6,7 @@ use Illuminate\Console\Command;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Process;
 use Symfony\Component\Process\ExecutableFinder;
+use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Process\Process as SymfonyProcess;
 
 class UpgradeCommand extends Command
@@ -28,6 +29,8 @@ class UpgradeCommand extends Command
 
     protected $description = 'Downloads a new archive for Pterodactyl from GitHub and then executes the normal upgrade commands.';
 
+    private ?ProgressBar $bar = null;
+
     /**
      * Splits the upgrade over two processes: composer install replaces every class
      * under vendor/ while this one is running, and PHP cannot unload what it has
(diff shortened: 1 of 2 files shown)


robertdrakedennis:
Hi as part of our contributing guide I'm going to need to see this in a bit of a different format before I review the pr.

svaningelgem:
> Hi as part of our contributing guide I'm going to need to see this in a bit of a different format before I review the pr.

Fair pushback. I read it before posting, but I noticed some things I'd still change.

Changes done and it's now much closer to how I prefer to see my own code.

Thanks for reminding me not to always trust bots 😉

svaningelgem:
Hi @robertdrakedennis , anything that you want me to change?

PR #5725, opened 2026-09-07
Fix tooltip title to display version correctly

JojoJules:
On the admin nodes list page, hovering over a node's status/heart icon shows a tooltip with the Wings daemon version. The version is currently displayed with a duplicated `v` prefix, e.g. `vv1.13.3` instead of `v1.13.3`.

### DIFF
--- a/resources/views/admin/nodes/index.blade.php
+++ b/resources/views/admin/nodes/index.blade.php
@@ -87,7 +87,7 @@
                 timeout: 5000
             }).done(function (data) {
                 $(element).find('i').tooltip({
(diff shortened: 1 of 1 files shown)


AndyIsHereBoi:
I'm pretty sure this has already been fixed, this is like the 10th time I have seen a issue/pr on this problem

JojoJules:
> I'm pretty sure this has already been fixed, this is like the 10th time I have seen a issue/pr on this problem

Its not fixed. Still shows double v
image

